PR #1, opened 2019-08-12
Implement global sync

PeterTh:
Adds a global sync feature, primarily as an incremental development and debugging aid to improve the developer experience.


psalz:
LGTM!

PR #5, opened 2019-12-05
runtime features

n-io:
Added logging for runtime features on debug level. Also split matmul example into matrix multiply and chained matrix multiply of the form (AB)(CD).

### DIFF
--- a/include/ranges.h
+++ b/include/ranges.h
@@ -3,6 +3,8 @@
 #include <CL/sycl.hpp>
 
 #include "workaround.h"
+#define FMT_HEADER_ONLY
+#include <spdlog/fmt/fmt.h>
 
 #if !WORKAROUND_HIPSYCL
 #define __host__
@@ -64,6 +66,17 @@ struct subrange {
 	subrange(chunk<Dims> other) : offset(other.offset), range(other.range) {}
 
 	bool operator==(const subrange& rhs) { return offset == rhs.offset && range == rhs.range; }
+
+	std::string to_string() const {
+		if(dims == 1)
+			return fmt::format("subrange({}) at offset({})", range[0], offset[0]);
+		else if(dims == 2)
+			return fmt::format("subrange({},{}) at offset({},{})", range[0], range[1], offset[0], offset[1]);
+		else
+			return fmt::format("subrange({},{},{}) at offset({},{},{})", range[0], range[1], range[2], offset[0], offset[1], offset[2]);
+	}
+
+	size_t num_elems() const { return range.size(); }
 };
 
 } // namespace celerity
--- a/include/runtime.h
+++ b/include/runtime.h
@@ -79,6 +79,11 @@ namespace detail {
 			return buffer_ptrs.count(bid) == 1;
 		}
 
+		size_t getElemSizeOfBuffer(buffer_id bid) const {
+			std::lock_guard<std::mutex> lock(buffer_mutex);
+			return buffer_ptrs.find(bid)->second->get_element_size();
+		}
+
 		std::shared_ptr<raw_data_read_handle> get_buffer_data(buffer_id bid, const cl::sycl::id<3>& offset, const cl::sycl::range<3>& range) const {
 			std::lock_guard<std::mutex> lock(buffer_mutex);
 			assert(buffer_ptrs.count(bid) == 1);
--- a/src/worker_job.cc
+++ b/src/worker_job.cc
@@ -75,6 +75,14 @@ namespace detail {
 
 	bool push_job::execute(const command_pkg& pkg, std::shared_ptr<logger> logger) {
 		if(data_handle == nullptr) {
+			// const celerity::detail::push_data
+			const auto data = boost::get<push_data>(pkg.data);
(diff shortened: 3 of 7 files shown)


include/ranges.h
```diff
@@ -64,6 +66,17 @@ struct subrange {
...
 	bool operator==(const subrange& rhs) { return offset == rhs.offset && range == rhs.range; }
+
+	std::string to_string() const {
+		if(dims == 1)
+			return fmt::format("subrange({}) at offset({})", range[0], offset[0]);
+		else if(dims == 2)
+			return fmt::format("subrange({},{}) at offset({},{})", range[0], range[1], offset[0], offset[1]);
+		else
+			return fmt::format("subrange({},{},{}) at offset({},{},{})", range[0], range[1], range[2], offset[0], offset[1], offset[2]);
+	}
```
psalz: I recently merged a commit by @PeterTh (3ff6530) that introduces generic printing capabilities for `celerity::subrange` and `celerity::chunk`. You should be able to use those for your purposes.

include/ranges.h
```diff
@@ -64,6 +66,17 @@ struct subrange {
...
+
+	size_t num_elems() const { return range.size(); }
```
psalz: The subrange is a generic "range with an offset", so I think its somewhat misleading to refer to "elements" here. After all, the subrange doesn't always describe the subset of a buffer, but instead could also describe the number of "items" in a kernel execution space.

That being said, `subrange.range` is accessible anyways, so why don't you just use `subrange.range.size()` for this?

include/runtime.h
```diff
@@ -79,6 +79,11 @@ namespace detail {
 			return buffer_ptrs.count(bid) == 1;
 		}
 
+		size_t getElemSizeOfBuffer(buffer_id bid) const {
+			std::lock_guard<std::mutex> lock(buffer_mutex);
+			return buffer_ptrs.find(bid)->second->get_element_size();
+		}
```
psalz: Please use `standard_case` instead of `camelCase` for member functions.

src/worker_job.cc
```diff
@@ -75,6 +75,14 @@ namespace detail {
 
 	bool push_job::execute(const command_pkg& pkg, std::shared_ptr<logger> logger) {
 		if(data_handle == nullptr) {
+			// const celerity::detail::push_data
+			const auto data = boost::get<push_data>(pkg.data);
+			size_t elemSize = runtime::get_instance().getElemSizeOfBuffer(data.bid);
```
psalz: Please use `standard_case` for local variables as well!

src/worker_job.cc
```diff
@@ -75,6 +75,14 @@ namespace detail {
...
+			size_t numElems = data.sr.num_elems();
+			size_t bufTransferSize = elemSize * numElems;
+			logger->debug(logger_map(
```
psalz: Please use the `trace` logging level here. We don't use `debug` anywhere else yet (and logging is a bit all of the place and in need of refactoring right now anyway, so we shouldn't add even more variations).

psalz:
Hi, thanks for your contribution!

Unfortunately there are some issues with this PR, the biggest of which being that it breaks the unit tests. Please make sure to always run the tests before filing a PR, at least until we've got our CI setup working on GitHub :). As an side, please try to provide a commit message that uses imperative mood and is a bit more descriptive.

The reason for the tests failing is that you're accessing the `runtime` singleton from within the `graph_generator` in order to log your message. However, this breaks the inversion of control (`graph_generator` should have no knowledge of the runtime) and there is no runtime during unit testing. You could alleviate this by passing a `logger` into the `graph_generator` for this purpose, but it seems a bit overkill for a single message. That being said, I don't think the `graph_generator` is the right place to log this message anyway. May I suggest moving it to `distr_queue::submit` instead? Here we already access the `runtime` singleton anyway, and you have direct access to the task object. You could then simply do something like this:

```cpp
auto tid = detail::runtime::get_instance().get_task_manager().create_compute_task(cgf);
auto ctsk = static_cast<const detail::compute_task*(
    detail::runtime::get_instance().get_task_manager().get_task(tid).get());
detail::runtime::get_instance().get_logger()->trace(
    detail::logger_map({{"event", fmt::format("Creating COMPUTE task with global size {} (total work item count {})",
    subrange<3>(ctsk->get_global_offset(), ctsk->get_global_size()), ctsk->get_global_size().size())}}));
```

I've added some additional review notes directly at the relevant code locations.

Finally, I don't think we should add another matrix multiplication example into the main repository. The first and foremost goal of these examples is to provide a reference for new users of Celerity, and I don't think having two essentially identical examples has much additional value. The reason why the example currently does two matrix multiplications instead of just one is to demonstrate Celerity's ability to automatically manage data dependencies during distribution execution (which is one of its main features). 

I understand you might need this for benchmarking, but I think it should be easy to maintain this in your own fork (as the examples aren't changed all that often).

src/worker_job.cc
n-io: Debug not being used anywhere else seems like a good reason to use it for this.

include/ranges.h
n-io: Not trying to print, trying to log. Unfortunately, the commit does not cater for that.

PR #10, opened 2020-05-12
parallel_for_work_group

fknorr:
Introduce handler::parallel_for_work_group

- Split tasks along work group boundaries
- Add the matmul_tiled example


psalz:
To give some context on this PR being on hold right now: We are currently awaiting an answer on https://github.com/KhronosGroup/SYCL-Docs/issues/99, as the current [workaround](https://github.com/celerity/celerity-runtime/blob/6fd83c709c5b0eaf15362e15595e8ac590cb2892/include/handler.h#L133) for `parallel_for_work_group` not having an `offset` parameter turned out to have a non-negligible performance overhead.

illuhad:
fyi, I would recommend waiting until the SYCL 2020 provisional spec is release before putting more work in hierarchical parallelism.

PR #12, opened 2020-05-20
Enable SYCL-native kernel profiling

fknorr:
So far, execution times were either measured on the host, which is pretty inaccurate, or through OpenCL, which was only possible for ComputeCpp. Once changes from https://github.com/illuhad/hipSYCL/pull/251 are merged, hipSYCL will support `cl::sycl::event::get_profiling_info<>()`, providing accurate timings for both ComputeCpp and hipSYCL. This PR must not be merged before the upstream hipSYCL PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Celerity's runtime behavior:
   `CELERITY_DEVICES="<platform_id> <first device_id> <second device_id> ... <nth device_id>"`.
   Note that this should normally not be required, as Celerity will attempt to
   automatically assign a unique device to each worker on a host.
-- `CELERITY_PROFILE_OCL` controls whether OpenCL-level profiling information
+- `CELERITY_PROFILE_KERNEL` controls whether SYCL queue profiling information
   should be queried (currently not supported when using hipSYCL).
 
 ## Disclaimer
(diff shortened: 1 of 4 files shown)


psalz:
I've hijacked this PR in the hope that we can get it finally merged (at least partially). I've kept the renaming of `CELERITY_PROFILE_OCL` to `CELERITY_PROFILE_KERNEL`, as well as the switch to pure SYCL queue profiling (instead of using OpenCL APIs). However, I had to disable the functionality again for hipSYCL, as [illuhad/hipSYCL#428](https://github.com/illuhad/hipSYCL/pull/428) unfortunately still has not been merged.

fknorr:
I agree this is the best solution until hipSYCL implements proper profiling support.

psalz:
> We might want to add an override though for people using the patched version of hipSYCL to re-enable this.

The problem with this is that currently the `sycl::property::queue::enable_profiling` property does not exist in hipSYCL, so this would have to be a compile time option. I would argue that anyone willing to build your fork *and* recompile Celerity to get this to run (which is mostly us, realistically) might as well just patch it manually.

PR #13, opened 2020-05-20
Improve build system, switch to C++17

fknorr:
Several improvements to the CMake / Compiler / IDE infrastructure.

- Switch to C++17
- Add IDE config files
- Improve Celerity's CMake config options


fknorr:
We should probably verify that install targets still work on Windows (they do on Linux).

PR #15, opened 2020-07-16
Add workarounds for Boost 1.69+ hipSYCL/CUDA issues

psalz:
For some time now we had some issues compiling against newer versions of Boost (1.69+) when targeting the hipSYCL CUDA backend. The exact issue can vary from version to version (and sometimes there are several), but everything boils down to the fact that Boost is not really aware that CUDA code can also be compiled using Clang (as opposed to NVCC). Since Boost does compiler detection all over the place, it sometimes ends up generating code targeted at NVCC, Clang or both, resulting in chaos and usually some very long template backtraces.

While not very pretty, with these workarounds I was able to compile Celerity against all recent versions of Boost (1.65 - 1.73) using hipSYCL CUDA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,13 +128,10 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
   target_compile_options(celerity_runtime PRIVATE -Wall -Wextra -Wno-unused-parameter)
 endif()
 
-if(CELERITY_SYCL_IMPL STREQUAL "hipSYCL")
-  # Boost currently (as of 1.71) does not enable variadic macros
-  # when it detects CUDA compilation.
-  # Since we are using Clang however (via hipSYCL) instead of NVCC,
-  # the macros can and should be enabled.
-  # See https://github.com/boostorg/preprocessor/issues/24
-  target_compile_definitions(celerity_runtime PUBLIC BOOST_PP_VARIADICS=1)
+if(CELERITY_SYCL_IMPL STREQUAL "hipSYCL" AND HIPSYCL_PLATFORM_CANONICAL MATCHES "cuda|rocm")
+  # Boost does not anticipate us compiling CUDA with Clang, so we have to enable several workarounds.
+  target_compile_definitions(celerity_runtime PUBLIC __CELERITY_ENABLE_BOOST_WORKAROUNDS__)
+  target_compile_options(celerity_runtime PRIVATE "-include${CMAKE_CURRENT_SOURCE_DIR}/include/boost_workarounds.h")
 endif()
 
 # Examples
--- a/include/boost_workarounds.h
+++ b/include/boost_workarounds.h
@@ -1,0 +1,54 @@
+#pragma once
+
+/**
+ * As of April 2020, Boost (1.73) does not have explicit support for CUDA compilation using Clang.
+ * Instead, Boost assumes (in most places) that CUDA code is always compiled with NVIDIA's NVCC.
+ * Detection of CUDA compilation is based on the __CUDACC__ macro. Unfortunately, this macro is also
+ * set by Clang, alongside the __clang__ macro. This combination is not anticipated by Boost, and
+ * thus, chaos ensues.
+ *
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -128,13 +128,10 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
...
-if(CELERITY_SYCL_IMPL STREQUAL "hipSYCL")
-  # Boost currently (as of 1.71) does not enable variadic macros
-  # when it detects CUDA compilation.
-  # Since we are using Clang however (via hipSYCL) instead of NVCC,
-  # the macros can and should be enabled.
-  # See https://github.com/boostorg/preprocessor/issues/24
-  target_compile_definitions(celerity_runtime PUBLIC BOOST_PP_VARIADICS=1)
+if(CELERITY_SYCL_IMPL STREQUAL "hipSYCL" AND HIPSYCL_PLATFORM_CANONICAL MATCHES "cuda|rocm")
+  # Boost does not anticipate us compiling CUDA with Clang, so we have to enable several workarounds.
+  target_compile_definitions(celerity_runtime PUBLIC __CELERITY_ENABLE_BOOST_WORKAROUNDS__)
```
fknorr, line 133: Identifiers starting with double underscores are technically UB, suggest renaming this to `CELERITY_ENABLE_BOOST_WORKAROUNDS`

include/boost_workarounds.h
```diff
@@ -0,0 +1,54 @@
...
+ *  boost::optional<bar> foo;
+ *  foo = bar{};
+ * ```
+ *
+ * Unfortunately there is no easy way to enable all Clang intrinsics, as that particular
+ * header (boost/type_traits/intrinsics.hpp) checks for __CUDACC__ directly.
+ *
+ * Required for Boost 1.65 - 1.68 (if using the Clang config above), 1.69 - 1.73 (not fixed yet).
+ */
+#define BOOST_IS_BASE_OF(T, U) (__is_base_of(T, U) && !is_same<T, U>::value)
```
fknorr, line 46: Is this replacing an existing #define from Boost? If so, we should include `boost/type_traits/intrinsics.hpp` explicitly in this file and `#undef` before defining this to avoid side-effects from inclusion order.

include/boost_workarounds.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Enable variadic preprocessor macros. Used by the WORKAROUND macros in workaround.h.
+ *
+ * Required for Boost 1.65, 1.66, 1.69 - 1.72 (fixed in 1.73).
+ * See https://github.com/boostorg/preprocessor/issues/24
+ */
+#define BOOST_PP_VARIADICS 1
```
fknorr, line 54: Likewise, we could assert here that Boost.Preprocessor has not been included yet, altough this might be more complicated.

CMakeLists.txt
psalz: Hmm, yes. I'd just like to convey that this is not really something that should be set by users (unlike e.g. `CELERITY_STRICT_CGF_SAFETY`). How about `CELERITY_DETAIL_ENABLE_BOOST_WORKAROUNDS`..?

include/boost_workarounds.h
psalz: No, this will simply be undefined for Clang-CUDA. However if this issue is fixed in the future, we will produce a redefinition warning here, so your point still stands.

include/boost_workarounds.h
psalz: Yes, maybe this header should explicitly assert that Boost has not yet been included at all, and `#error` otherwise. Because right now, a user could still do

```cpp
#include <boost/optional.hpp>
#include <celerity/celerity.h>
```

and break everything.

psalz:
Note to self: Apart from the issues above this change also seems to break atomics in Boost Lockfree for some setups/architectures. Investigate!

roma2 (commented):
Duplicate of #boost-workarounds

psalz:
Closing in favor of #23.

PR #18, opened 2020-10-05
Get rid of nested SYCL command group submissions

fknorr:
Previously, in the live pass, a Celerity CGF was executed within a SYCL CGF. Buffer accesses done inside the Celerity CGF submitted further buffer copy operations to SYCL, so there were nested `sycl::handler::submit` calls. This was probably not conformant and broke with the new hipSYCL runtime.

This moves the Celerity CGF call out of the SYCL CGF, so that get_access calls can submit copy operations in a defined manner. Calling `sycl::handler::require` on the accessor is not straightforward, since it must happen after the call to `get_access` and inside the SYCL CGF. We now use the copy constructor / assignment operator of `celerity::accessor` that happens on kernel submission to call `require()`.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -7,7 +7,6 @@
 
 #include "access_modes.h"
 #include "buffer_storage.h"
-#include "handler.h"
 
 
 namespace celerity {
@@ -144,7 +143,7 @@ class accessor<DataT, Dims, Mode, cl::sycl::access::target::global_buffer>
 	friend accessor<T, D, M, cl::sycl::access::target::global_buffer> detail::make_device_accessor(Args&&...);
 
 	// see init_from
-	detail::live_pass_device_handler* live_cgh = nullptr;
+	cl::sycl::handler* const* eventual_sycl_cgh = nullptr;
 	cl::sycl::accessor<DataT, Dims, Mode, cl::sycl::access::target::global_buffer, cl::sycl::access::placeholder::true_t> sycl_accessor;
 	cl::sycl::id<Dims> backing_buffer_offset;
 
@@ -152,9 +151,9 @@ class accessor<DataT, Dims, Mode, cl::sycl::access::target::global_buffer>
 	accessor(cl::sycl::buffer<DataT, Dims>& faux_buffer)
 	    : sycl_accessor(cl::sycl::accessor<DataT, Dims, Mode, cl::sycl::access::target::global_buffer, cl::sycl::access::placeholder::true_t>(faux_buffer)) {}
 
-	accessor(detail::live_pass_device_handler& live_cgh, cl::sycl::buffer<DataT, Dims>& buffer, const cl::sycl::range<Dims>& range,
+	accessor(cl::sycl::handler* const* eventual_sycl_cgh, cl::sycl::buffer<DataT, Dims>& buffer, const cl::sycl::range<Dims>& range,
 	    cl::sycl::id<Dims> backing_buffer_offset)
-	    : live_cgh(&live_cgh),
+	    : eventual_sycl_cgh(eventual_sycl_cgh),
 	      // We pass a range and offset here to avoid interference from SYCL, but the offset must be relative to the *backing buffer*.
 	      sycl_accessor(cl::sycl::accessor<DataT, Dims, Mode, cl::sycl::access::target::global_buffer, cl::sycl::access::placeholder::true_t>(
 	          buffer, range, backing_buffer_offset)),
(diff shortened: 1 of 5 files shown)


fknorr:
CI: `matmul` breaks with the old runtime (hipSYCL stable), but `runtime_tests` work fine. With the new runtime, everything passes.
```
$ /lib-installs-local/openmpi-4.0.2/bin/mpirun -n 2 build/examples/matmul/matmul
[2020-10-06 14:06:33.046] [default] [info] [rank = 1] sycl = hipSYCL 0.8.2, build = release, pid = 7225, event = initialized
[2020-10-06 14:06:33.046] [default] [info] [rank = 0] sycl = hipSYCL 0.8.2, build = release, pid = 7224, event = initialized
[2020-10-06 14:06:33.079] [default] [info] [rank = 1] Using platform 'hipSYCL [SYCL over CUDA/HIP] on NVIDIA CUDA', device 'GeForce RTX 2070' (automatically selected platform 0, device 1)
[2020-10-06 14:06:33.079] [default] [info] [rank = 0] Using platform 'hipSYCL [SYCL over CUDA/HIP] on NVIDIA CUDA', device 'GeForce RTX 2070' (automatically selected platform 0, device 0)
[2020-10-06 14:06:33.220] [bench] [info] [rank = 0] event = userConfig, matSize = 1024
[2020-10-06 14:06:33.220] [bench] [info] [rank = 0] id = 0, name = main program, event = beginSection
[2020-10-06 14:06:33.221] [bench] [info] [rank = 1] id = 0, name = main program, event = beginSection
[hipSYCL Error] check_error: Received unknown HIP error 700, throwing memory_allocation_error.
[hipSYCL Error] task_graph: task_done_callback caught async error,  invoking async handler.
[hipSYCL Error] check_error: Received unknown HIP error 700, throwing memory_allocation_error.
[hipSYCL Error] task_graph: task_done_callback caught async error,  invoking async handler.
[2020-10-06 14:06:33.382] [default] [error] [rank = 1] SYCL asynchronous exception: Unknown error occured
[2020-10-06 14:06:33.382] [default] [error] [rank = 1] SYCL asynchronous exception: Unknown error occured
[hipSYCL Error] check_error: Received hipErrorUnknown,  throwing exception.
[hipSYCL Error] task_graph: submit() caught async error,  invoking async handler.
[hipSYCL Error] SYCL exception details: 'Unknown HIP error', HIP error code = 1030
[2020-10-06 14:06:33.383] [default] [error] [rank = 1] SYCL asynchronous exception: Unknown HIP error
```

psalz:
I did some investigating to try and figure out why matmul (and any other program run on more than 1 GPU) would crash with the old hipSYCL runtime. The crash seems to happen because according to `cuda-memcheck`, apparently there is an access to an invalid memory address. Turns out this is a hipSYCL bug I've already encountered once, a long time ago. I don't remember the details, but for some reason the old hipSYCL runtime would sometimes (when interacting with it from multiple threads) get the active GPU device confused, and resort back to the default device (id 0). This is what is happening here; the allocation does take place, only on the wrong GPU. Everything works when running `CELERITY_DEVICES="0 0 0" mpirun -n 2 ./examples/matmul/matmul` (i.e., forcing both workers to run on the same GPU).

I'm glad to see that this bug has been fixed in the hipSYCL new runtime. I don't think there is any point in trying to work around this for the old runtime; instead we can just bump our minimum required version to what looks to be hipSYCL@0.9.0. However, before we can do that (and update our CI runner to use that hipSYCL version), we need https://github.com/illuhad/hipSYCL/pull/307 to be merged. Also it'd be great if we could codify this requirement in our CMakeLists, however it looks like hipSYCL currently doesn't specify a version for its exported library targets (we should look into adding that).

Also, minor nitpick: I think you accidentally mixed up the wording in your commit title -- "command SYCL group" instead of "SYCL command group" :).

illuhad:
> I did some investigating to try and figure out why matmul (and any other program run on more than 1 GPU) would crash with the old hipSYCL runtime. The crash seems to happen because according to cuda-memcheck, apparently there is an access to an invalid memory address. Turns out this is a hipSYCL bug I've already encountered once, a long time ago. I don't remember the details, but for some reason the old hipSYCL runtime would sometimes (when interacting with it from multiple threads) get the active GPU device confused, and resort back to the default device (id 0).

I think this is a fundamental limitation of the old runtime. The buffer class was hardcoded to manage exactly one device and one host allocation. Additionally, the old runtime was fundamentally incapable of performing lazy allocation. So, the buffer would just `hipMalloc()` in the constructor. And because at that time the buffer has no information where the data is going to be used, it cannot call `hipSetDevice()` before the `hipMalloc` call. In consequence, the device that the buffer allocates on depends on whether previous operations have called `hipSetDevice()` in a multi-gpu context...

This is of course terrible, but fortunately with the new runtime a buffer manages an arbitrary amount of allocations on multiple devices, that are only allocated lazily when needed on a specific device. Therefore this issue should be fixed now :)

psalz:
hipSYCL 0.9.0, which includes the fix for the vector type alignments, was just released (🎉) and we should now be able to merge this. I've updated the CI runner to 0.9.0; I couldn't figure out how to force a re-run of the workflow with the new UI (possibly because the previous run was too long ago), so I just force-pushed the same commit again.

psalz (approved):

PR #20, opened 2020-10-27
Introduce custom preprocessor flag for debugging

### LINKED ISSUE #19: buffer_manager causes a segment fault
**Describe the bug :boom:**
Some weird error.  
Building `celerity_runtime` with the examples and tests works just fine. All runtime tests passes and the examples runs also fine. 
Now the weird think, `add_celerity_to_target` cause a segment fault when it try's to create a `celerity::buffer`. The segment fault occur in `buffer_manager.h` on line 130.

**To Reproduce :repeat:**
1. Open CMakeList.txt of any example
2. Add `find_package(Celerity 0.2.1 REQUIRED CONFIG HINTS "/opt/celerity/lib")`
3. Remove `target_link_libraries` and `add_sycl_to_target`
4. Add (for matmul as example)
```
add_celerity_to_target(
  TARGET matmul
  SOURCES matmul.cc
)
```
5. Build and run

**Output**
```bash
[2020-10-26 21:06:53.208] [default] [info] [rank = 0] sycl = hipSYCL 0.8.2, build = release, pid = 12708, event = initialized
[2020-10-26 21:06:53.220] [default] [info] [rank = 0] Using platform 'hipSYCL [SYCL over CUDA/HIP] on NVIDIA CUDA', device 'GeForce GTX 1050' (automatically selected platform 0, device 0)
[2020-10-26 21:06:53.280] [bench] [info] [rank = 0] event = userConfig, matSize = 1024
[XPS-15-9560:12708] *** Process received signal ***
[XPS-15-9560:12708] Signal: Segmentation fault (11)
[XPS-15-9560:12708] Signal code: Address not mapped (1)
[XPS-15-9560:12708] Failing at address: 0x71
[XPS-15-9560:12708] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7f92c81723c0]
[XPS-15-9560:12708] [ 1] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x434973]
[XPS-15-9560:12708] [ 2] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x433ebd]
[XPS-15-9560:12708] [ 3] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x433b8c]
[XPS-15-9560:12708] [ 4] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x433a35]
[XPS-15-9560:12708] [ 5] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x42d4f5]
[XPS-15-9560:12708] [ 6] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x42cff9]
[XPS-15-9560:12708] [ 7] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x429641]
[XPS-15-9560:12708] [ 8] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x4291d7]
[XPS-15-9560:12708] [ 9] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x41fd94]
[XPS-15-9560:12708] [10] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x41a51c]
[XPS-15-9560:12708] [11] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f92c7c470b3]
[XPS-15-9560:12708] [12] /home/salih/projects/c++/celerity-runtime/cmake-build-debug/examples/matmul/matmul[0x41a28e]
[XPS-15-9560:12708] *** End of error message ***

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

**Environment :earth_americas:**
- OS: Ubuntu 20.04
- Sycl: hipSYCL 0.8.2
- Celerity: 0.2.1
- Clang 10 and CUDA 10.1

psalz:
This introduces the `CELERITY_DETAIL_ENABLE_DEBUG` flag, which is set for debug builds of the Celerity runtime, regardless of a user target's build type. This is more reliable than using `NDEBUG`, as headers and implementation files are always guaranteed to receive the flag, both during library compilation and when linking to it. Also it means having less of those confusing double-negative `!defined(NDEBUG)` checks.

Unfortunately the implementation turned out to be quite annoying (of course...), as CMake doesn't really expose a mechanism for querying the build configuration of a library target that is being linked (which could then be used e.g. in generator expressions). The workaround for now is to inspect the `IMPORTED_CONFIGURATIONS` of the `Celerity::celerity_runtime` library, and use that to conditionally set the flag. For this to work however we require that only a single build configuration is imported in the first place, as otherwise it becomes more difficult to tell which configuration ends up being linked (albeit not impossible).

While thus far we effectively already only really supported one installed build configuration (as we'd otherwise have to set the `CMAKE_<CONFIG>_POSTFIX` variables), instead of adding support for multiple configurations, I've decided to enforce single config installs with an error for now. While I do see merit in having that feature, I can also see it causing a lot of unnecessary headaches during development, when rebuilding one configuration of the runtime, while some executable still links to the other one... So maybe we could look into e.g. only enabling this for versioned releases.

This should fix #19.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,24 @@ target_link_libraries(celerity_runtime PUBLIC
   spdlog::spdlog
 )
 
+# For debug builds, we set the CELERITY_DETAIL_ENABLE_DEBUG preprocessor flag,
+# which allows Celerity to control debug functionality within headers regardless
+# of a user target's build type. (This flag is not intended to be modified by
+# end users directly).
+#
+# To make this work, we need to ensure that user targets also receive this flag
+# whenever they link to a Celerity runtime that was built with the DEBUG
+# configuration. Unfortunately there doesn't seem to be a way of doing this using
+# generator expressions, which is why we have to do it manually within
+# celerity-config.cmake instead.
+target_compile_definitions(celerity_runtime PUBLIC
+  # We still mark this as PUBLIC during builds (but not installation),
+  # so that the examples and tests receive the correct flag as well.
+  $<BUILD_INTERFACE:
+    $<$<CONFIG:Debug>:CELERITY_DETAIL_ENABLE_DEBUG>
+  >
+)
+
 if(CELERITY_SYCL_IMPL STREQUAL "ComputeCpp")
 	target_link_libraries(celerity_runtime PUBLIC ComputeCpp::ComputeCpp)
 endif()
--- a/cmake/celerity-config.cmake.in
+++ b/cmake/celerity-config.cmake.in
@@ -19,6 +19,39 @@ include("${CMAKE_CURRENT_LIST_DIR}/../celerity/vendor/cmake/spdlogConfig.cmake")
 
 include("${CMAKE_CURRENT_LIST_DIR}/celerity-targets.cmake")
 
+# We currently assume that only a single Celerity configuration is installed at a time.
+# Unfortunately CMake doesn't remove old configurations when re-installing a target,
+# so we advise to do a clean install (although it would suffice to just remove the
+# configuration-specific target files).
+# NOTE: This assumption is required for the CELERITY_DETAIL_ENABLE_DEBUG mechanism below.
+get_target_property(
+  CELERITY_IMPORTED_CONFIGS
+  Celerity::celerity_runtime
+  IMPORTED_CONFIGURATIONS
+)
+list(LENGTH CELERITY_IMPORTED_CONFIGS CELERITY_CONFIG_COUNT)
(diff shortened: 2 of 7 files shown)


fknorr (approved):
Looks good to me!

PR #21, opened 2020-12-11
Silence tons of warnings from CMake 3.19

fodinabor:
Since CMake 3.19 added a new [policy](https://cmake.org/cmake/help/latest/policy/CMP0110.html) for CTest names and Catch2 exploiting a somewhat edgy behavior beforehand, CMake spills a looot of warnings about this.. Explicitly setting the policy for CMake 3.19+ silences this. The update of Catch2 to the latest 2.x version (2.13.3) is necessary, as in this Catch2 version, the compatibility with the new behavior is introduced.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,3 +1,7 @@
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19.0") 
+    cmake_policy(SET CMP0110 NEW) # don't spam about add_test names..
+endif()
+
 include("${PROJECT_SOURCE_DIR}/vendor/Catch2/contrib/ParseAndAddCatchTests.cmake")
 
 # Add includes to library so they show up in IDEs


psalz:
Cool, thanks! Forgive me if I don't merge this right away; I'd like to take this opportunity to revisit our CI setup, so that we can have CI run on external PRs as well :-).

fodinabor:
I'm in no hurry with this, so please do so :)

psalz:
Alright, CI should now be triggered on external PRs as well! Would you be so kind as to force push this branch so we can give it a try? (It should suffice to just do a `git commit --amend` without any changes).

fodinabor:
I did a rebase yesterday, which I just pushed.. should suffice as well I guess :)

psalz:
Okay, CI obviously didn't quite work out. First, I'm not sure why this triggered a `pull_request` instead of a `pull_request_target`. Secondly, it seems like the latter doesn't work as I had thought, as ["runs in the context of the base repository"](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) not just means the workflow of the base repository, but also the code -- making it completely useless for our purposes.

Anyway, the tests went through, so I'll merge this now and deal with the remaining CI issues some other time ;-). Thank you!

PR #23, opened 2021-03-25
Remove dependency on Boost

psalz:
Ever since we stopped using Boost.Graph to represent task and command graphs (which was a while ago at this point), Boost has actually caused us more pain than gain. In particular, Boost does not (yet) distinguish between CUDA code being compiled by NVCC and Clang, instead always assuming the former, which causes all kinds of horrible, hard to debug problems.

As it turns out, we already only actually required Boost functionalities in a few select locations, and these uses have all been replaced by (mostly) equivalent vanilla implementations.

This patch includes a breaking change in the (semi-experimental) `host_memory_layout` interface: Instead of returning a `boost::static_vector` of `dimension` objects, the dimensions can now be accessed directly through `host_memory_layout::operator[]`.

Supersedes #15.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -80,10 +80,25 @@ class host_memory_layout {
 		size_t extent{};
 	};
 
-	/** Since contiguous dimensions can be merged when generating the memory layout, host_memory_layout is not generic over a fixed dimension count */
-	constexpr static size_t max_dimensionality = 4;
+	// TODO: This is a temporary replacement for Boost's static_vector; entire mechanism needs more comprehensive API overhaul.
+	class dimension_vector {
+	  public:
+		dimension_vector(size_t size) : this_size(size) {}
 
-	using dimension_vector = boost::container::static_vector<dimension, max_dimensionality>;
+		dimension& operator[](size_t idx) { return values[idx]; }
+		const dimension& operator[](size_t idx) const { return values[idx]; }
+
+		size_t size() const { return this_size; }
+
+	  private:
+		/**
+		 * Since contiguous dimensions can be merged when generating the memory layout, host_memory_layout is not generic over a fixed dimension count
+		 * TODO: Implement this ^
+		 */
+		constexpr static size_t max_dimensionality = 4;
+		std::array<dimension, max_dimensionality> values;
+		size_t this_size;
+	};
 
 	explicit host_memory_layout(const dimension_vector& dimensions) : dimensions(dimensions) {}
 
--- a/include/graph_generator.h
+++ b/include/graph_generator.h
@@ -19,6 +19,20 @@ namespace detail {
 	class abstract_command;
 	class horizon_command;
 
+	// TODO: Move to utility header..?
+	// Implementation from Boost.ContainerHash, licensed under the Boost Software License, Version 1.0.
+	inline void hash_combine(std::size_t& seed, std::size_t value) { seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2); }
+
+	struct pair_hash {
+		template <typename U, typename V>
+		std::size_t operator()(const std::pair<U, V>& p) const {
+			std::size_t seed = 0;
+			hash_combine(seed, std::hash<U>{}(p.first));
+			hash_combine(seed, std::hash<V>{}(p.second));
+			return seed;
+		}
+	};
+
 	class graph_generator {
 		friend struct graph_generator_testspy;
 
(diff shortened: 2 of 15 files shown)


include/accessor.h
```diff
@@ -51,6 +56,10 @@ class host_memory_layout {
 	 * - A two-dimensional layout is modelled as an interval of rows, which manifests as an offset (a multiple of the row width) and a stride (the row width
 	 * itself). Each row (aka 1-dimensional hyperplane) is modelled by the same one-dimensional layout.
 	 * - and so on for arbitrary dimensioned layouts.
+	 *
+	 *
+	 * Since contiguous dimensions can be merged when generating the memory layout, host_memory_layout is not generic over a fixed dimension count.
+	 * TODO: Implement merging of contiguous dimensions
 	 */
```
fknorr, line 63: We should improve the naming here. `dimension` has never really been accurate since we want to decouple buffer dimensionality from the number of strides by merging contiguous dimensions in the layout. Maybe `class stride` is a better option?

include/accessor.h
```diff
@@ -80,20 +89,27 @@ class host_memory_layout {
 		size_t extent{};
 	};
 
-	/** Since contiguous dimensions can be merged when generating the memory layout, host_memory_layout is not generic over a fixed dimension count */
-	constexpr static size_t max_dimensionality = 4;
-
-	using dimension_vector = boost::container::static_vector<dimension, max_dimensionality>;
-
-	explicit host_memory_layout(const dimension_vector& dimensions) : dimensions(dimensions) {}
+	size_t get_dimensions() const { return dimensions.size(); }
```
fknorr, line 92: This should be `get_num_dimensions` or `get_num_strides` depending on how we want to name the type (it returns a count rather than a sequence)

include/accessor.h
```diff
@@ -80,20 +89,27 @@ class host_memory_layout {
...
 
 	/** The layout maps per dimension, in descending dimensionality */
```
fknorr, line 94: Here's another mention of dimensionality that should probably go.

fknorr (commented):
Otherwise, looks good to me!

include/graph_generator.h
```diff
@@ -19,6 +19,14 @@ namespace detail {
 	class abstract_command;
 	class horizon_command;
 
+	// TODO: Move to utility header..?
+	struct pair_hash {
+		template <typename U, typename V>
+		std::size_t operator()(const std::pair<U, V>& p) const {
+			return p.first ^ p.second;
```
facuMH, line 26: XOR of 2 values doesn't seem like that good a hash, maybe you intended something else ?

test/runtime_tests.cc
```diff
@@ -356,6 +356,23 @@ namespace detail {
 		REQUIRE(is_safe_cgf<decltype(safe)>);
 	}
 
+	namespace foo {
```
facuMH, line 359: is this name really representative of the namespace and it functionality/purpose ?

vendor/ctpl_stl.h
```diff
@@ -138,8 +159,8 @@ namespace ctpl {
                 this->cv.notify_all();  // stop all waiting threads
             }
             for (int i = 0; i < static_cast<int>(this->threads.size()); ++i) {  // wait for the computing threads to finish
-                if (this->threads[i]->joinable())
-                    this->threads[i]->join();
+                    if (this->threads[i]->joinable())
```
facuMH, line 162: why the extra identation?

include/accessor.h
psalz: I wasn't happy with having a breaking API change in the first place so I've decided to revert this to its original form, and I propose we move this discussion to a dedicated issue / PR.

include/graph_generator.h
psalz: Thanks! You are right of course, I meant to combine the two hashes, not the values. I've changed this to use a re-implementation of Boost's `hash_combine`, which seems to be an accepted "good way" of doing it

test/runtime_tests.cc
psalz: I would say that `foo` is indeed a good way of indicating that this is a test name without any additional purpose. I am however open to suggestions ;-).

vendor/ctpl_stl.h
psalz: Good question. This is vendor code, so we're leaving its original formatting!

psalz:
@fknorr Can you please check whether the `distr_io` example still works properly with the reverted API?

fknorr:
It does!

fknorr (approved):

PR #28, opened 2021-05-10
Improve config parsing, in particular for CELERITY_DEVICES

psalz:
- In `parse_uint`: Check whether anything was parsed and if not, indicate that parsing failed (previously it would simply return 0).
- In `get_env`: Trim whitespace before returning values.
- CELERITY_DEVICES is now ignored when set to an empty value.
- Fix a bug in the exception message for when CELERITY_DEVICES does not   provide enough devices for the current run configuration.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -3,6 +3,7 @@
 #include <cstdlib>
 #include <iterator>
 #include <sstream>
+#include <string_view>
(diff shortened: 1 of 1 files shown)


facuMH (approved):
looks ok

PR #29, opened 2021-05-10
Fix concurrent CI workflow runs

psalz:
By default, the CI checkout step cleans the git workspace, removing all untracked files. This causes problems with concurrent workflow runs, as build directories can be deleted while tests are still running. An instance of this problem can be seen e.g. in #77, where the run for the `push` trigger is seemingly interrupted by the second run, triggered by `pull_request`.

Since we only work within uniquely named build directories, and clean up all artifacts ourselves, I think we can set this option to false.


psalz:
Fixed two bugs: Ensure `cleanup` runs after `report` (since we are deleting the downloaded build logs), and specify the actual command for deletion (`rm -f`, before it was just the filename ...).

facuMH (approved):
for now I agree this might be a good solution

PR #31, opened 2021-05-10
Run ComputeCpp CI on host device (instead of PTX)

psalz:
Run ComputeCpp tests on host device instead of NVIDIA for now (until we have Intel hardware to test on), as the PTX backend is no longer officially supported.


psalz:
Looks like this needs more work, running matmul on host device for debug builds is *SLOW* (I just cancelled it after 15 minutes).

psalz:
Superseded by #32.

PR #32, opened 2021-05-27
Add Intel CI platform support, run CCPP on Intel

psalz:
Summary:

- Add support for second self-hosted actions runner on Intel hardware
- Run ComputeCpp only on Intel going forward
- Run hipSYCL only on NVIDIA (for now)
- Upgrade clang-format version used in CI to 10

Context:

This reworks the CI setup once again, this time DRYing up the matrix build mechanism considerably. We now support two types of self-hosted runners, which are labelled as either "intel" or "nvidia".

As it turns out, different jobs from the same workflow run can be executed on different runners. This has never occurred so far as we only had a single runner. Therefore it is also not safe to assume that side-effects from predecessor ("needs") jobs still exist or ever existed on the current runner -- which includes things like build directories.

The canonical way of sharing artifacts between jobs is to upload them in one job and then download them again in the next. As this is rather inefficient and wasteful for large artifacts such as entire builds, we instead opt to combine building and testing into a single job. While this change somewhat decreases the granularity of reporting, it makes the workflow definition a lot cleaner and should fix any remaining inter-job race conditions encountered in the past (such as addressed in #29).

As part of upgrading the CI infrastructure, we now use clang-format-10, which does not produce any formatting changes for the codebase.

### DIFF
--- a/.github/workflows/celerity_ci.yml
+++ b/.github/workflows/celerity_ci.yml
@@ -21,29 +21,25 @@ on:
       - '**:**'
 
 jobs:
-  checkout:
-    runs-on: [self-hosted]
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          # We clean up artifacts ourselves.
-          # Cleaning here can cause problems with concurrent workflow runs,
-          # where build folders are being removed mid-run.
-          clean: false
-          submodules: true
-
-  build:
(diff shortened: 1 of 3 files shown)


facuMH (approved):
seems good

PR #34, opened 2021-06-02
Don't read uninitialized memory in buffer_manager tests

psalz:
Some of the tests for the `buffer_manager` try to determine whether certain superfluous coherence transfers have NOT been performed. To do so, these tests used to read uninitialized memory, checking that the superflouous data was not there.

While this was never a great idea, the thinking was that it would be highly unlikely to encounter the expected values in uninitialized memory. While this has worked so far for dedicated GPUs, as it turns out, this is not as unlikely for integrated GPUs that share RAM with the CPU.

This change introduces a special test mode for the `buffer_manager` that, when enabled, initializes all newly allocated buffers with a known bit pattern. We can then check for this bit pattern to establish that no coherence transfers have taken place.

### DIFF
--- a/include/buffer_manager.h
+++ b/include/buffer_manager.h
@@ -224,7 +224,6 @@ namespace detail {
 
 			audit_buffer_access(bid, new_buffer.is_allocated(), mode);
 
-#ifdef CELERITY_TEST
 			if(test_mode && new_buffer.is_allocated()) {
 				auto device_buf = static_cast<device_buffer_storage<DataT, Dims>*>(new_buffer.storage.get())->get_device_buffer();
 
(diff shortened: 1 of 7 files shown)


facuMH (approved):
seems good

include/buffer_manager.h
```diff
@@ -418,6 +453,20 @@ namespace detail {
...
+		 * @brief Enables test mode, ensuring that newly allocated buffers are always initialized to
+		 *        a known bit pattern, see buffer_manager::test_mode_pattern.
+		 */
+		static void enable_test_mode() { test_mode = true; }
+
+		static constexpr unsigned char test_mode_pattern = 0b10101010;
+
+#endif
+	  private:
+		inline static bool test_mode = false;
```
fknorr, line 469: This should probably be part of the `CELERITY_TEST` guard.

fknorr (approved):

Issue #16, opened 2020-08-06
Implement chained subscript operator on Celerity accessors


n-io:
Hello,

I have the following piece of code:
```
QueueManager::getInstance().with_master_access([&](celerity::handler& cgh) {
      auto result = output_buf.template get_access<cl::sycl::access::mode::read>(cgh, cl::sycl::range<2>(args.problem_size, args.problem_size));
      
      for (size_t i = 0; i < args.problem_size; i++) {
        for (size_t j = 0; j < args.problem_size; j++) {
          output[i*args.problem_size + j] = result[i][j];
        }
      }
```

It yields the following error (using a hipSYCL CPU-backend):

```
/p/project/celerity/code/celerity-bench/single-kernel/sobel.cc:108:51: error: no match for ‘operator[]’ (operand types are ‘celerity::accessor<hipsycl::sycl::vec<float, 4>, 2, (hipsycl::sycl::access::mode)1024, (hipsycl::sycl::access::target)2018>’ and ‘size_t’ {aka ‘long unsigned int’})
           output[i*args.problem_size + j] = result[i][j];
                                                   ^
```

Using a hipSYCL GPU-backend the error message appears similar:

```
/data/code/celerity-bench/single-kernel/sobel.cc:108:51: error: no viable overloaded operator[] for type 'celerity::accessor<hipsycl::sycl::vec<float, 4>, 2, hipsycl::sycl::access::mode::read, hipsycl::sycl::access::target::host_buffer>'
          output[i*args.problem_size + j] = result[i][j];
                                            ~~~~~~^~
```

The code has not changed since the last time I used it successfully on the master branch. Please could you have a look and advise?

Many thanks in advance!

psalz:
This is because we now return a custom `celerity::accessor` type from calls to `celerity::buffer::get_access`, as opposed to simply returning `cl::sycl::accessor`. Our custom accessor currently doesn't support the "chained subscript operator" for accessing elements, however you can simply use the overload taking a `cl::sycl::id` instead!

We should implement the chained operator at some point however. I've renamed the issue so we can keep track of this going forward.

PR #37, opened 2021-06-14
Feature/multidim subscriptor

facuMH:
Add multi dimensional subscriptor for host and device accessors
in response to this issue  #16.
Created an intermediate class to keep record of the other dimensions.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -46,6 +46,8 @@ namespace detail {
 	template <typename TagT>
 	constexpr cl::sycl::target deduce_access_target();
 
+	template <typename DataT, int Dims, cl::sycl::access::mode Mode, cl::sycl::target Target, int Index>
+	class accessor_subscript_proxy;
 } // namespace detail
 
 /**
@@ -204,6 +206,11 @@ class accessor<DataT, Dims, Mode, cl::sycl::target::device> : public detail::acc
 #pragma GCC diagnostic pop
 	}
 
+	template <int D = Dims>
+	std::enable_if_t<(D > 1), detail::accessor_subscript_proxy<DataT, D, Mode, cl::sycl::target::device, 1>> operator[](const size_t d0) const {
+		return {*this, d0};
+	}
+
 	friend bool operator==(const accessor& lhs, const accessor& rhs) {
 		return lhs.sycl_accessor == rhs.sycl_accessor && lhs.backing_buffer_offset == rhs.backing_buffer_offset;
 	}
@@ -336,6 +343,11 @@ class accessor<DataT, Dims, Mode, cl::sycl::target::host_buffer> : public detail
 	template <cl::sycl::access_mode M = Mode, int D = Dims>
 	std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M) && (D > 0), DataT> operator[](cl::sycl::id<Dims> index) const {
 		return *(get_buffer().get_pointer() + get_linear_offset(index));
+	}
+
+	template <int D = Dims>
+	std::enable_if_t<(D > 1), detail::accessor_subscript_proxy<DataT, D, Mode, cl::sycl::target::host_buffer, 1>> operator[](const size_t d0) const {
+		return {*this, d0};
 	}
 
 	/**
@@ -476,6 +488,47 @@ namespace detail {
 			return cl::sycl::target::host_buffer;
 		}
 	}
+
+
+	template <typename DataT, cl::sycl::access::mode Mode, cl::sycl::target Target>
+	class accessor_subscript_proxy<DataT, 3, Mode, Target, 2> {
+		using AccessorT = celerity::accessor<DataT, 3, Mode, Target>;
+
+	  public:
+		accessor_subscript_proxy(const AccessorT& acc, const size_t d0, const size_t d1) : acc(acc), d0(d0), d1(d1) {}
+
+		decltype(std::declval<AccessorT>()[{0, 0, 0}]) operator[](const size_t d2) const { return acc[{d0, d1, d2}]; }
+
+	  private:
+		const AccessorT& acc;
+		size_t d0;
+		size_t d1;
+	};
+
+	template <typename DataT, int Dims, cl::sycl::access::mode Mode, cl::sycl::target Target>
+	class accessor_subscript_proxy<DataT, Dims, Mode, Target, 1> {
+		template <int D>
+		using AccessorT = celerity::accessor<DataT, D, Mode, Target>;
+
+	  public:
+		accessor_subscript_proxy(const AccessorT<Dims>& acc, const size_t d0) : acc(acc), d0(d0) {}
+
+		// Note that we currently have to use SFINAE over constexpr-if + decltype(auto), as ComputeCpp 2.6.0 has
+		// problems inferring the correct type in some cases (e.g. when DataT == sycl::id<>).
+		template <int D = Dims>
+		std::enable_if_t<D == 2, decltype(std::declval<AccessorT<2>>()[{0, 0}])> operator[](const size_t d1) const {
+			return acc[{d0, d1}];
+		}
+
+		template <int D = Dims>
+		std::enable_if_t<D == 3, accessor_subscript_proxy<DataT, 3, Mode, Target, 2>> operator[](const size_t d1) const {
+			return {acc, d0, d1};
+		}
+
+	  private:
+		const AccessorT<Dims>& acc;
+		size_t d0;
+	};
 
 } // namespace detail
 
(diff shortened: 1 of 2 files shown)


include/accessor.h
```diff
@@ -477,6 +495,34 @@ namespace detail {
...
+	class multidim_subscriptor {
+		using AccessorT = celerity::accessor<DT, Dims, M, T>;
+
+	  public:
+		multidim_subscriptor(const AccessorT& acc, const size_t d0) : acc(acc), d0(d0), d1(4) {}
+
+		multidim_subscriptor(const AccessorT& acc, const size_t d0, const size_t d1) : acc(acc), d0(d0), d1(d1) {}
+
+		template <int D = Dims>
+		std::enable_if_t<Index != 0 && D == 3, multidim_subscriptor<DT, Dims, M, T, 0>> operator[](const size_t dx) const {
```
fknorr, line 508: Instead of those `enable_if` overloads, we should have three specializations of the class. This will not duplicate much code, but will avoid keeping `d1` around which is unused in the 2D case (and initialized to `4` for some reason?). Also, without that change, the constructors would require an `enable_if` as well for completeness.

test/runtime_tests.cc
```diff
@@ -1967,5 +1967,99 @@ namespace detail {
...
+		auto bid = bm.register_buffer<cl::sycl::id<3>, 3>(range);
+		access_target tgt = access_target::HOST;
+
+		SECTION("DEVICE ACCESSOR") {
+			auto& q = get_device_queue();
+			auto sr = subrange<3>({}, range_cast<3>(range));
+			live_pass_device_handler cgh(nullptr, sr, q);
+
+			auto acc3d = get_device_accessor<cl::sycl::id<3>, 3, cl::sycl::access::mode::read_write>(cgh, bid, range, {});
+			cgh.parallel_for<class test_3D_subscriptor>(range, [=](cl::sycl::item<3> item) {
```
fknorr, line 1985: Class names should be all-lowercase

test/runtime_tests.cc
```diff
@@ -1967,5 +1967,99 @@ namespace detail {
...
+		auto bid = bm.register_buffer<cl::sycl::id<2>, 2>(range3);
+		access_target tgt = access_target::HOST;
+
+		SECTION("DEVICE ACCESSOR") {
+			auto& q = get_device_queue();
+			auto sr = subrange<2>({}, range_cast<2>(range2));
+			live_pass_device_handler cgh(nullptr, sr, q);
+
+			auto acc2d = get_device_accessor<cl::sycl::id<2>, 2, cl::sycl::access::mode::read_write>(cgh, bid, range2, {});
+			cgh.parallel_for<class test_2D_subscriptor>(range2, [=](cl::sycl::item<2> item) {
```
fknorr, line 2032: Class names should be all-lowercase

test/runtime_tests.cc
```diff
@@ -1967,5 +1967,99 @@ namespace detail {
...
+			});
+			cgh.get_submission_event().wait();
+		}
+
+		SECTION("HOST ACCESSOR") {
+			distr_queue q;
+			q.submit([=](handler& cgh) {
+				auto acc2d = get_host_accessor<cl::sycl::id<2>, 2, cl::sycl::access::mode::read_write>(bid, range2, {});
+				cgh.host_task(on_master_node,
+				    [=] {
```
fknorr, line 2045: Did you run clang-format on this?

test/runtime_tests.cc
```diff
@@ -1967,5 +1967,99 @@ namespace detail {
...
+	TEST_CASE_METHOD(buffer_manager_fixture, "accessor multi-subscriptor for 2D", "[accessor]") {
+		size_t N = 2;
+		size_t N2 = 3;
+		cl::sycl::range range3 = cl::sycl::range<3>(N, N2, 1);
+		cl::sycl::range range2 = cl::sycl::range<2>(N, N2);
+		auto& bm = get_buffer_manager();
+		auto bid = bm.register_buffer<cl::sycl::id<2>, 2>(range3);
+		access_target tgt = access_target::HOST;
+
+		SECTION("DEVICE ACCESSOR") {
```
fknorr, line 2026: Why all-caps?

test/runtime_tests.cc
```diff
@@ -1967,5 +1967,99 @@ namespace detail {
...
+			auto acc2d = get_device_accessor<cl::sycl::id<2>, 2, cl::sycl::access::mode::read_write>(cgh, bid, range2, {});
+			cgh.parallel_for<class test_2D_subscriptor>(range2, [=](cl::sycl::item<2> item) {
+				size_t i = item[0];
+				size_t j = item[1];
+				acc2d[i][j] = item.get_id();
+			});
+			cgh.get_submission_event().wait();
+		}
+
+		SECTION("HOST ACCESSOR") {
```
fknorr, line 2040: Why all-caps?

fknorr (changes requested):

include/accessor.h
```diff
@@ -204,13 +206,21 @@ class accessor<DataT, Dims, Mode, cl::sycl::target::device> : public detail::acc
...
+
 	friend bool operator==(const accessor& lhs, const accessor& rhs) {
 		return lhs.sycl_accessor == rhs.sycl_accessor && lhs.backing_buffer_offset == rhs.backing_buffer_offset;
 	}
 
 	friend bool operator!=(const accessor& lhs, const accessor& rhs) { return !(lhs == rhs); }
 
   private:
+	template <typename DT, int D, cl::sycl::access::mode M, cl::sycl::target T, int Index>
+	friend class multidim_subscriptor;
```
psalz, line 222: Why does this have to be a `friend`?

include/accessor.h
```diff
@@ -392,6 +407,9 @@ class accessor<DataT, Dims, Mode, cl::sycl::target::host_buffer> : public detail
 	}
 
   private:
+	template <typename DT, int D, cl::sycl::access::mode M, cl::sycl::target T, int Index>
+	friend class multidim_subscriptor;
```
psalz, line 411: Why does this have to be a `friend`?

test/runtime_tests.cc
psalz: To add to that; try to have test case names + sections form (somewhat) complete sentences, so we get nice error messages like this:
image

psalz (changes requested):

test/runtime_tests.cc
facuMH: yes. it likes it this way.

fknorr (approved):
LGTM

include/accessor.h
psalz: I agree with this suggestion, however, we actually only need two specializations: One for the final 3D proxy (storing both `d0` and `d1`) as well as one for the final 2D / intermediate 3D proxy (storing only `d0`).

include/accessor.h
```diff
@@ -477,6 +489,51 @@ namespace detail {
...
+	};
+
+	template <typename DT, cl::sycl::access::mode M, cl::sycl::target T>
+	class accessor_subscript_proxy<DT, 3, M, T, 1> {
+		using AccessorT = celerity::accessor<DT, 3, M, T>;
+
+	  public:
+		accessor_subscript_proxy(const AccessorT& acc, const size_t d0) : acc(acc), d0(d0) {}
+
+		accessor_subscript_proxy<DT, 3, M, T, 0> operator[](const size_t dx) const { return accessor_subscript_proxy<DT, 3, M, T, 0>{acc, d0, dx}; }
```
psalz, line 515: As mentioned in response to @fknorr's comment, this or the specialization below is redundant and can be combined into a single specialization with `decltype(auto)` return type and `constexpr if` (or SFINAE, because of that CCPP compiler bug... in any case, we could get rid of a few lines of duplicated code).

include/accessor.h
```diff
@@ -477,6 +489,51 @@ namespace detail {
...
+
+
+	template <typename DT, cl::sycl::access::mode M, cl::sycl::target T>
+	class accessor_subscript_proxy<DT, 2, M, T, 1> {
+		using AccessorT = celerity::accessor<DT, 2, M, T>;
+
+	  public:
+		accessor_subscript_proxy(const AccessorT& acc, const size_t d0) : acc(acc), d0(d0) {}
+
+		DT& operator[](const size_t dx) const { return acc[{d0, dx}]; }
```
psalz, line 530: The return type is not always a reference -- only if using a writing access mode. So you'd need `decltype(auto)` (or again, SFINAE) here. Also, this case is evidently not covered by the tests currently!

include/accessor.h
```diff
@@ -477,6 +489,51 @@ namespace detail {
 		}
 	}
 
+
+	template <typename DT, cl::sycl::access::mode M, cl::sycl::target T>
+	class accessor_subscript_proxy<DT, 3, M, T, 0> {
```
psalz, line 494: Shouldn't the `Index` indicate which index will be set by this proxy? Because right now it's kind of the reverse, with the final proxy, which sets `d2` -- i.e., index 2, having `Index == 0`.

psalz (changes requested):

psalz (approved):

psalz:
LGTM

PR #38, opened 2021-07-27
Properly quote labels in print_graph

fknorr:
We previously generated invalid labels in dot graphs by omitting the surrounding quotes. There is currently no real way for the label string to contain quotes or backslashes, so we don't care about additional escaping.


psalz (approved):
LGTM!

PR #40, opened 2021-08-03
Revise accessor host_memory API

fknorr:
The host-accessor API `get_host_memory` was designed to collapse adjacent dimensions where possible in order to allow efficient non-strided operations from the user side whenever the allocation happens to match the range accessed by the kernel in one or more dimensions. However, this collapsing was never implemented, and it is also not particularly useful in conjunction with APIs like HDF5 that allow expressing multidimensional subranges natively. Until now, the `distr_io` abused this knowledge about the implementation to set up its HDF5 hyperslabs.

This patch deprecates `host_memory_layout` and `accessor::get_host_memory` in favor of the new `buffer_allocation_window` and `accessor::get_allocation_window`. This new API delivers an explicitly-dimensioned description of allocation vs. kernel subranges while excluding any collapsing semantics. These should be added in a later PR once an actual use case arises.

**TODO**: Bikeshed names `buffer_allocation_window` and `accessor::get_allocation_window`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ if(CELERITY_BUILD_EXAMPLES)
 
   find_package(PkgConfig)
   if (PkgConfig_FOUND)
-	pkg_check_modules(HDF5 IMPORTED_TARGET hdf5-1.12.0)
+	pkg_search_module(HDF5 IMPORTED_TARGET hdf5-openmpi hdf5-1.12.0 hdf5)
 	if(HDF5_FOUND)
       add_subdirectory(examples/distr_io)
       set_property(TARGET distr_io PROPERTY FOLDER "examples")
--- a/docs/host-tasks.md
+++ b/docs/host-tasks.md
@@ -140,9 +140,9 @@ Collective host tasks are special in that they receive an implicit one-dimension
 the participating nodes. To access buffers in a meaningful way, these node indices must be translated to buffer regions.
 In the typical Celerity fashion, this is handled via range mappers.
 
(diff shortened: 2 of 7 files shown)


facuMH (approved):

fknorr:
Rebased.

facuMH (commented):
is there a specific reason for which you did not try a test case with the new accessor initialization instead of buffer.get_access ?

psalz (commented):
Sorry for the delay -- looks good to me!

I agree that we may have to do some bikeshedding regarding the naming (cc @PeterTh). My main concern is that so far we've used the terms "virtual buffer" for "buffer" and "backing buffer" for "allocation" (and we didn't really have a term for "window", I think) -- although I'm not sure whether we ever used them in public documentation. Also I do like "window" and in particular "allocation" because I think (hope) most people will have an intuitive understanding of what they mean. I'm wondering whether it is clear what "buffer" refers to in this context, though.

fknorr:
I don't even advocate for changing the internal names (backing buffer / virtual buffer), the public API should expose more abstract names if it increases readability.

If the only other use of "buffer" in the API is `celerity::buffer`, that name should also be unambiguous.

`window` is not a thing yet however. I am hesitant to call it a `subrange`; that name would not be very descriptive as the layout exposes multiple subranges. `slice`, `span` or `view` would also work IMO.

PeterTh:
I like `buffer_allocation_window`.

fknorr:
@facuMH good point, I changed the test in question to use 2020 style accessor initialization.

psalz (approved):
Okay then let's ship it!

PR #33, opened 2021-05-31
Add support for SYCL 2020 accessor constructors along with new enums

facuMH:
Since Celerity supports many implementations of SYCL and they each have a different degree of support for the SYCL 2020 functionalities, in this first approach I intend to introduce to our users a new option for instantiating accessors, using the new Tags to specify Mode and Target.

A compatibility layer had to be implemented for ComputeCpp since they have not yet implemented some enums and types needed to provide this new functionalities to our users.

PR #42, opened 2021-08-18
Fix SYCL accessor subrange calculation

fknorr:
The offset and ranges we used to construct SYCL accessors were incorrect since #33. This did not manifest in the tests as the SYCL implementations we test with do not assert their correctness. This PR corrects these ranges and adds a regression test.


psalz:
LGTM!

psalz (approved):

PR #43, opened 2021-08-19
Use a binary output format for wave_sim example

PeterTh:
Uses a compact binary format and simplifies visualization script to generate animation in one step.


psalz:
Thank you, this is great!

A few points:
- This *should've* failed CI, because the [name of the artifact](https://github.com/celerity/celerity-runtime/blob/c6444b1909f04f061ea2fb81484298e0a87d96e9/ci/run-integration-tests.sh#L30) produced has changed. However unfortunately, our CI script silently ignored this. I've fixed this in #44. Would you be so kind as to rebase on top of that and adjust the name in our CI script?
- The wave_sim example comes with a [README](https://github.com/celerity/celerity-runtime/blob/c6444b1909f04f061ea2fb81484298e0a87d96e9/examples/wave_sim/README.md) that would also need adjustment now.
- The script currently produces a confusing error when no file name is provided. Maybe you could print a usage message instead? (Or should we just hardcode the filename..?)

Lastly, I've done some measurements and it appears that a majority of the time is spent not on plotting, but rather on producing the animation GIF. As discussed yesterday, I think this would be a good opportunity to ditch ImageMagick and instead switch to ffmpeg for producing the animation. I've pieced together the following command that seems to produce good results (and is blazingly fast compared to making the GIF):
```
ffmpeg -framerate 60 -pattern_type glob -i '/tmp/wave_sim_plot_*.png' -c:v libx264 -crf 18 -pix_fmt yuv420p wave_sim.mp4
```
However, I'm not sure how to best integrate it into the script. Could you take a stab at it?

PeterTh:
I took a stab at it.

psalz (approved):
LGTM!

PR #44, opened 2021-08-20
Make run-integration-tests.sh more robust and readable

psalz:
Previously a missing artifact would be silently ignored. By setting `-o errexit` the script now aborts if calculating the MD5 checksum
fails.

Properly quote variables where appropriate, and improve readability by removing unneeded curly braces.


fknorr (approved):
Consider writing all log / error output to stderr. Otherwise, LGTM

PR #46, opened 2021-09-04
Replace sycl::item with celerity::item in public API

fknorr:
Re-using the SYCL item as we do currently has several issues:

- Kernel offsets are deprecated, so we will have to add a celerity item for transparently including per-chunk offsets at some point anyway. This will first become visible with reductions, where the SYCL spec does not mandate a `handler::parallel_for` overload accepting both an offset and reduction variables (and at least DPC++ does not implement such an overload).
- `item.get_range()` reports the range of the current chunk, not the cluster-global size as the user might expect
- `item.get_linear_id()` is broken in the presence of offsets in DPC++

This patch introduces a new `celerity::item` (an `id`/`range` pair internally) which will report the cluster-global size as its range. The change is backwards compatible and still allows kernels accepting a `cl::sycl::item<>` that will continue to work in the same fashion as it does now.

### DIFF
--- a/examples/syncing/syncing.cc
+++ b/examples/syncing/syncing.cc
@@ -14,7 +14,7 @@ int main(int argc, char* argv[]) {
 
 	q.submit([=](handler& cgh) {
 		celerity::accessor b{buff, cgh, access::one_to_one{}, celerity::write_only, celerity::no_init};
-		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](celerity::item<1> item) { b[item] = item[0]; });
+		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](celerity::item<1> item) { b[item] = item.get_linear_id(); });
 	});
 
 	q.submit(celerity::allow_by_ref, [=, &host_buff](handler& cgh) {
--- a/examples/syncing/syncing.cc
+++ b/examples/syncing/syncing.cc
@@ -14,7 +14,7 @@ int main(int argc, char* argv[]) {
 
 	q.submit([=](handler& cgh) {
 		celerity::accessor b{buff, cgh, access::one_to_one{}, celerity::write_only, celerity::no_init};
-		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](cl::sycl::item<1> item) { b[item] = item[0]; });
+		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](celerity::item<1> item) { b[item] = item[0]; });
 	});
 
 	q.submit(celerity::allow_by_ref, [=, &host_buff](handler& cgh) {
--- a/include/handler.h
+++ b/include/handler.h
@@ -9,6 +9,7 @@
 
 #include "device_queue.h"
 #include "host_queue.h"
+#include "item.h"
 #include "range_mapper.h"
 #include "ranges.h"
 #include "task.h"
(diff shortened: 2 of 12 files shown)


include/handler.h
```diff
@@ -328,10 +329,29 @@ namespace detail {
 
 	// The current mechanism for hydrating the SYCL placeholder accessors inside Celerity accessors requires that the kernel functor
 	// capturing those accessors is copied at least once during submission (see also live_pass_device_handler::submit_to_sycl).
-	// As of SYCL 2020 kernel functors are passed as const references, so we have to do this manually here.
+	// As of SYCL 2020 kernel functors are passed as const references, so we explicitly copy or capture by reference in bind_kernel_with_*
```
psalz, line 332: You mean "capture by value"?

include/handler.h
```diff
@@ -328,10 +329,29 @@ namespace detail {
...
+	template <typename Functor, int Dims>
+	auto bind_kernel_with_celerity_item(const Functor& kernel, const cl::sycl::range<Dims>& global_range) {
+		return [=](cl::sycl::item<Dims> s_item) { kernel(make_item<Dims>(s_item.get_id(), global_range)); };
+	}
+
 	template <typename Functor>
-	Functor copy_kernel(const Functor& kernel) {
-		return Functor{kernel};
+	[[deprecated("Support for kernels receiving cl::sycl::item<Dims> will be removed in the future, change parameter type to celerity::item<Dims>")]] //
+	auto
```
psalz, line 341: For readability I'd prefer `Functor` here, underlining the fact that we are copying the kernel.

include/handler.h
```diff
@@ -328,10 +329,29 @@ namespace detail {
...
+	}
+
+	template <typename Functor, int Dims>
+	auto bind_kernel(const Functor& kernel, const cl::sycl::range<Dims>& global_range) {
+		if constexpr(std::is_invocable_v<Functor, celerity::item<Dims>>) {
+			return bind_kernel_with_celerity_item(kernel, global_range);
+		} else if constexpr(std::is_invocable_v<Functor, cl::sycl::item<Dims>>) {
+			return bind_kernel_with_sycl_item(kernel);
+		} else {
+			static_assert(constexpr_false<Functor>, "Kernel function must be invocable with celerity::item<Dims> or cl::sycl::item<Dims> (deprecated)");
```
psalz, line 353: Does this still allow for kernels to receive ` cl::sycl::id<Dims>`? (I.e., does `is_invocable_v` consider implicit conversions?).

psalz (commented):
Looks good. I'll need some more time to think about the broader implications though.

facuMH (approved):
lgtm

include/handler.h
fknorr: Yes! Fixed.

include/handler.h
fknorr: Done.

include/handler.h
fknorr: Yes, `is_invocable_v` checks if a call to `std::invoke` with those argument types would succeed, including any necessary implicit conversions, template instantiations, ...

include/item.h
```diff
@@ -0,0 +1,49 @@
...
+
+	operator cl::sycl::id<Dims>() const { return id; } // NOLINT(google-explicit-constructor)
+
+	size_t operator[](int dimension) const { return id[dimension]; }
+
+	cl::sycl::range<Dims> get_range() const { return range; }
+
+	size_t get_range(int dimension) const { return range[dimension]; }
+
+	size_t get_linear_id() const { return detail::get_linear_index(range, id); }
```
psalz, line 37: As just discussed in person, we should probably adjust the behavior of `get_linear_id` to that prescribed by SYCL 2020, i.e., subtracting the kernel offset before linearizing, such that the first element always has a linear id of 0.

include/item.h
```diff
@@ -0,0 +1,49 @@
...
+	template <int Dims>
+	item<Dims> make_item(cl::sycl::id<Dims> global_id, cl::sycl::range<Dims> global_range) {
+		return item<Dims>{global_id, global_range};
+	}
+
+} // namespace detail
+
+// We replace sycl::item with celerity::item to correctly expose the cluster global size instead of the chunk size to the user.
+template <int Dims>
+class item {
```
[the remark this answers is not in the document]
fknorr: Included, although they do not seem particularly useful. I don't believe you can really have multiple distinct item in a basic `parallel_for` kernel unless you move some in through a `buffer<optional<item>>`.

examples/syncing/syncing.cc
```diff
@@ -14,7 +14,7 @@ int main(int argc, char* argv[]) {
 
 	q.submit([=](handler& cgh) {
 		celerity::accessor b{buff, cgh, access::one_to_one{}, celerity::write_only, celerity::no_init};
-		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](cl::sycl::item<1> item) { b[item] = item[0]; });
+		cgh.parallel_for<class mat_mul>(cl::sycl::range<1>(N), [=](celerity::item<1> item) { b[item] = item.get_linear_id(); });
```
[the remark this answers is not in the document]
fknorr: Done.

include/item.h
fknorr: Adjusted, also added `get_offset()` from `sycl::item`. This now requires us to include an additional `sycl::id<Dims> global_offset`. I'm not entirely convinced we should keep global offsets around indefinitely since SYCL 2020 has now deprecated them.

include/item.h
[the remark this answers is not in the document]
psalz: You are right of course, I did not consider that... Well, at least we are closer to the spec ;-).

include/item.h
psalz: Agreed. And we probably don't need to wait until SYCL Next to do so (but we should at least have a grace period with deprecation warning). I'll add it to *the list*.

psalz (approved):
Sorry for the delay; LGTM!

PR #50, opened 2021-10-12
Make kernel names optional

fknorr:
In SYCL 2020, explicit device kernel names are optional. This is implemented in hipSYCL and DPC++ so far. With this PR, Celerity will expose that feature for both of these backends.

For ComputeCpp, kernel names are still required, but Celerity will now wrap them to suppress the `-Rsycl-kernel-naming` diagnostic for locally-declared types such as `parallel_for<class kernel>`, since this diagnostic triggers on idiomatic Celerity code and cannot be disabled otherwise.

### DIFF
--- a/include/handler.h
+++ b/include/handler.h
@@ -51,12 +51,36 @@ namespace detail {
 		return name.substr(0, name.length() - 1);
 	}
 
+	struct unnamed_kernel {};
+
+	template <typename KernelName>
+	constexpr bool is_unnamed_kernel = std::is_same_v<KernelName, unnamed_kernel>;
+
+#if WORKAROUND_COMPUTECPP
+	template <typename KernelName>
+	struct kernel_name_wrapper;
+#endif
+
+	template <typename KernelName>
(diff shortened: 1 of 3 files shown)


include/handler.h
```diff
@@ -63,13 +63,37 @@ namespace detail {
...
+#if WORKAROUND_COMPUTECPP
+	template <typename KernelName>
+	struct kernel_name_wrapper;
+#endif
+
+	template <typename KernelName>
+	struct bound_kernel_name {
+		static_assert(!is_unnamed_kernel<KernelName>);
+#if WORKAROUND_COMPUTECPP
+		using type = kernel_name_wrapper<KernelName>; // Suppress -Rsycl-kernel-naming diagnostic for local types
```
psalz, line 80: Praise the lord

psalz (changes requested):
`handler.h` is not correctly formatted, otherwise LGTM!

psalz (approved):

PR #51, opened 2021-10-28
Fix CMake installation for ComputeCpp and DPC++ backends

fknorr:
We previously did not test the installed version of Celerity in the CI, causing us to miss updating them when introducing DPC++ support.

This PR

- Ensures that `add_sycl_to_target` and `add_celerity_to_target` are available on installation for DPC++
- Installs our patched `FindComptueCpp.cmake`
- Allows building examples both as a CMake subdirectory like before and a standalone project, linking against installed Celerity
- In the CI, installs Celerity and then builds examples (integration tests) against the installed runtime

@n-io Can you please test if this allows you to work with Celerity + DPC++?


ci/run-integration-tests.sh
```diff
@@ -2,11 +2,11 @@
 
 set -o errexit -o pipefail -o nounset
 
-if [[ ! -d examples || ! -d CMakeFiles ]]; then
+if [[ ! -d CMakeFiles ]]; then
     echo "Warning: This script should be run from within a build directory" 1>&2
 fi
 
-if [[ $# -le 1 ]]; then
+if [[ $# -le 2 ]]; then
```
psalz, line 9: Why 2? `$#` does not include `$0`!

cmake/AddToTarget.cmake
```diff
@@ -0,0 +1,39 @@
...
+    set(one_value_args TARGET)
+    set(multi_value_args)
+    cmake_parse_arguments(ADD_SYCL
+      "${options}"
+      "${one_value_args}"
+      "${multi_value_args}"
+      ${ARGN}
+      )
+    # Wno-deprecated-declarations: Kernel offsets are used all over the place
+    set(DPCPP_FLAGS "-fsycl;-sycl-std=2020;-fsycl-targets=${CELERITY_DPCPP_TARGETS};-DCELERITY_DPCPP=1;-Wno-deprecated-declarations;${DPCPP_FLAGS}")
```
psalz, line 13: DPC++ supports a macro `SYCL2020_DISABLE_DEPRECATION_WARNINGS` - should we be using that instead..? 🤔

psalz (commented):
Thank you, a sorely needed improvement!

ci/run-integration-tests.sh
fknorr: Exactly, otherwise it would have to be `-lt 3` (I want to see at least one node-count)

cmake/AddToTarget.cmake
fknorr: Actually, we don't need to suppress warnings at all anymore (see the comment!)

ci/run-integration-tests.sh
psalz: Yes but this is `-le` not `-lt`

fknorr: Right!

cmake/AddToTarget.cmake
psalz: Even better!

psalz (approved):

n-io:
The package configuration file will now install to `cmake/lib/Celerity/celerity-config.cmake` instead of `cmake/lib/celerity-config.cmake`. This does not currently match the documentation of `readme.md` section 'Using Celerity as a Library'.

PR #52, opened 2021-11-04
Task Horizons

PeterTh:
This implements task horizons. Significant changes include:
- all nodes generate the task graph
- generation of task horizons is controlled by the task manager
- command graph horizons are generated based on and related to task horizons (no separate logic)
- horizon commands actually do something when executed (notify the task manager)
- task horizons are used to clean up the last writer structure and delete tasks


src/task_manager.cc
```diff
@@ -43,6 +44,16 @@ namespace detail {
 		}
 	}
 
+	void task_manager::notify_horizon_executed(task_id tid) {
```
PeterTh, line 47: I just noticed that this assumes it is never called from 2 threads in parallel. If that can happen, we'd need a lock for the `executed_horizons` data structure.

test/system/distr_tests.cc
```diff
@@ -132,5 +134,83 @@ namespace detail {
...
+			cgh.parallel_for<class UKN(produce_d)>(cl::sycl::range<1>{N}, [=](celerity::item<1> item) {
+				read_write_a[item] += read_write_b[item] + 1;
+				read_write_b[item] = read_write_a[item];
+			});
+		});
+
+		q.submit([=](handler& cgh) {
+			accessor write_a{buff_a, cgh, celerity::access::all{}, celerity::write_only, celerity::no_init};
+			cgh.parallel_for<class UKN(produce_e)>(
+			    cl::sycl::range<1>{N}, [=](celerity::item<1> item) { write_a[item] = static_cast<int>(item.get_linear_id()); });
```
psalz, line 174: Kernel names are confusing ("produce_e" => There is no `buff_e`); also there is really no point in actually doing anything in these kernels, is there? I suggest making them no-ops instead.

test/task_graph_tests.cc
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+
+	TEST_CASE("task_manager keeps track of max pseudo critical path length and task front", "[task_manager][task-graph][task-front]") {
+		using namespace cl::sycl::access;
+		task_manager tm{1, nullptr, nullptr};
+		test_utils::mock_buffer_factory mbf(&tm);
+		auto buf_a = mbf.create_buffer(cl::sycl::range<1>(128));
+
+		SECTION("with true dependencies") {
```
psalz, line 313: Should this also test anti- and order-dependencies?

test/task_graph_tests.cc
```diff
@@ -0,0 +1,448 @@
...
+
+	TEST_CASE("task horizons are being generate with correct dependencies", "[task_manager][task-graph][task-horizon]") {
+		task_manager tm{1, nullptr, nullptr};
+		tm.set_horizon_step(2);
+
+		using namespace cl::sycl::access;
+		test_utils::mock_buffer_factory mbf(&tm);
+		auto buf_a = mbf.create_buffer(cl::sycl::range<1>(128));
+
+		SECTION("with true dependencies") {
```
psalz, line 342: Should this also test anti- and order-dependencies?

test/task_graph_tests.cc
```diff
@@ -0,0 +1,448 @@
...
+			check_path_length_and_front(tm, 2, {tid_c});
+
+			const auto tid_d = test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) {});
+			check_path_length_and_front(tm, 2, {tid_c, tid_d});
+
+			maybe_print_graph(tm);
+		}
+	}
+
+	TEST_CASE("task horizons are being generate with correct dependencies", "[task_manager][task-graph][task-horizon]") {
```
psalz, line 334: Typo "generateD"

test/task_graph_tests.cc
```diff
@@ -0,0 +1,448 @@
...
+		auto* previous_horizon = task_manager_testspy::get_previous_horizon_task(tm);
+		CHECK(task_manager_testspy::get_num_horizons(tm) == 1);
+		CHECK(previous_horizon != nullptr);
+
+		task_id tid_6 = test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf_b.get_access<mode::read_write>(cgh, fixed<1>({0, 128})); });
+		task_id tid_7 = test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf_b.get_access<mode::read_write>(cgh, fixed<1>({0, 128})); });
+		task_id tid_8 = test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf_b.get_access<mode::read_write>(cgh, fixed<1>({0, 128})); });
+
+		{
+			INFO("check that previous tasks are still last writers before the first horizon is applied");
```
psalz, line 420: Nice pattern 👍

psalz (changes requested):
Thank you, looks very good! A few minor nitpicks:

src/task_manager.cc
psalz: That's fine, worker jobs all run in the same thread!

include/task_manager.h
```diff
@@ -109,10 +117,48 @@ namespace detail {
...
+		task* previous_horizon_task = nullptr;
+
+		// Queue of horizon tasks for which the associated commands were executed
+		std::queue<task_id> executed_horizons;
+		// marker task id for "nothing to delete" - we can safely use 0 here
+		static constexpr task_id nothing_to_delete = 0;
+		// task_id ready for deletion, 0 if nothing to delete (set on notify, used on new task creation)
+		std::atomic<task_id> horizon_task_id_for_deletion = nothing_to_delete;
+		// How many horizons to delay before deleting tasks
+		static constexpr int horizon_deletion_lag = 3;
```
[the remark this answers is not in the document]
PeterTh: It's not really intended to be configurable, but I'd rather not put a magic number in the code.

test/runtime_tests.cc
```diff
@@ -2436,5 +2437,42 @@ namespace detail {
 
 #endif // !WORKAROUND_COMPUTECPP && (!WORKAROUND_HIPSYCL || CELERITY_HIPSYCL_SUPPORTS_REDUCTIONS)
 
+	// This test case requires actual command execution, which is why it is not in graph_compaction_tests
```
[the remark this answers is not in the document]
PeterTh: Yes, but then I'm not actually testing a significant part of the functionality.

test/task_graph_tests.cc
PeterTh: I thought about it, but none of the horizon code ever looks at the *types* of dependencies, so it's probably not necessary. I'd keep it open for future expansion.

include/task_manager.h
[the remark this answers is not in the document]
BlackMark29A: It is not obvious to me why this must not be less than 3, so maybe a comment explaining the reason for that would be helpful. Also if this is the case, there probably should be a `static_assert` making sure of that.
Also, why is `task_horizon_step_size` initialized to 4?

src/task_manager.cc
```diff
@@ -143,16 +154,89 @@ namespace detail {
...
-				tsk->add_dependency({task_map.at(prev->second).get(), dependency_kind::ORDER_DEP});
+				add_dependency(tsk.get(), task_map.at(prev->second).get(), dependency_kind::ORDER_DEP);
 				last_collective_tasks.erase(prev);
 			}
 			last_collective_tasks.emplace(cgid, tid);
 		}
 	}
 
-	void task_manager::invoke_callbacks(task_id tid) {
+	task_id task_manager::get_new_tid() { return next_task_id++; }
```
BlackMark29A, line 164: It would probably make sense to move this into the header to allow the compiler to inline this. This would also be more consistent with other very simple one statement functions that are already defined in the header.

src/task_manager.cc
```diff
@@ -143,16 +154,89 @@ namespace detail {
...
+
+	void task_manager::add_dependency(task* depender, task* dependee, dependency_kind kind) {
+		assert(depender != dependee);
+		assert(depender != nullptr && dependee != nullptr);
+		depender->add_dependency({dependee, kind});
+		execution_front.erase(dependee);
+		max_pseudo_critical_path_length = std::max(max_pseudo_critical_path_length, depender->get_pseudo_critical_path_length());
+	}
+
+	bool task_manager::need_new_horizon() const { //
```
BlackMark29A, line 188: What's with the empty comment? And this could probably also be moved into the header.

include/task_manager.h
```diff
@@ -88,14 +89,22 @@ namespace detail {
 		 */
 		void shutdown() { task_map.clear(); }
 
+		void set_horizon_step(const int step) { task_horizon_step_size = step; }
```
BlackMark29A, line 92: I assume negative step sizes are semantically invalid, so this should probably have an `assert(step >= 0)` to document and sanity check the provided value.

test/system/distr_tests.cc
```diff
@@ -132,5 +134,83 @@ namespace detail {
 
 #endif // !WORKAROUND_COMPUTECPP && (!WORKAROUND_HIPSYCL || CELERITY_HIPSYCL_SUPPORTS_REDUCTIONS)
 
+	TEST_CASE("generating same task graph on different nodes", "[task-graph]") {
+		distr_queue q;
+		REQUIRE(runtime::get_instance().get_num_nodes() > 1);
+
+		const int N = 1000;
```
BlackMark29A, line 141: Is there a reason why this is not `constexpr`?

test/runtime_tests.cc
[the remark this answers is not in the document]
psalz: Fair enough!

test/task_graph_tests.cc
psalz: I see. In that case I'd advocate removing the section altogether, as it's existence seems to imply there *is* some significance to the type of dependencies.

include/task_manager.h
psalz: Agreed, in fact I think it must even be `> 1` (right?).

PeterTh: I don't think it necessarily has to be. In fact, one of the graph compaction tests tests step size 1.

test/task_graph_tests.cc
PeterTh: Agreed,

src/task_manager.cc
facuMH: the empty comment was to deal with the auto formatter doing weird formatting. but I will move it to into the header.

test/task_graph_tests.cc
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+
+	static inline GridRegion<3> make_region(int min, int max) { return GridRegion<3>(GridPoint<3>(min, 0, 0), GridPoint<3>(max, 1, 1)); }
+
+	TEST_CASE("task horizons update previous writer data structure", "[task_manager][task-graph][task-horizon]") {
+		task_manager tm{1, nullptr, nullptr};
+		tm.set_horizon_step(2);
+
+		using namespace cl::sycl::access;
```
psalz, line 398: Since we have `celerity::access_mode` now, we should prefer it over `cl::sycl::access::mode`!

psalz (changes requested):

psalz:
I just randomly stumbled upon this functionality, which I believe is now broken:

https://github.com/celerity/celerity-runtime/blob/14e0cefd26e623783eebdbfe78268971623f0e45/src/runtime.cc#L229-L235

That whole functionality is somewhat bogus though. I believe the original intent was for it to act as a safeguard against dangling reference-captured buffers inside CGFs. In any case, deleting tasks breaks this.

PR #53, opened 2021-11-11
Replace cl::sycl::{id,range} with celerity::{id,range} alias

fknorr:
This aliases `id` and `range` in the Celerity namespace for namespacing consistency with `celerity::nd_range`.
Also re-exports `celerity::property_list` and `celerity::property::reduction::initialize_to_identity` (on SYCL backends that support this property).

Internal APIs are not refactored to avoid an unnecessarily large change, since we never plan to actually re-implement those types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,14 @@ Celerity without much hassle. If you know SYCL already, this will probably
 look very familiar to you:
 
 ```cpp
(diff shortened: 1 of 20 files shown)


psalz (approved):
Other than the unformatted files - LGTM!

PR #56, opened 2021-11-15
Add DPC++ version string to runtime

fknorr:
We were printing "unknown" as the SYCL implementation when building against DPC++.

### DIFF
--- a/src/runtime.cc
+++ b/src/runtime.cc
@@ -1,7 +1,6 @@
 #include "runtime.h"
 
 #include <queue>
-#include <sstream>
 #include <string>
 #include <unordered_map>
 
@@ -24,6 +23,9 @@
 #include "scheduler.h"
 #include "task_manager.h"
 #include "user_bench.h"
+
+#define CELERITY_STRINGIFY2(x) #x
+#define CELERITY_STRINGIFY(x) CELERITY_STRINGIFY2(x)
 
 namespace celerity {
 namespace detail {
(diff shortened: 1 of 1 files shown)


src/runtime.cc
```diff
@@ -60,16 +62,16 @@ namespace detail {
 #endif
 	}
 
-	std::string get_sycl_version() {
-		std::stringstream ss;
+	const char* get_sycl_version() {
 #if defined(__COMPUTECPP__)
-		ss << "ComputeCpp " << COMPUTECPP_VERSION_MAJOR << "." << COMPUTECPP_VERSION_MINOR << "." << COMPUTECPP_VERSION_PATCH;
-		return ss.str();
+		return "ComputeCpp " CELERITY_STRINGIFY(COMPUTECPP_VERSION_MAJOR) "." CELERITY_STRINGIFY(COMPUTECPP_VERSION_MINOR) //
```
psalz, line 67: I mean.... We do have access `fmt::format` as well ;-).

psalz (approved):
LGTM!

src/runtime.cc
PeterTh: But this is compile time! Please consider the highly significant performance gains in building the version string.

src/runtime.cc
```diff
@@ -60,16 +62,16 @@ namespace detail {
...
+		    "." CELERITY_STRINGIFY(COMPUTECPP_VERSION_PATCH);
 #elif defined(__HIPSYCL__) || defined(__HIPSYCL_TRANSFORM__)
-		ss << "hipSYCL " << HIPSYCL_VERSION_MAJOR << "." << HIPSYCL_VERSION_MINOR << "." << HIPSYCL_VERSION_PATCH;
-		return ss.str();
+		return "hipSYCL " CELERITY_STRINGIFY(HIPSYCL_VERSION_MAJOR) "." CELERITY_STRINGIFY(HIPSYCL_VERSION_MINOR) "." CELERITY_STRINGIFY(HIPSYCL_VERSION_PATCH);
+#elif CELERITY_DPCPP
+		return "DPC++ / Clang " __clang_version__;
 #else
-		return "Unknown";
+#error "unknown SYCL implementation"
```
PeterTh, line 74: I think this is a noteworthy behavior change. This commit will prevent Celerity from compiling on unknown SYCL platforms, while previously it would.

Do we want this?

psalz: Yes we discussed this offline; the reason why this bug went unnoticed is because we fell back to `"Unknown"` here. This was an idealistic decision made back in the day, when in reality you cannot compile Celerity against a new SYCL implementation without explicitly supporting it in our CMake infrastructure first.

PeterTh: Yeah, I agree with this point. I wanted to note it down though since it's a somewhat significant change that might not be implied by the commit/pull request names.

PeterTh (approved):
LGTM

src/runtime.cc
fknorr: I honestly forgot about `fmt` when doing this :)

PR #58, opened 2021-11-16
Expose SYCL 2020 ::sycl namespace to users

psalz:
This was already the case for ComputeCPP and DPC++, however hipSYCL requires an explicit alias.

### DIFF
--- a/include/celerity.h
+++ b/include/celerity.h
@@ -24,4 +24,8 @@ namespace runtime {
 } // namespace runtime
 } // namespace celerity
 
+// Celerity includes <CL/sycl.hpp> internally, but we want to expose the SYCL 2020 ::sycl namespace to Celerity users.
+// TODO: Remove this once Celerity includes <sycl/sycl.hpp> internally.
+#include <sycl/sycl.hpp>
+
 #endif


include/celerity.h
```diff
@@ -24,4 +24,14 @@ namespace runtime {
 } // namespace runtime
 } // namespace celerity
 
+// Expose SYCL 2020 ::sycl namespace to Celerity users.
+// Implementations currently don't agree on whether the legacy include path <CL/sycl.hpp> should also
+// expose the new namespace. See https://github.com/KhronosGroup/SYCL-Docs/issues/137.
+// TODO: Remove this once Celerity includes <sycl/sycl.hpp> internally.
+#if WORKAROUND_HIPSYCL
+namespace sycl = cl::sycl;
```
fknorr, line 32: I think we should really `#include <SYCL/sycl.hpp>` here instead.
When declaring a namespace alias as it's done here, it is illegal to re-open the namespace with `namespace sycl { ... }` afterwards, as a user might want to do to specialize trait types, for example.

fknorr (changes requested):

include/celerity.h
psalz: You're right, that makes more sense! Fixed.

fknorr (approved):
LGTM!

PR #59, opened 2021-11-16
Prefer selecting any available device over aborting

psalz:
...in case not enough/no GPU device(s) can be found.

Our previous approach was a bit harsh in that unless `CELERITY_DEVICES` was set, the runtime would abort if it could not find at least one GPU device. For toying with the API it can be useful to run on a CPU or even the host device, so we should enable that use case by default.

The new selection strategy is as follows:

1. Try to find a platform that can provide a unique GPU for each worker node
2. Try to find a platform that can provide a unique device of any type for each worker node
3. Try to find any available GPU (which will be shared by all worker nodes)
4. Try to find any available device (which will be shared by all worker nodes)
5. Abort

### DIFF
--- a/src/device_queue.cc
+++ b/src/device_queue.cc
@@ -48,26 +48,42 @@ namespace detail {
 			} else {
 				const auto host_cfg = cfg.get_host_config();
 
-				// Try to find a platform that can provide a unique device for each node.
-				bool found = false;
-				const auto platforms = cl::sycl::platform::get_platforms();
-				for(size_t i = 0; i < platforms.size(); ++i) {
-					auto&& platform = platforms[i];
-					const auto devices = platform.get_devices(cl::sycl::info::device_type::gpu);
-					if(devices.size() >= host_cfg.node_count) {
-						how_selected = fmt::format("automatically selected platform {}, device {}", i, host_cfg.local_rank);
-						device = devices[host_cfg.local_rank];
-						found = true;
-						break;
+				const auto try_find_device_per_node = [&host_cfg, &device, &how_selected](cl::sycl::info::device_type type) {
+					// Try to find a platform that can provide a unique device for each node.
+					const auto platforms = cl::sycl::platform::get_platforms();
+					for(size_t i = 0; i < platforms.size(); ++i) {
+						auto&& platform = platforms[i];
+						const auto devices = platform.get_devices(type);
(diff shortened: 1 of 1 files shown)


src/device_queue.cc
```diff
@@ -48,26 +48,42 @@ namespace detail {
...
-				const auto platforms = cl::sycl::platform::get_platforms();
-				for(size_t i = 0; i < platforms.size(); ++i) {
-					auto&& platform = platforms[i];
-					const auto devices = platform.get_devices(cl::sycl::info::device_type::gpu);
-					if(devices.size() >= host_cfg.node_count) {
-						how_selected = fmt::format("automatically selected platform {}, device {}", i, host_cfg.local_rank);
-						device = devices[host_cfg.local_rank];
-						found = true;
-						break;
+				const auto try_find_device_per_node = [&host_cfg, &device, &how_selected](cl::sycl::info::device_type type) {
```
PeterTh, line 51: I think it would be a bit nicer to factor this and the subsequent lambda out as separate functions (since the surrounding function is already quite huge), but it's not really necessary and maybe more in the scope of a refactoring PR.

src/device_queue.cc
```diff
@@ -48,26 +48,42 @@ namespace detail {
...
+					return false;
+				};
 
-				if(!found) {
-					queue_logger.warn("No suitable platform found that can provide {} devices, and CELERITY_DEVICES not set", host_cfg.node_count);
-					// Just use the first device available
-					const auto devices = cl::sycl::device::get_devices(cl::sycl::info::device_type::gpu);
-					if(devices.empty()) { throw std::runtime_error("Automatic device selection failed: No GPU device available"); }
-					device = devices[0];
+				const auto try_find_one_device = [&device](cl::sycl::info::device_type type) {
```
PeterTh, line 66: This path does not set `how_selected`.

src/device_queue.cc
psalz: Yes, let's do it in a separate change (I've added it to the refactoring wishlist).

src/device_queue.cc
psalz: `how_selected` is initialized to `"automatically selected"` on line 29 (yes not ideal, we'll address it during refactoring ;-)).

PeterTh (approved):

PR #62, opened 2021-12-02
Microbenchmarks - work in progress

PeterTh:
Work in progress open for discussion.

I currently vendored the args library directly into the benchmarks folder, which is probably not what we want. I do however expect these benchmarks to potentially have a lot of arguments, and I don't want to parse them ad hoc. I also saw that we already do just that in several of the `examples`, but it might be a bit tricky to re-use a library across both that and the benchmarks when both are designed to be able to be built either independently or as part of the overall tree. I'm open for suggestions on how to best do this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "vendor/Catch2"]
 	path = vendor/Catch2
 	url = https://github.com/catchorg/Catch2
+[submodule "vendor/args"]
+	path = vendor/args
+	url = https://github.com/Taywee/args.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,11 @@ endmacro()
 add_submodule_directory(vendor/spdlog)
 add_submodule_directory(vendor/Catch2)
 
+# we do not want to build the args example or tests
+set(ARGS_BUILD_EXAMPLE OFF CACHE INTERNAL "")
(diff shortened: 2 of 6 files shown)


psalz:
Catch2 [ships with Clara](https://github.com/catchorg/Catch2/blob/devel/docs/own-main.md#adding-your-own-command-line-options), a small command line parsing library. Although it seems the standalone project [has been discontinued](https://github.com/catchorg/Clara/blob/master/README.md), it is still alive and well as part of the Catch2 distribution. Could we use that instead of args.hxx?

PeterTh:
After some offline discussion and trying out the different options, I now decided to include `args` as a submodule, similar to `spdlog` and `catch2`. I think this is unproblematic, since it's a very small repo, comes with its own CMake file, and is header-only. I disabled the options which build its tests and example when it is included in the Celerity context.

More importantly, the tasks microbenchmark now does slightly more interesting stuff (build a tree and an upside-down tree), but I get some strange behaviour in terms of generated dependencies which I still need to look into.

PeterTh:
Investigate re-using graph layout generators between isolated unit test microbenchmarks and this PR.

fknorr:
In this distributed setting, would be interesting to benchmark common communication patterns (gather / scatter / alltoall) as well as scalar reductions from host buffers against the optimized MPI implementations.

PeterTh:
We do have some microbenchmarks by now; will revisit this in the future.

PR #63, opened 2021-12-07
Warning when not enough cores assigned

facuMH:


include/affinity.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <assert.h>
+
+typedef uint64_t uint64;
```
PeterTh, line 5: This is a bit silly, I'd say just use uint64_t.

include/affinity.h
```diff
@@ -0,0 +1,32 @@
...
+typedef uint32_t uint32;
+
+#define MAX_CORES ((uint64)2048)
+#define AFFINITY_MASK_BITS_PER_QUAD ((uint64)64)
+#define AFFINTY_MASK_NUM_QUADS (MAX_CORES / AFFINITY_MASK_BITS_PER_QUAD)
+
+struct _affinity_mask {
```
PeterTh, line 12: This (and most of the contents of this header) are really more of an implementation detail and should probably be hidden.

include/affinity.h
```diff
@@ -0,0 +1,32 @@
...
+#else
+
+#include <unistd.h>
+typedef cpu_set_t native_cpu_set;
+
+#endif
+
+static native_cpu_set g_affinity_base_mask;
+
+void affinity_int_base_mask();
```
PeterTh, line 31: This also doesn't need to be part of the interface as long as we don't use it.

include/affinity.imp.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#ifdef _WIN32
+
+#include "affinity.win.h"
```
PeterTh, line 5: Do we want to follow this approach in terms of multiplatform support?

I'd personally at least not want this to be directly in `include/`

include/affinity.unix.h
```diff
@@ -0,0 +1,24 @@
...
+void affinity_int_base_mask() {
+	static bool initialized = false;
+	if(initialized) return;
+
+	assert(pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &g_affinity_base_mask) == 0 && "Error retrieving base affinity mask.");
+	initialized = true;
+}
+
+uint32 affinity_cores_available() {
+	affinity_int_base_mask();
```
PeterTh, line 18: This is a very C-style implementation -- in C++ we have better methods for ensuring that the base affinity mask is initialized by the time it is accessed. (i.e. static initialization)

include/runtime.h
```diff
@@ -91,6 +91,9 @@ namespace detail {
 
 		uint64_t sync_id = 0;
 
+		// a priori wee need 3 threads plus 1 for workers. This depends on the application invoking celerity.
+		const uint64_t min_cores_needed = 4;
```
PeterTh, line 95: `constexpr`?

src/runtime.cc
```diff
@@ -96,6 +97,12 @@ namespace detail {
 		cfg = std::make_unique<config>(argc, argv, *default_logger);
 		graph_logger->set_level(cfg->get_log_level());
 
+		uint32 cores;
+		if((cores = affinity_cores_available()) < min_cores_needed) {
+			std::string err_msg = "NOT ENOUGH CORES. AVAILABLE:" + std::to_string(cores);
```
PeterTh, line 102: Does this need to scream? ;)
(also there's no need to build the string in a cumbersome way here, you can use the fmt support of the logger)

include/affinity.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <assert.h>
```
psalz, line 3: Use C++ wrapper headers (`<cassert>`), here and below.

include/affinity.h
```diff
@@ -0,0 +1,32 @@
...
+
+typedef uint64_t uint64;
+typedef uint32_t uint32;
+
+#define MAX_CORES ((uint64)2048)
+#define AFFINITY_MASK_BITS_PER_QUAD ((uint64)64)
+#define AFFINTY_MASK_NUM_QUADS (MAX_CORES / AFFINITY_MASK_BITS_PER_QUAD)
```
psalz, line 10: These could be made `constexpr`.

include/affinity.imp.h
psalz: Agreed, especially since we might be getting more of these soon (cc @BlackMark29A).

include/affinity.win.h
```diff
@@ -0,0 +1,27 @@
...
+	affinity_int_base_mask();
+
+	int bit_mask_length = sizeof(native_cpu_set) * 8;
+	native_cpu_set power = 1;
+	uint32 count = 0;
+	for(uint32 i = 0; i < bit_mask_length) {
+		if((power & g_affinity_base_mask) != 0) { ++count; }
+		power = power << 1;
+	}
+}
```
psalz, line 27: This entire Windows implementation is clearly broken (function doesn't return anything, loop doesn't terminate, where is `sys_affinity_mask` coming from..?); I'd feel a lot more comfortable merging this having Windows CI in place...

src/runtime.cc
```diff
@@ -96,6 +97,12 @@ namespace detail {
 		cfg = std::make_unique<config>(argc, argv, *default_logger);
 		graph_logger->set_level(cfg->get_log_level());
 
+		uint32 cores;
```
psalz, line 100: We have C++17, no need to leak this into the outer scope!

psalz (changes requested):
Thanks! I've added a few notes. I agree with Peter that the implementation could use a modernization pass, also somebody has to at least *once* compile this on Windows first.

include/affinity.imp.h
BlackMark29A: How about this folder structure:
`include/affinity.h`
`include/platform_specific/affinity_unix.h`
`include/platform_specific/affinity_windows.h`

Or do we want to have the platform specific code somewhere else entirely, not even inside `include`?

include/affinity.h
BlackMark29A, line 10: Typo in `AFFINTY_MASK_NUM_QUADS`, should be `AFFINITY_MASK_NUM_QUADS`

Also why are these needed? They are currently unused, since `_affinity_mask` is unused. But even if they were used, is it really necessary to hardcode the maximum core count? Shouldn't these be provided by the platform specific implementation used?

BlackMark29A (changes requested):

include/affinity.win.h
facuMH: so should I remove this implementation until we have a win CI working ?

src/runtime.cc
```diff
@@ -12,6 +12,7 @@
 
 #include <mpi.h>
 
+#include "affinity.imp.h"
```
BlackMark29A, line 15: The `imp` implies that this is an implementation detail, which means it shouldn't be part of the public interface and therefore this should not be how the affinity header is included. This should just be `#include "affinity.h"` and the implementation details should be handled inside this header.

I would get rid of the `affinity.imp.h` entirely, and just directly include the platform specific headers inside `affinity.h`.

include/platform_specific/affinity.win.h
```diff
@@ -0,0 +1,29 @@
...
+	initialized = true;
+}
+
+uint32_t affinity_cores_available() {
+	affinity_int_base_mask();
+
+	int bit_mask_length = sizeof(native_cpu_set) * 8;
+	native_cpu_set power = 1;
+	uint32_t count = 0;
+	for(uint32_t i = 0; i < bit_mask_length; i++) {
```
BlackMark29A, line 24: Since this really is just `popcount` it should be a utility function, and once the code base switches to C++20 can be replaced with: https://en.cppreference.com/w/cpp/numeric/popcount

BlackMark29A (changes requested):

src/platform_specific/affinity.unix.cc
```diff
@@ -11,9 +12,12 @@ uint32_t affinity_cores_available() {
 	native_cpu_set affinity_base_mask;
 	assert(pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &affinity_base_mask) == 0 && "Error retrieving base affinity mask.");
 
-	uint32_t count = 0;
+	static uint32_t count = 0;
```
BlackMark29A, line 15: That's a bug. This will increase the `count` each time this function is called. To make use of proper static initialization you need to wrap this in a helper function, or maybe a lambda.

include/affinity.h
```diff
@@ -1,14 +1,6 @@
...
-#ifdef _WIN32
-
-#include "platform_specific/affinity.win.h"
-
-#else
-
-#include "platform_specific/affinity.unix.h"
-
-#endif
+uint32_t affinity_cores_available();
```
BlackMark29A, line 3: This should be in the `celerity` namespace, shouldn't it?

include/utils.h
```diff
@@ -1,14 +1,19 @@
...
 namespace celerity {
 namespace detail {
 
-	void available_cores_check(std::shared_ptr<logger> log) {
-		uint32_t cores;
-		if((cores = affinity_cores_available()) < min_cores_needed) {
-			log->warn(fmt::format("Too few cores available. Recomended: {}; available: {}", min_cores_needed, cores));
+	template <typename BITMASK>
+	constexpr inline uint32_t popcount(const BITMASK bit_mask) noexcept {
+		static_assert(std::is_integral_v<BITMASK> && std::is_unsigned_v<BITMASK>);
```
BlackMark29A, line 10: You should probably provide an error message for this `static_assert`.

src/platform_specific/affinity.win.cc
```diff
@@ -1,22 +1,22 @@
...
 	assert(GetProcessAffinityMask(GetCurrentProcess(), &affinity_base_mask, &sys_affinity_mask) == 0 && "Error retrieving base affinity mask.");
 
-	int bit_mask_length = sizeof(native_cpu_set) * 8;
-	native_cpu_set power = 1;
-	uint32_t count = 0;
-	for(uint32_t i = 0; i < bit_mask_length; i++) {
-		if((power & affinity_base_mask) != 0) { ++count; }
-		power = power << 1;
-	}
+	static uint32_t count = celerity::details::popcount(affinity_base_mask);
```
BlackMark29A, line 17: Not really a bug here, but the `affinity_base_mask` will still be computed each time the function is called, even if the `count` that's returned is cached. So this should probably also be wrapped in a helper function/lambda.

BlackMark29A (changes requested):

include/utils.h
```diff
@@ -1,14 +1,19 @@
-#include "affinity.h"
-#include "logger.h"
+#pragma once
+
+#include <type_traits>
 
 namespace celerity {
 namespace detail {
```
BlackMark29A, line 6: I don't think `popcount` should be in `detail`, it's not an implementation detail. Maybe it should be in a `utils` namespace? Or just in the `celerity` namespace.

@psalz Does celerity not have a place for utility functions, is this really the first utility function for celerity?

src/platform_specific/affinity.win.cc
psalz: I argued (offline) in favor of not doing any caching at all, since we call this function exactly *once*. Should it ever end up on some hot path (which I seriously doubt), we can optimize it then. Either way, `count` shouldn't be `static`.

include/utils.h
psalz: > I don't think popcount should be in detail, it's not an implementation detail. Maybe it should be in a utils namespace? Or just in the celerity namespace.

That's actually fine, imo. Everything in the `celerity` namespace is part of the public (user-facing) API, everything else lives inside `detail`. We could have a `detail::utils` namespace though - no strong opinion there.

> @psalz Does celerity not have a place for utility functions, is this really the first utility function for celerity?

Nope, [this](https://github.com/celerity/celerity-runtime/blob/da1c240e0146b32f1fa095346f5bd99a18ca3bf2/include/graph_generator.h#L24) [was](https://github.com/celerity/celerity-runtime/blob/da1c240e0146b32f1fa095346f5bd99a18ca3bf2/include/intrusive_graph.h#L18) the approach taken so far :P.

BlackMark29A: Right, `popcount` is not something that celerity provides. In which case I would be in favor of putting it inside `detail::utils` and having this header be the `utils` header where `TODO move to utils` code can be refactored to in the future.

src/platform_specific/affinity.unix.cc
```diff
@@ -0,0 +1,23 @@
+
+#ifdef __linux__
+
+#include <cassert>
+#include <pthread.h>
+#include <sched.h>
+#include <unistd.h>
+
+typedef cpu_set_t native_cpu_set;
```
BlackMark29A, line 9: 1) This alias seems rather pointless, since `cpu_set_t` is just as good a name as `native_cpu_set` and this type is only needed in the function below.
2) I don't think `typedef` should be used at all in modern C++ since we have `using` now.

src/platform_specific/affinity.win.cc
```diff
@@ -0,0 +1,22 @@
+#ifdef _WIN32
+
+#include <Windows.h>
+#include <cassert.h>
+#include <io.h>
+
+#include "affinity.h"
+#include "utils.h"
+
+typedef DWORD_PTR native_cpu_set;
```
BlackMark29A, line 10: Maybe move this inside the `affinity_cores_available` function as `using native_cpu_set = DWORD_PTR;`

src/platform_specific/affinity.win.cc
```diff
@@ -0,0 +1,22 @@
...
+namespace detail {
+
+	uint32_t affinity_cores_available() {
+		using native_cpu_set = DWORD_PTR;
+
+		native_cpu_set affinity_base_mask;
+		[[maybe_unused]] native_cpu_set sys_affinity_mask;
+		auto base_mask_error = GetProcessAffinityMask(GetCurrentProcess(), &affinity_base_mask, &sys_affinity_mask);
+		assert(base_mask_error == 0 && "Error retrieving base affinity mask.");
+		return celerity::detail::util::popcount(affinity_base_mask);
```
BlackMark29A, line 18: Since we're already in namespace `celerity::detail` here, this can be simplified to `util::popcount(affinity_base_mask);`

src/runtime.cc
```diff
@@ -96,6 +98,10 @@ namespace detail {
 		cfg = std::make_unique<config>(argc, argv, *default_logger);
 		graph_logger->set_level(cfg->get_log_level());
 
+		if(uint32_t cores = affinity_cores_available(); cores < min_cores_needed) {
```
BlackMark29A, line 101: `cores` should also be `const`.

test/utils_tests.cc
```diff
@@ -0,0 +1,36 @@
...
+#else
+#include <sched.h>
+			SECTION("in Posix") {
+				cpu_set_t cpu_mask;
+				CPU_ZERO(&cpu_mask);
+				CPU_SET(0, &cpu_mask);
+				pthread_setaffinity_np(pthread_self(), sizeof(cpu_mask), &cpu_mask);
+			}
+#endif
+			auto cores = affinity_cores_available();
```
BlackMark29A, line 30: This should also be `const`.

BlackMark29A (approved):
Apart from some minor pedantic things, this is looking good to me now.

test/utils_tests.cc
```diff
@@ -0,0 +1,36 @@
...
+		TEST_CASE("affinity check", "[utils]") {
+			// REQUIRE_FALSE(runtime::is_initialized());
+
+#ifdef _WIN32
+			SECTION("in Windows") {
+				DWORD_PTR cpu_mask = 1;
+				SetProcessAffinityMask(GetCurrentProcess(), cpu_mask);
+			}
+#else
+#include <sched.h>
```
BlackMark29A, line 22: Never ever include headers inside a namespace, unless you are absolutely sure that is what you want. This pollutes the `celerity::detail::utils` namespace with whatever is inside `sched.h`, which is very likely not what you want.

include/affinity.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <cstdint>
+
+namespace celerity {
+namespace detail {
+
+	uint32_t affinity_cores_available();
+
+	// a priori we need 3 threads plus 1 for workers. This depends on the application invoking celerity.
```
psalz, line 10: Could we rephrase this to "plus at least 1 for host-task workers`? Because that's why it's application dependent. Also this does not yet factor in CUDA/MPI/whatnot. (Maybe we should bump this to at least 5?).

include/utils.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <cstdint>
+#include <type_traits>
+
+namespace celerity {
+namespace detail {
+	namespace util {
```
psalz, line 8: Since we already have a `test_utils` (plural) namespace I'd prefer naming this `utils` for consistency.

src/platform_specific/affinity.unix.cc
```diff
@@ -0,0 +1,29 @@
...
+
+#include <pthread.h>
+#include <sched.h>
+
+#include "affinity.h"
+
+namespace celerity {
+namespace detail {
+
+	uint32_t affinity_counter(const cpu_set_t& base_mask) {
```
psalz, line 13: I don't think this function is needed, there is a `CPU_COUNT` macro that does exactly this!

src/platform_specific/affinity.unix.cc
```diff
@@ -0,0 +1,29 @@
...
+		auto count = 0;
+		for(auto i = 0; i < CPU_SETSIZE; i++) {
+			if(CPU_ISSET(i, &base_mask)) { ++count; }
+		}
+		return count;
+	}
+
+	uint32_t affinity_cores_available() {
+		cpu_set_t affinity_base_mask;
```
psalz, line 22: Why is it called "base mask"? As far as I can tell this term was introduced as part of the original Insieme implementation and has no real meaning in this context.

src/runtime.cc
```diff
@@ -96,6 +98,10 @@ namespace detail {
 		cfg = std::make_unique<config>(argc, argv, *default_logger);
 		graph_logger->set_level(cfg->get_log_level());
 
+		if(const uint32_t cores = affinity_cores_available(); cores < min_cores_needed) {
+			default_logger->warn("Too few cores available. Recomended at least: {}; available: {}", min_cores_needed, cores);
```
psalz, line 102: Typo "Recom**m**ended". In general, can we rephrase to be a bit more descriptive? Something like
> Celerity has detected that only X logical cores are available to this process. It is recommended to assign at least 4 logical cores. Performance may be negatively impacted.

psalz (changes requested):
I've added a few more comments. In general I see these two things still outstanding:
- Do we care about physical or logical cores? Currently we count logical cores.
- Still need to test this on Windows before we can merge. Do we maybe even need to link to some Windows runtime library?

src/platform_specific/affinity.win.cc
```diff
@@ -11,11 +11,11 @@ namespace detail {
...
 		using native_cpu_set = DWORD_PTR;
 
-		native_cpu_set affinity_base_mask;
+		native_cpu_set available_cores;
 		[[maybe_unused]] native_cpu_set sys_affinity_mask;
-		const auto base_mask_error = GetProcessAffinityMask(GetCurrentProcess(), &affinity_base_mask, &sys_affinity_mask);
-		assert(base_mask_error == 0 && "Error retrieving base affinity mask.");
-		return util::popcount(affinity_base_mask);
+		const auto affinity_error = GetProcessAffinityMask(GetCurrentProcess(), &available_cores, &sys_affinity_mask);
+		assert(affinity_error == 0 && "Error retrieving base affinity mask.");
```
BlackMark29A, line 17: This is a bug. `GetProcessAffinityMask` returns `TRUE` (yes, all caps, because it's of type `BOOL`) on success, so this `assert` will fail if the function call was successful.

BlackMark29A (changes requested):

src/platform_specific/affinity.win.cc
psalz: (Also the message is still referring to a "base" affinity mask)

PR #64, opened 2021-12-16
Standalone example projects

fknorr:
With 0.3.0, examples are built from a single CMakeLists file. While this is great in terms of CMake code deduplication, it hinders the most important user-facing feature of examples: Providing a "copy-paste programming" starting point to newcomers.

This PR adds minimalist CMakeLists files to each example folder, which the updated CI will now also use to build the examples as integration tests. Third-party dependencies like `stb` are symlinked into the respective example folders to allow users to start off with a `cp -rL` copy. The top-level examples folder now includes a basic HowTo as well.

The downside to this is slight duplication between `examples/CMakeLists.txt` and `examples/*/CMakeLists.txt`. Special dependencies, e.g. to HDF5, have to be listed in two places, and version numbers (`cmake_minimum_required` and the Celerity version) have to be updated in every example project. I argue though that this is necessary for the new _standalone experience_.
 
There is one new hacky interaction between the build system and the CI: Some examples might need to be skipped because they are not compatible with the Celerity feature set based on the chosen SYCL implementation (e.g. reduction support). The CI does not and should not know about example binary names, so we now try building all subdirectories of the `examples` folder; if any CMake invocation there raises an error containing the substring `Skip this example`, we do not report the build as failed.


examples/README.md
```diff
@@ -0,0 +1,47 @@
...
+[`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html)
+to include the Celerity installation directory. Depending on your SYCL implementation,
+you may also have to specify its library installation directory and target options.
+
+### With hipSYCL (exemplary)
+
+```shell
+cd my-project
+cmake -B build \
+    -DCMAKE_PREFIX_PATH="../celerity-runtime-install/lib/cmake;../hipSYCL-install/lib/cmake" \
```
psalz, line 27: Here and below: `-DCMAKE_PREFIX_PATH="../celerity-runtime-install;../hipSYCL-install"` should suffice!

examples/convolution/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.13)
+project(convolution LANGUAGES CXX)
+
+find_package(Celerity 0.3.0 REQUIRED)
+
+add_executable(convolution convolution.cc)
+set_property(TARGET convolution PROPERTY CXX_STANDARD 17)
```
psalz, line 7: General note: We should probably do this automatically in `add_celerity_to_target`. I've added it to *the list*.

examples/convolution/stb
```diff
@@ -0,0 +1 @@
+../../vendor/stb
```
psalz, line 1: I don't think this works on Windows, but we can worry about that later.

psalz (changes requested):
This is awesome! Just a few very minor remarks.

It would be cool to also have the new *standalone experience* documented on the website, but we can do that in a separate change!

examples/convolution/stb
fknorr: There's symlink support for windows-git now, but I think it requires special configuration. We should probably just place a copy of `stb` in both example folders, git will de-duplicate them anyway.

examples/convolution/CMakeLists.txt
fknorr: ... but we need to make sure not to down-grade from C++20 if we do this.

examples/README.md
fknorr: I'll get rid of the hipSYCL line after #65 .

examples/README.md
```diff
@@ -0,0 +1,47 @@
+# Celerity Example Projects
+
+These example projects are a useful starting point for developing your own Celerity
+application. All examples will also be built automatically in-tree with the runtime
+when the `CELERITY_BUILD_EXAMPLES` CMake option is set (true by default).
+
+## Setup
+
+Begin by copying an example project folder that best fits your use case,
+while resolving symlinks:
```
PeterTh, line 10: I'd be a bit worried about the symlink part on some target platforms, but it's fine for now.

fknorr:
I've have now copied stb instead of symlinking it (git will dedup this anyway); and added a paragraph to the website.

psalz (approved):

PR #66, opened 2021-12-17
Work around potential race condition in wave_sim example

psalz:
For lower `--sample-rate`s (1 or 2), subsequent `store` host tasks don't have a clear ordering in the task graph. This means they could be executed in any order (leading to wrong results) or even concurrently (potentially causing a crash). Since we don't have a proper way of expressing order-dependencies between host tasks with side effects yet, this introduces synchronization after each store task as a workaround.

Note that this issue has been known for a while actually, and in CI we worked around it by setting a `--sample-rate` of `2`, which should prevent concurrent execution of host tasks during the main loop (since they are spaced far enough to create a transitive dependency chain). What tripped us up however was the initial store before the main loop:

image

Here tasks 4 and 7 could be executed concurrently. What I don't understand yet however is why we saw a sudden increase in crashes of `wave_sim` ever since #60 was merged...

### DIFF
--- a/examples/wave_sim/wave_sim.cc
+++ b/examples/wave_sim/wave_sim.cc
@@ -64,28 +64,22 @@ void update(celerity::distr_queue& queue, celerity::buffer<float, 2> up, celerit
 }
 
 template <typename T>
-void store(celerity::distr_queue& queue, celerity::buffer<T, 2> up, std::vector<std::vector<float>>& result_frames) {
+void store(celerity::distr_queue& queue, celerity::buffer<T, 2> up, celerity::buffer<T, 1> sampled_frames, size_t frame_idx) {
 	const auto range = up.get_range();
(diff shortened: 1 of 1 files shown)


fknorr (commented):
Sure, this should solve the problem, and I agree that at some point we want to have a way of expressing side-effects to create proper order-dependencies in the TDAG. On the other hand `slow_full_sync` in an example is pretty ugly.

Why don't we write the frames to an intermediate Celerity buffer and serialize that in a final host task at the end of the program? Sure, this will consume more memory, but with the default settings and a sample rate of 1, we would end up with just above 400MB.

psalz:
You're right, that'd be a cleaner way of solving this (and I agree that the sync is super ugly). Will do.

psalz:
Updated to use a `celerity::buffer` instead of `std::vector` for storing sampled frames, as suggested.

fknorr (approved):
LGTM!

PR #67, opened 2021-12-21
Buffer Unregister

facuMH:
Add a test for a desired situation where unregistering a buffer is accepted and delete error log for when a buffer's reference count is 0.

Since we want to get rid of live-pass evaluation of outer CGF in the future, there is probably no point spending effort on keeping the warning mechanism for detecting reference-captured buffers.

### DIFF
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -578,6 +578,41 @@ namespace detail {
 		REQUIRE_FALSE(bm.has_active_buffers());
 	}
 
+	TEST_CASE("buffer_manager allows buffer deallocation", "[buffer_manager][dealloc]") {
+		celerity::distr_queue q;
+		buffer_id b_id;
+		auto& bm = runtime::get_instance().get_buffer_manager();
+		auto& tm = runtime::get_instance().get_task_manager();
+		constexpr int new_horizon_step = 2;
+		tm.set_horizon_step(new_horizon_step);
+		{
+			celerity::buffer<int, 1> b(celerity::range<1>(128));
+			b_id = celerity::detail::get_buffer_id(b);
+			q.submit([=](celerity::handler& cgh) {
+				celerity::accessor a{b, cgh, celerity::access::all(), celerity::write_only};
+				cgh.parallel_for<class UKN(i)>(b.get_range(), [=](celerity::item<1> it) {});
+			});
+			REQUIRE(bm.has_buffer(b_id));
+		}
+		celerity::buffer<int, 1> c(celerity::range<1>(128));
+		// we need horizon_step_size * 3 + 1 tasks to generate the third horizon,
+		// and one extra task to trigger the clean_up process
+		for(int i = 0; i < (new_horizon_step * 3 + 2); i++) {
+			q.submit([=](celerity::handler& cgh) {
+				celerity::accessor a{c, cgh, celerity::access::all(), celerity::write_only};
+				cgh.parallel_for<class UKN(i)>(c.get_range(), [=](celerity::item<1>) {});
+			});
+			// this sync is inside the loop because otherwise there is a race between the prepass and the executor informing the TDAG
+			// of the executed horizons, meaning that task deletion is not guaranteed.
+			q.slow_full_sync();
+		}
+		// require buffer b was indeed unregistered.
+		REQUIRE_FALSE(bm.has_buffer(b_id));
+
+		// TODO: check whether error was printed or not
+		maybe_print_graph(celerity::detail::runtime::get_instance().get_task_manager());
+	}
+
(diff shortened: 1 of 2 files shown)


test/runtime_tests.cc
```diff
@@ -763,6 +763,35 @@ namespace detail {
...
+			celerity::buffer<int, 1> b(celerity::range<1>(128));
+			b_id = celerity::detail::get_buffer_id(b);
+			q.submit([=](celerity::handler& cgh) {
+				celerity::accessor a{b, cgh, celerity::access::all(), celerity::write_only};
+				cgh.parallel_for<class UKN(i)>(b.get_range(), [=](celerity::item<1> it) {});
+			});
+			REQUIRE(bm.has_buffer(b_id));
+		}
+		celerity::buffer<int, 1> c(celerity::range<1>(128));
+		// 10 tasks so that 1st task horizon is arleady applied considering task_horizon_step_size=4
```
PeterTh, line 780: Typo.

Also, should we set the task horizon step size in tests which depend on a specific value?

test/runtime_tests.cc
```diff
@@ -763,6 +763,35 @@ namespace detail {
...
+		for(int i = 0; i < 14; i++) {
+			q.submit([=](celerity::handler& cgh) {
+				celerity::accessor a{c, cgh, celerity::access::all(), celerity::write_only};
+				cgh.parallel_for<class UKN(i)>(c.get_range(), [=](celerity::item<1>) {});
+			});
+			q.slow_full_sync();
+		}
+		// require bufffer b was indeed unregistered.
```
PeterTh, line 788: Another typo.

PeterTh (changes requested):

test/runtime_tests.cc
```diff
@@ -763,6 +763,39 @@ namespace detail {
 		REQUIRE_FALSE(bm.has_active_buffers());
 	}
 
+	TEST_CASE("buffer_manager allows buffers deallocation", "[buffer_manager][dealloc]") {
```
fknorr, line 766: Typo, should be "buffer deallocation" (singular)

test/runtime_tests.cc
```diff
@@ -763,6 +763,39 @@ namespace detail {
...
+		}
+		celerity::buffer<int, 1> c(celerity::range<1>(128));
+		// we need horizon_step_size * 3 + 1 tasks to generate the third horizon,
+		// and one extra task to trigger the clean_up process
+		for(int i = 0; i < (new_horizon_step * 3 + 2); i++) {
+			q.submit([=](celerity::handler& cgh) {
+				celerity::accessor a{c, cgh, celerity::access::all(), celerity::write_only};
+				cgh.parallel_for<class UKN(i)>(c.get_range(), [=](celerity::item<1>) {});
+			});
+			q.slow_full_sync();
```
fknorr, line 790: It should be enough for this to happen once after the loop.

fknorr (approved):
Thanks! Aside from my nitpicks, LGTM.

PeterTh (approved):
Approved, but please do the changes @fknorr requested before merging.

test/runtime_tests.cc
```diff
@@ -763,6 +763,39 @@ namespace detail {
 		REQUIRE_FALSE(bm.has_active_buffers());
 	}
 
+	TEST_CASE("buffer_manager allows buffers deallocation", "[buffer_manager][dealloc]") {
+		celerity::distr_queue q;
+		buffer_id b_id;
+		auto& bm = runtime::get_instance().get_buffer_manager();
+		auto& tm = runtime::get_instance().get_task_manager();
+		int new_horizon_step = 2;
```
BlackMark29A, line 771: This is never modified and should probably be `const`, or even `constexpr`.

BlackMark29A (approved):
Looks good to me.

test/runtime_tests.cc
facuMH: the `slow_full_sync()` needs to be inside the `for` otherwise there is a race between the prepass and the executor informing the tdag of executed horizons. for more info pls reach @psalz.
I will add a comment on the test explaining this.

fknorr: You're right! The explanation is the same as in the `tasks behind the deletion horizon are deleted` runtime test: At least the second-to-last task must be awaited explicitly so that the last horizon is executed, otherwise creation of the final task does not trigger a cleanup.

test/runtime_tests.cc
psalz: Or it could just be a literal that is passed into `set_horizon_step` directly...

psalz:
Looks good! Please squash & rebase to resolve conflicts, then we can merge!

test/runtime_tests.cc
fknorr: @psalz it's used for computing the number of tasks to submit in the loop below - which I would like to keep.

psalz: You're right, sorry!

PR #69, opened 2021-12-23
Correctly apply horizons to collective-group dependencies

fknorr:
Tasks and commands in the same collective group are serialized using `ORDER_DEP` dependencies to guarantee a single total order of execution across all workers. To generate these dependencies, the last task / command in each group is tracked similar to the last writers of buffers.

When a new horizon is applied, the tracked last collective group task / command ids must be replaced by the previous horizon. This follows the same rule as last buffer writers. This was not done previously, which will cause access to dangling task and command ids once the last collective operation in a group is deleted from the graph.

This PR correctly replaces the tracked task / command ids when they are subsumed by a horizon and adds regression tests. Also, the `last_collective_commands` map in `graph_generator` is moved for consistency with last-buffer-writer tracking.

BugGraph

### DIFF
--- a/test/task_graph_tests.cc
+++ b/test/task_graph_tests.cc
@@ -505,5 +505,33 @@ namespace detail {
 		maybe_print_graph(tm);
 	}
 
+	TEST_CASE("collective host tasks do not order-depend on their predecessor if it is shadowed by a horizon", "[task_manager][task-graph][task-horizon]") {
+		// Regression test: the order-dependencies between host tasks in the same collective group are built by tracking the last task in each collective group.
+		// Once a horizon is inserted, new collective host tasks must order-depend on that horizon instead.
+
+		task_manager tm{1, nullptr, nullptr};
+		tm.set_horizon_step(2);
+
+		const auto first_collective = test_utils::add_host_task(tm, experimental::collective, [&](handler& cgh) {});
+
+		// generate exactly two horizons
+		test_utils::mock_buffer_factory mbf(&tm);
+		auto buf = mbf.create_buffer(range<1>(1));
+		for(int i = 0; i < 5; ++i) {
+			test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf.get_access<access_mode::discard_write>(cgh, all{}); });
+		}
+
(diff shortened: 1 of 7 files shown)


test/task_graph_tests.cc
```diff
@@ -505,5 +505,33 @@ namespace detail {
...
+			test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf.get_access<access_mode::discard_write>(cgh, all{}); });
+		}
+
+		// This must depend on the first horizon, not first_collective
+		const auto second_collective = test_utils::add_host_task(tm, experimental::collective, [&](handler& cgh) {});
+
+		for(const auto& dep : tm.get_task(second_collective)->get_dependencies()) {
+			const auto type = dep.node->get_type();
+			CHECK(type == task_type::HORIZON);
+			CHECK(dep.kind >= dependency_kind::ORDER_DEP);
```
psalz, line 530: `>=` ?

psalz (approved):
LGTM!

test/task_graph_tests.cc
fknorr: Right, `==` is enough here. Upgrading dependencies is not incorrect in the general case, but should never be needed here.

PeterTh (approved):

BlackMark29A (approved):
Since there is a requirement for `dependency_kind` to be ordered, and this assumption must hold for the `std::max(...)` part of the code to work as intended, this assumption should probably be `static_assert`ed to document this requirement, and make it less brittle:

```cpp
static_assert(dependency_kind::ANTI_DEP < dependency_kind::ORDER_DEP && dependency_kind::ORDER_DEP < dependency_kind::TRUE_DEP);
```

Edit:
Actually, never mind, the `std::max` part doesn't work on `dependency_kind`, and this assumption was only used by `dep.kind >= dependency_kind::ORDER_DEP` which has since been changed.

Otherwise this looks good to me.

fknorr:
No, you're actually right @BlackMark29A . There is not a literal `std::max()`, but `intrusive_graph_node::add_dependency()` makes exactly that assumption. I've added the static assert you proposed there.

PR #70, opened 2021-12-24
Fix race conditions between task and command horizons

fknorr:
This PR addresses two race conditions:

- A data race in an assertion within `task_manager::notify_horizon_executed`: `task_map` is accessed without a lock, but the function is called from the executor thread while the map is updated from the main thread.
- A race between horizon task creation in the main thread and horizon job execution in the executor thread: workers can receive horizon commands before reaching the corresponding horizon task, so the executor must delay notifying the task manager until the horizon task is available. A similar mechanism exists for (compute) task commands.

I could not come up with a good way to write a regression test (`distr_tests`) for this. Artificially delaying compute- or horizon task creation in workers does not seem to trigger the error. It does however occur somewhat reliably with `wave_sim` on 4 nodes and the Intel OpenCL platform. This explains some of the `wave_sim` failures after 0.3.0.

### DIFF
--- a/src/task_manager.cc
+++ b/src/task_manager.cc
@@ -43,13 +43,19 @@ namespace detail {
 	}
 
 	void task_manager::notify_horizon_executed(task_id tid) {
-		assert(task_map.at(tid)->get_type() == task_type::HORIZON);
+#ifndef NDEBUG
+		{
+			std::lock_guard lock{task_mutex};
+			assert(task_map.count(tid) != 0);
+			assert(task_map.at(tid)->get_type() == task_type::HORIZON);
+		}
 		assert(executed_horizons.empty() || executed_horizons.back() != tid);
-		executed_horizons.push(tid);
+#endif
 
+		executed_horizons.push(tid); // no locking needed - see definition
 		if(executed_horizons.size() >= horizon_deletion_lag) {
 			// actual cleanup happens on new task creation
-			horizon_task_id_for_deletion = executed_horizons.front();
+			horizon_task_id_for_deletion.store(executed_horizons.front()); // atomic
 			executed_horizons.pop();
 		}
 	}
--- a/include/task_manager.h
+++ b/include/task_manager.h
@@ -134,7 +134,8 @@ namespace detail {
 		// The latest horizon task created. Will be applied as last writer once the next horizon is created.
 		task* current_horizon_task = nullptr;
 
-		// Queue of horizon tasks for which the associated commands were executed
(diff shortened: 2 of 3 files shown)


src/task_manager.cc
```diff
@@ -43,13 +43,17 @@ namespace detail {
...
 		assert(task_map.at(tid)->get_type() == task_type::HORIZON);
 		assert(executed_horizons.empty() || executed_horizons.back() != tid);
-		executed_horizons.push(tid);
+#endif
 
+		executed_horizons.push(tid); // no locking needed - see definition
 		if(executed_horizons.size() >= horizon_deletion_lag) {
 			// actual cleanup happens on new task creation
-			horizon_task_id_for_deletion = executed_horizons.front();
+			horizon_task_id_for_deletion.store(executed_horizons.front()); // atomic
```
PeterTh, line 56: I don't understand this change -- the assignment operator on `std::atomic`s is atomic.

fknorr: That's clarification only. `notify_horizon_executed` doesn't take a lock (in release builds) but still accesses data members, so I figure it's good to know why that is safe.

psalz (approved):
LGTM

psalz:
Since you asked, I was able to somewhat consistently reproduce the bug in a test with something like this:
```c++
runtime::get_instance().get_task_manager().register_task_callback([rank](auto, auto) {
	if(rank == 1) { std::this_thread::sleep_for(std::chrono::milliseconds(100)); }
});
```
...which is obviously horrible and brittle (relying on the implementation detail that we invoke the callback before generating a horizon). So yeah, I don't think we can reliably test this unfortunately.

facuMH (approved):
looks ok.

src/task_manager.cc
```diff
@@ -43,13 +43,17 @@ namespace detail {
 	}
 
 	void task_manager::notify_horizon_executed(task_id tid) {
+#ifndef NDEBUG
+		std::lock_guard lock{task_mutex};
```
BlackMark29A, line 47: What is the intended scope of this `lock_guard`? The way it is now it will remain locked for the entire function. I assume that is not intended, and it should only be locked within the `#ifndef` block, which would require an additional block scope.

BlackMark29A (approved):
Looks good to me now.

src/task_manager.cc
PeterTh: I feel like this is a bit over-clarified with using `.store` and `// atomic` if the type is already `atomic`, but it's not big deal.

PR #71, opened 2021-12-27
Avoid UB from static destruction order

fknorr:
So far, `runtime` and `user_benchmarker` are both singletons. On shutdown, the runtime explicitly `reset()`s the global `user_benchmarker`. This causes a double-free if the linker decides to call the destructor of the (global) benchmarker instance before the (global) runtime instance.

This PR assigns the benchmarker to the runtime as a non-static member, avoiding a race on static destruction without changing any externally visible semantics.

### DIFF
--- a/include/runtime.h
+++ b/include/runtime.h
@@ -15,6 +15,11 @@
 #include "types.h"
 
 namespace celerity {
+
+namespace experimental::bench::detail {
+	class user_benchmarker;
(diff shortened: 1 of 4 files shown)


psalz (approved):
Other than formatting, LGTM!

facuMH (approved):
LGTM

PR #72, opened 2021-12-28
Fix deprecation warnings for ComputeCpp & DPC++

fknorr:
ComputeCpp's Clang 8 has trouble handling `#pragma GCC diagnostic` sections whenever that code triggers a deprecation warning on instantiating a template. This is solved by moving all tests of deprecated APIs from `runtime_tests.cc` to a separate test binary `runtime_deprecation_tests.cc`, which centrally disables `-Wdeprecated-declarations`. This also moves `buffer_manager_fixture` from `runtime_tests.cpp` to `test_utils.hh` in order to avoid duplication between the old and new runtime tests `.cc` file.

We were also still using deprecated kernel offsets and enum variants for which DPC++ emits a warning. These were refactored according to SYCL 2020.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -253,7 +253,10 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 
   private:
 #if WORKAROUND_COMPUTECPP || WORKAROUND_DPCPP
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations" // target::gobal_buffer is now target::device, but only for very recent versions of DPC++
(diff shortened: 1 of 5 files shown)


fknorr:
As it turns out, some of the deprecation warnings that evaded `#pragma GCC diagnostic ignored` were the host compiler's fault: Our CI setup defaulted to GCC as the C++ compiler, which appears to have similar bug. I have updated the Docker images to use Clang 10 instead, which now finally gets rid off all spurious warnings.

psalz (approved):
Good change, thanks! Together with your recent workaround for the kernel naming warning, ComputeCpp's compiler output during CI is actually readable now.

facuMH (approved):
LGTM

PR #76, opened 2022-01-05
Fix race around reference-capture in matmul example

fknorr:
PR #60 introduced a data race in the matmul example: `verification_passed` is reference-captured in a host task, but no synchronization takes place before reading it from the main thread. This PR adds explicit synchronization and also moves the verification step to a separate method for legibility.

### DIFF
--- a/examples/matmul/matmul.cc
+++ b/examples/matmul/matmul.cc
@@ -60,31 +60,18 @@ void multiply(celerity::distr_queue queue, celerity::buffer<T, 2> mat_a, celerit
 	});
 }
 
-int main(int argc, char* argv[]) {
-	bool verification_passed = true;
-
-	celerity::distr_queue queue;
-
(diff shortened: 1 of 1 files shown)


psalz (approved):
LGTM!

BlackMark29A (approved):
It took me a minute to figure out that the removal of the block scope introduced the race condition, by removing the implicit synchronization that happens because the queue goes out of scope.  

Making the synchronization explicit, with comment, makes this hidden complexity visible, which makes the code easier to read, and less brittle against further refactoring mistakes. So I think this is much better than just re-introducing the block scope.

fknorr:
The real culprit here is `allow_by_ref`, which makes you pause to think when first writing code that has to synchronize with the main thread, but not in a refactoring like this. Celerity should have a safer, more data-flow-centric replacement for reference captures.

PR #77, opened 2022-01-10
Support ComputeCpp 2.7.0 and 2.8.0

fknorr:
This adds support for ComputeCpp 2.7.0 and 2.8.0, both with the "legacy" Clang-8 based compiler and the experimental Clang-13 based toolchain.

The most significant change required is related to SYCL accessors involving accessor offsets. Their specification changed from SYCL 1.2.1 to 2020, and in addition, their implementation in ComputeCpp 2.7.0 is buggy. As a robust solution to this inconsistency, I've added the internal `ranged_sycl_access` API, which allows using the same set of workarounds in all affected locations (i.e. `celerity::accessor` and `device_buffer_storage`). A new test binary `sycl_tests` is added with explicit unit tests of the new API and with the intent of adding more conformance and workaround tests in the future.

With this PR we now also instruct ComputeCpp to use SYCL language version 2020, which gets rid of a few other, minor workarounds in Celerity.

The experimental compiler supports unnamed kernels in principle, however with Celerity, using them either produces nonsensical compiler errors or crashes at runtime during kernel compilation. Since partial support for unnamed kernels seems to stay with us for some time, I've also added the `CELERITY_FEATURE_UNNAMED_KERNELS` feature detection macro that is categorically disabled for ComputeCpp now.

The CI matrix has become quite large now, testing five ComputeCpp distributions both in debug and release builds. We should probably sacrifice some of these configurations for the sake of finite CI run times.


psalz:
CI took 1 hour and 18 minutes - ouch! I suggest we get rid of 2.6 entirely, and for 2.7 only test the stable compiler (i.e., 2.7, 2.8, 2.8-experimental).

fknorr:
With most release builds removed, the run now took 38 minutes.

CMakeLists.txt
```diff
@@ -219,6 +222,8 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
   target_compile_options(celerity_runtime PRIVATE -Wall -Wextra -Wno-unused-parameter -Werror=return-type -Werror=init-self)
 endif()
 
+target_compile_options(celerity_runtime PRIVATE "${CELERITY_CXX_FLAGS}")
```
psalz, line 256: Instead of adding `${CELERITY_CXX_FLAGS}` to a target manually in `add_celerity_to_target`, couldn't this just be made ` PUBLIC`?

psalz (approved):
LGTM Thanks!

CMakeLists.txt
fknorr: You're right!

PeterTh (approved):

PR #78, opened 2022-01-10
Support local memory in ComputeCpp

fknorr:
In the two-phase execution model, `celerity::local_accessor` requires a placeholder constructor for `sycl::local_accessor` that can be invoked without a `sycl::handler`. hipSYCL supports this natively, and for DPC++ we currently hack around this by passing a dereferenced null pointer as the `sycl::handler &`. We are able to do this somewhat safely since their `sycl::local_accessor` does not actually use the handler argument. For ComputeCpp however this does not hold, so local memory support was disabled for that implementation so far.

This PR introduces new workarounds for DPC++ and ComputeCpp by abusing inheritance and friend-declarations to conjure a temporary `sycl::handler` that can be used to construct the `sycl::local_accessor`. Since none of the implementations keep a reference to the handler around, we are able to provide a true placeholder-accessor factory method without invoking UB.

The `CELERITY_FEATURE_LOCAL_ACCESSOR` feature is now always enabled.


include/accessor.h
```diff
@@ -556,39 +571,26 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
-#else
   private:
 #if WORKAROUND_DPCPP
 	using sycl_accessor = cl::sycl::accessor<DataT, Dims, cl::sycl::access::mode::read_write, cl::sycl::access::target::local>;
 #else
 	using sycl_accessor = cl::sycl::local_accessor<DataT, Dims>;
 #endif
 
+	template <typename Index>
+	using subscript_type = decltype(std::declval<sycl_accessor>()[std::declval<const Index&>()]);
```
PeterTh, line 582: I feel like this is a bit far away from its only use (at line 623 as far as I can see).

PeterTh (commented):
A general question: is it the case that with the new workaround you introduce here we still need separate DPCPP and COMPUTECPP code paths? (i.e. the workaround for neither of them works on the other, and there's no single code path which subsumes both?)

I guess you probably checked that, but it's nice to want things ;)

fknorr:
@PeterTh I was able to partially unify the workaround paths, but there doesn't seem to be a way to use the same exact code for both.

include/accessor.h
```diff
@@ -50,19 +50,38 @@ namespace detail {
...
+// Hack: DPC++ and ComputeCpp do not implement the SYCL 2020 sycl::local_accessor default constructor yet and always require a handler for construction.
+// Since there is no SYCL handler in the pre-pass, we abuse inheritance and friend declarations to conjure a temporary sycl::handler that can be passed to the
+// existing local_accessor constructor. This works because neither accessor implementation keep references to the handler internally or interacts with context
+// surrounding the fake handler instance.
+#if WORKAROUND_DPCPP
+	// The DPC++ handler declares `template<...> friend class accessor<...>`, so we specialize sycl::accessor with a made-up type and have it inherit from
+	// sycl::handler in order to be able to call the private constructor of sycl::handler.
+	struct hack_accessor_specialization_type {};
+	using hack_null_sycl_handler = cl::sycl::accessor<celerity::detail::hack_accessor_specialization_type, 0, cl::sycl::access::mode::read,
+	    cl::sycl::access::target::host_buffer, cl::sycl::access::placeholder::true_t, void>;
```
psalz, line 62: This is insane 😹

psalz (approved):
Crazy hack, LGTM!

PR #80, opened 2022-01-24
Overhaul logging mechanism

psalz:
This changes the logging mechanism used throughout the runtime from having to pass around the `detail::logger` class to simple free-standing macros such as `CELERITY_INFO(...)`. While the new mechanism no longer allows for arbitrarily nested "contexts", a single context can be set through `CELERITY_SET_LOG_CTX` (currently only used in worker jobs).

Full list of changes:
- Get rid of JSON output for trace log level
- Remove logger class, use macros instead
- Only a single level of context supported
- Adjust log level for many messages, distinguish between debug/trace
- Add additional information in various logging messages
- Default log level to info (release builds) and debug (debug builds)


include/command_graph.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <memory>
+#include <ostream>
```
fknorr, line 4: I don't like iostreams as an include dependency. Since this is for a debugging feature, would it be feasible to return `std::string` from the graph-printing facilities instead?

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+#include <spdlog/spdlog.h>
+
+#include <spdlog/fmt/ostr.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+
+#include "print_utils.h"
+
+#define CELERITY_SET_LOG_CTX(ctx) CELERITY_DETAIL_SET_LOG_CTX(ctx)
+
+#define CELERITY_TRACE(...) SPDLOG_TRACE(*celerity::detail::active_log_ctx + __VA_ARGS__)
```
fknorr, line 21: Prepending an arbitrary string to the format string will break if it contains characters that `fmt` interprets as placeholders.

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+
+} // namespace detail
+} // namespace celerity
+
+template <>
+struct fmt::formatter<celerity::detail::log_map> {
+	constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+
+	template <typename FormatContext>
+	auto format(const celerity::detail::log_map& map, FormatContext& ctx) {
```
fknorr, line 59: Wouldn't it be easier to move this code into the `log_context` constructor instead? It appears to be the only use.

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+	using log_level = spdlog::level::level_enum;
+
+	struct log_context {
+		std::string value;
+		log_context() = default;
+		explicit log_context(const log_map& values) { value = fmt::format("[{}] ", values); }
+	};
+
+	inline const std::string null_log_ctx;
+	inline thread_local const std::string* active_log_ctx = &null_log_ctx;
```
fknorr, line 41: This could also be a stack (i.e. `std::vector`) of contexts without much additional implementation effort (the guard would then be a `log_ctx_pusher` or `push_log_ctx`).

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+
+	struct log_ctx_setter {
+		log_ctx_setter(log_context& ctx) { celerity::detail::active_log_ctx = &ctx.value; }
+		~log_ctx_setter() { celerity::detail::active_log_ctx = &celerity::detail::null_log_ctx; }
+	};
+
+#define CELERITY_DETAIL_SET_LOG_CTX(ctx)                                                                                                                       \
```
fknorr, line 48: I feel like this macro makes code using it harder to understand, not easier. I wouldn't be opposed to name this guard type explicitly, like
```c++
detail::set_log_ctx push{ctx};
```
If we decide to keep the macro, I would advocate for adding a `__COUNTER__` to the variable name.

fknorr (commented):
Thank you, human readable logs are a very welcome change! A few things:

include/command_graph.h
psalz: Good point. `std::string` would probably work. We could also include `<iosfwd>` instead, either works for me.

include/log.h
psalz: It's also used in the `user_bench` utilities!

include/log.h
psalz: True, although I guess whether we want to do something about it depends on whether this is supposed to be a documented API that users can also use.

include/log.h
psalz: Sure, but since it's being called on what is likely the hottest code path in our codebase, and we don't need it right now, I didn't see any reason to implement it that way!

include/log.h
psalz: I see your point in that it doesn't really convey it's RAII nature (which could be somewhat remedied by appending `_RAII` to the name). However I do think that mixing `SCREAM_CASE` macros with some `snake_case` type in `detail::` also doesn't make it easier to grok the connection between the two. The other alternative would be to move logging into variadic template functions that are defined as no-ops if the compile time log level doesn't match. Of course we'd then lose the option of logging source locations.

Adding the `__COUNTER__` would make sense, yes.

BlackMark29A: I agree that a consistent interface should be used, but I also feel like making the RAII mechanism visible on the call-site is a good idea.

What about something like this:

```cpp
const auto raii_ctx = CELERITY_SET_LOG_CTX(ctx);
```

include/log.h
fknorr: @psalz couldn't that be solved rather easily with
```c++
#define CELERITY_TRACE(fmt_literal, ...) SPDLOG_TRACE("[{}] " fmt_literal, *active_log_ctx ,##__VA_ARGS__)
```
@BlackMark29A I'm not sure I like that idea. This would mean our stable API surface increases by the entirety of `libfmt`.

include/log.h
fknorr: Right!

include/log.h
fknorr: @BlackMark29A that would stop us from completely removing the code at compile time.

I've been thinking, maybe it's enough to just improve the naming, e.g. to `CELERITY_DETAIL_LOG_CTX_GUARD`?

BlackMark29A: > that would stop us from completely removing the code at compile time

Why is this a problem? It won't have any runtime overhead, if it's set to an empty struct, for example.

include/log.h
psalz: I did not know of that `##__VA_ARGS__` trick - pretty neat! Alas, it seems like it's unsupported in MSVC.

src/print_graph.cc
```diff
@@ -43,24 +42,20 @@ namespace detail {
...
 				if(d.node->get_id() == 0) continue; // Skip INIT task
-				ss << fmt::format("{} -> {} [{}];", d.node->get_id(), tsk->get_id(), dependency_style(d.kind));
+				os << fmt::format("{} -> {} [{}];", d.node->get_id(), tsk->get_id(), dependency_style(d.kind));
 			}
 		}
 
-		ss << "}";
-		auto str = ss.str();
-		str = std::regex_replace(str, std::regex("\n"), "\\n");
-		str = std::regex_replace(str, std::regex("\""), "\\\"");
```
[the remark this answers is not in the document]
psalz: Because we no longer print graphs embedded in a JSON message.

src/command_graph.cc
```diff
@@ -27,11 +27,11 @@ namespace detail {
 		}
 	}
 
-	void command_graph::print_graph(logger& graph_logger) const {
+	void command_graph::print_graph(std::ostream& os) const {
 		if(command_count() < 200) {
```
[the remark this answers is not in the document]
psalz: We could, but then again the number is completely arbitrary anyway. I've been thinking whether there should be an environment variable to control this instead.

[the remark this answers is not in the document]
BlackMark29A: I think an environment variable makes sense here, so I would be in favor of that.

include/log.h
psalz: [Seems like it's supported in the new preprocessor though](https://docs.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170#comma-elision-in-variadic-macros). So we'd have to enable that through CMake; not sure how stable it is though.

fknorr: I agree we shouldn't rely on that. Is there another way to express this, maybe similarly to how the `WORKAROUND` macros are defined right now?

include/log.h
fknorr: Fair enough!

include/command_graph.h
psalz: The graph printing functions now return `std::optional<std::string>`, as they will not print anything above a (configurable) threshold.

include/log.h
psalz: Fixed through double-formatting as suggested by @fknorr during Monday's meeting.

include/log.h
psalz: Renamed to `CELERITY_LOG_SET_SCOPED_CTX`.

src/command_graph.cc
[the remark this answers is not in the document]
psalz: I've introduced a new config variable `CELERITY_GRAPH_PRINT_MAX_VERTS` to control this.k

include/command_graph.h
fknorr: Now you should be able to simply remove the `<ostream>` include.

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+	constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+
+	template <typename FormatContext>
+	auto format(const celerity::detail::log_map& map, FormatContext& ctx) {
+		auto&& out = ctx.out();
+		int i = 0;
+		for(const auto& [k, v] : map) {
+			if(i++ > 0) { fmt::format_to(out, ", "); }
+			fmt::format_to(out, "{}=", k);
+			if(std::holds_alternative<std::string_view>(v)) {
```
fknorr, line 65: Is there any benefit in using the variant here instead of making the `log_map` a `map<string, string>` and requiring the user to call `fmt::format` themselves or adding a templated insertion function to `log_map`? `string_view` is a potential footgun because it is implicitly converted from a string but requires the user to keep the backing string alive.

If this is about performance, we should allow disabling insertion into the `log_map` via macros just like we do for the rest of the logging API.

If we keep the variant, I suggest simplifying the if-statement (and the preceding `format_to`) to
```c++
std::visit([&](auto &variant) { fmt::format_to(out, "{}={}", k, variant); }, v);
```

include/task_manager.h
```diff
@@ -4,6 +4,8 @@
 #include <memory>
 #include <mutex>
 #include <optional>
+#include <ostream>
```
fknorr, line 7: Same here, `<ostream>` should not be needed.

src/print_graph.cc
```diff
@@ -1,13 +1,14 @@
 #include "print_graph.h"
 
 #include <regex>
+#include <sstream>
 
 #include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/ostr.h>
```
fknorr, line 7: Is this include required?

include/log.h
```diff
@@ -0,0 +1,75 @@
...
+#include <string_view>
+#include <unordered_map>
+#include <variant>
+
+// TODO: Make this configurable through CMake?
+#define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_TRACE
+#include <spdlog/spdlog.h>
+
+#include <spdlog/fmt/ostr.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
```
fknorr, line 15: I don't think any of

- `<ostream>`
- `<sstream>`
- `<spdlog/fmt/ostr.h>`
- `<spdlog/sinks/stdout_color_sinks.h>`

are actually used in this file.

test/unit_test_suite_celerity.cc
```diff
@@ -1,6 +1,7 @@
 #include "unit_test_suite_celerity.h"
 
-celerity::detail::logger graph_logger{"test-graph", celerity::detail::log_level::trace};
+#include <limits>
+#include <sstream>
```
fknorr, line 4: This include is not used.

fknorr (commented):
Thanks! A few more things, mostly about includes (also see my very first review comment thread on this PR). I would prefer if we kept iostreams out of as many translation units as possible because it tends to be quite expensive to compile.

src/print_graph.cc
psalz: Yes, this is required to enable formatting of types that support `operator<<(std::ostream&, T)` . I've added a clarifying comment.

include/log.h
psalz: Thanks, I've removed (hopefully) all superfluous includes here and in the other files.

include/log.h
psalz: It was a performance consideration, although creating the `unordered_map` was likely much more expensive anyway. I agree on the footgunnyness of `string_view`. I've gotten rid of the map altogether and replaced it with a `std::tuple` that can hold arbitrary types (as long as `fmt` knows how to format them).

I didn't know of `std::visit` btw, pretty neat.

psalz:
I've addressed the additional feedback by @fknorr and in that process also made some improvements to the `user_bench` utilities, along with some breaking changes (the whole thing is experimental and we don't use it in any of the examples currently, so I think it should be fine).

As mentioned in the comment above, I've mainly gotten rid of the `log_map` type as an alias for `std::unordered_map`, both to address limitations in the types that can be used, as well as to avoid the unnecessary overhead of creating a whole hash map only for it to be immediately turned into a string. Note that the `log_map` type still exists, but only as a helper to specialize `fmt::formatter`. While I could've moved the formatting of the `std::tuple` directly into the constructor of `log_context`, I didn't want to repeatedly call `fmt::format` and then concatenate everything into a `stringstream` because that seemed wasteful.

fknorr (approved):
Thank you! After some digging, the tie-in between fmt and iostreams in our code, aka `<fmt/ostr.h>` has two origins:

- `print_utils.h` and `.cpp` expose `ostream& operator<<` for the `subrange` and `chunk` types, which are internally implemented using `fmt`, so we are going back and forth between fmt and stringstreams.
- The Allscale headers define `ostream& operator<<` on the `GridBox` type. Its only use is in `region_map_testspy::print_regions`, and that is currently dead code as the only (transitive) invocation is commented out in `graph_compaction_tests:142`.

However, the major offender now is the Allscale assertions header which pulls in `<iostreams>` unconditionally. I feel like this is beyond the scope of this PR now, so I'll cease my iostreams crusade.

PR #81, opened 2022-01-25
Clean up command hierarchy around horizons

fknorr:
_This is the first in a series of PRs to integrate `SYNC` and `SHUTDOWN` commands into the TDAG/CDAG._

Before the introduction of task graph horizons, every `task` mapped to one or more `task_commands`, while horizons were generated on the CDAG directly. Now, CDAG horizons are simply the manifestation of TDAG horizons and therefore logically become `task_commands`, but so far this was not modelled as part of the command class hierarchy.

This PR extracts `task_command` as a common base of `execution_command` (the former `task_command`) and `horizon_command`, reducing code duplication in the process. It also removes some dead code (e.g. pseudo-critical path length tracking in `graph_generator`) that was missed when TDAG horizons were introduced.


test/test_utils.h
```diff
@@ -104,15 +104,16 @@ namespace test_utils {
 
 				const detail::command_id cid = pkg.cid;
 				commands[cid] = {nid, pkg, dependencies};
-				if(pkg.cmd == detail::command_type::TASK) { by_task[std::get<detail::task_data>(pkg.data).tid].insert(cid); }
+				if(pkg.cmd == detail::command_type::EXECUTION) { by_task[std::get<detail::execution_data>(pkg.data).tid].insert(cid); }
+				if(pkg.cmd == detail::command_type::HORIZON) { by_task[std::get<detail::horizon_data>(pkg.data).tid].insert(cid); }
```
psalz, line 108: And this does not break any tests? Interesting.

psalz (approved):
A net -38 LoC, LGTM!

include/command.h
```diff
@@ -119,23 +108,36 @@ namespace detail {
...
 	  public:
 		task_id get_tid() const { return tid; }
 
+	  private:
+		task_id tid;
+	};
+
+	class horizon_command final : public task_command {
+		friend class command_graph;
+		using task_command::task_command;
```
PeterTh, line 123: How neat. I wasn't actively aware of this.

include/executor.h
```diff
@@ -88,7 +88,8 @@ namespace detail {
 		template <typename Job, typename... Args>
 		void create_job(const command_pkg& pkg, const std::vector<command_id>& dependencies, Args&&... args) {
 			auto logger = execution_logger->create_context({{"job", std::to_string(pkg.cid)}});
-			if(pkg.cmd == command_type::TASK) { logger = logger->create_context({{"task", std::to_string(std::get<task_data>(pkg.data).tid)}}); }
+			if(pkg.cmd == command_type::HORIZON) { logger = logger->create_context({{"task", std::to_string(std::get<horizon_data>(pkg.data).tid)}}); }
```
PeterTh, line 91: Maybe this could be `"horizon"`?

test/test_utils.h
```diff
@@ -177,15 +179,19 @@ namespace test_utils {
...
 			if(most_recently_generated_task_horizon != most_recently_built_task_horizon) {
 				most_recently_built_task_horizon = most_recently_generated_task_horizon;
 				if(most_recently_built_task_horizon != nullptr) {
-					auto htid = most_recently_built_task_horizon->get_id();
-					get_graph_generator().build_task(htid, {nullptr});
+					const auto htid = most_recently_built_task_horizon->get_id();
+					// naive_split does not really do anything for horizons, but this mirrors the behavior of scheduler::schedule exactly.
+					detail::naive_split_transformer naive_split(num_nodes, num_nodes);
+					get_graph_generator().build_task(htid, {&naive_split});
+					return htid;
```
PeterTh, line 191: As I mentioned during the meeting, I think having to reproduce a lot of the logic in the test utils is not ideal. But I also don't have a better idea for unit tests which shouldn't have the entire runtime in the background.

PeterTh (approved):

include/command.h
fknorr: Re-exporting base-class constructors? Yeah, that is pretty neat!

test/test_utils.h
fknorr: `by_task` is only used when passing a `task_id` to `cdag_inpector::get_commands()`, which was never done for horizon tasks as far as I'm aware.

test/test_utils.h
fknorr: I think we could reasonably manage the set of active transformers in `graph_generator` and get rid of the `native_split` construction (and vector allocation) here. Thoughts @psalz ?

facuMH (approved):

test/test_utils.h
psalz: Hmm, I'm not sure. While the scheduler is pretty bare-bones at the moment, I have at various occasions experimented with multiple/different transformers depending on task types, hints etc -- which is something that I don't think the graph generator should be aware of. We can certainly revisit the current scheduler/transformer/ggen design, but I'd say its out of scope for this PR.

fknorr: Let's just keep it in mind then.

PR #83, opened 2022-02-01
Silence unused variable warning

psalz:
Also I've changed the check for `GetProcessAffinityMask` to compare against `0`. Even though the [signature](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprocessaffinitymask) says `BOOL`, the description talks about zero and non-zero return values, and I found a check `assert(error != false)` to be confusing.

### DIFF
--- a/src/platform_specific/affinity.unix.cc
+++ b/src/platform_specific/affinity.unix.cc
@@ -1,4 +1,3 @@
-
 #include <cassert>
 #include <cstdint>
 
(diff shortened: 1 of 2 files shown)


src/platform_specific/affinity.unix.cc
```diff
@@ -13,6 +12,7 @@ namespace detail {
 	uint32_t affinity_cores_available() {
 		cpu_set_t available_cores;
 		const auto affinity_error = pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &available_cores);
+		(void)affinity_error;
```
fknorr, line 15: I would like a `[[maybe_unused]]` on the declaration itself better (same below).

fknorr (changes requested):

BlackMark29A:
> I found a check `assert(error != false)` to be confusing

I agree, but I think that's the fault of a badly named variable. The variable is called `error`, and is semantically a `bool`, but if the value evaluates to `true` it is the non-error case. So I'd say that's a misnomer.

psalz:
Done. Due to a lack of creativity on my part, I've renamed the return values to `ret`.

fknorr (approved):
LGTM!

PR #84, opened 2022-02-01
Avoid creation of empty chunks

psalz:
So far, tasks with a global size that was smaller than the number of nodes would still be split into N chunks, one for each node, with some chunks having an empty range. This introduces uneccesary overhead in scheduling and execution of commands that are effectively no-ops. Additionally, since we currently do not support accessing empty buffer ranges (which is a separate issue), empty chunks with e.g. a one-to-one mapper would result in a crash.

### DIFF
--- a/src/transformers/naive_split.cc
+++ b/src/transformers/naive_split.cc
@@ -11,7 +11,7 @@ namespace celerity {
 namespace detail {
 
 	// We simply split in the first dimension for now
-	std::vector<chunk<3>> split_equal(const chunk<3>& full_chunk, const cl::sycl::range<3>& granularity, size_t num_chunks, int dims) {
+	static std::vector<chunk<3>> split_equal(const chunk<3>& full_chunk, const cl::sycl::range<3>& granularity, const size_t num_chunks, const int dims) {
 #ifndef NDEBUG
 		assert(num_chunks > 0);
 		for(int d = 0; d < dims; ++d) {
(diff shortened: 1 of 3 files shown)


src/transformers/naive_split.cc
```diff
@@ -53,21 +58,13 @@ namespace detail {
 	}
 
 	naive_split_transformer::naive_split_transformer(size_t num_chunks, size_t num_workers) : num_chunks(num_chunks), num_workers(num_workers) {
+		assert(num_workers > 0);
 		assert(num_chunks >= num_workers);
```
fknorr, line 62: This precondition is no longer necessary, maybe replace it by `num_chunks > 0`?

src/transformers/naive_split.cc
```diff
@@ -78,9 +75,19 @@ namespace detail {
 		assert(std::distance(original->get_dependents().begin(), original->get_dependents().end()) == 0);
 
 		chunk<3> full_chunk{tsk.get_global_offset(), tsk.get_global_size(), tsk.get_global_size()};
-		auto chunks = split_equal(full_chunk, tsk.get_granularity(), num_chunks, tsk.get_dimensions());
+		const auto chunks = split_equal(full_chunk, tsk.get_granularity(), num_chunks, tsk.get_dimensions());
+		assert(chunks.size() <= num_chunks); // We may have created less than requested
```
fknorr, line 79: Also, `chunks.size()` must not be 0 here.

test/system/distr_tests.cc
```diff
@@ -291,6 +291,26 @@ namespace detail {
...
+			return celerity::access::one_to_one{}(chnk);
+		};
+
+		q.submit([=](handler& cgh) {
+			accessor acc{buf, cgh, chunk_check_rm, write_only, no_init};
+			// The kernel has a size of 1 in dimension 0, so it will not be split into
+			// more than one chunk (assuming current naive split behavior).
+			cgh.parallel_for<class UKN(kernel)>(buf.get_range(), [=](item<2> it) { acc[it] = 0; });
+		});
+		q.slow_full_sync();
```
fknorr, line 312: Does the sync actually do anything here?

fknorr (approved):
LGTM! I suggest we address the case of degenerate tasks with empty execution ranges in a separate PR.

src/transformers/naive_split.cc
psalz: Done!

src/transformers/naive_split.cc
psalz: Done!

test/system/distr_tests.cc
psalz: You're right, it doesn't do anything. It was a remnant of an earlier, much hackier incarnation of this test. Removed.

src/transformers/naive_split.cc
```diff
@@ -78,9 +75,20 @@ namespace detail {
...
+		assert(!chunks.empty());
+
+		// Assign each chunk to a node
+		// We assign chunks next to each other to the same worker (if there is more chunks than workers), as this is likely to produce less
+		// transfers between tasks than a round-robin assignment (for typical stencil codes).
+		// FIXME: This only works if the number of chunks is an integer multiple of the number of workers, e.g. 3 chunks for 2 workers degrades to RR.
```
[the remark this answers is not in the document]
psalz: I didn't want to address this in this PR since it's completely unrelated; I just added that note so we don't forget. Thanks, I will look into it.

BlackMark29A (approved):
Looks good to me.

Issue #87, opened 2022-02-08
catch2 sigStackSize errors


maedoc:
Building against hipSYCL, CUDA 11.6, clang-12, Ubuntu 21.10, the build errors on 
```
[ 63%] Built target convolution
[ 65%] Building CXX object test/CMakeFiles/unit_test_suite.dir/unit_test_suite.cc.o
clang: warning: Unknown CUDA version. cuda.h: CUDA_VERSION=11060. Assuming the latest supported version 10.1 [-Wunknown-cuda-version]
In file included from /home/duke/src/celerity-runtime/test/unit_test_suite.cc:4:
/home/duke/src/celerity-runtime/vendor/Catch2/single_include/catch2/catch.hpp:10675:34: error: constexpr variable 'sigStackSize' must be initialized by a constant expression
    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
                                 ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/duke/src/celerity-runtime/vendor/Catch2/single_include/catch2/catch.hpp:10675:58: note: non-constexpr function 'sysconf' cannot be used in a constant expression
    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
                                                         ^
/usr/include/x86_64-linux-gnu/bits/sigstksz.h:32:22: note: expanded from macro 'MINSIGSTKSZ'
# define MINSIGSTKSZ SIGSTKSZ
                     ^
/usr/include/x86_64-linux-gnu/bits/sigstksz.h:28:19: note: expanded from macro 'SIGSTKSZ'
# define SIGSTKSZ sysconf (_SC_SIGSTKSZ)
                  ^
/usr/include/unistd.h:640:17: note: declared here
extern long int sysconf (int __name) __THROW;
                ^
In file included from /home/duke/src/celerity-runtime/test/unit_test_suite.cc:4:
/home/duke/src/celerity-runtime/vendor/Catch2/single_include/catch2/catch.hpp:10675:34: error: dynamic initialization is not supported for __device__, __constant__, __shared__, and __managed__ variables.
    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
[...]

PR #88, opened 2022-02-08
Update catch2 submodule to v2.13.8

PeterTh:
This is a version bump of our catch2 submodule to the latest numbered version which should not have any interface/functionality changes. 
This should hopefully resolve https://github.com/celerity/celerity-runtime/issues/87, which is caused by https://github.com/catchorg/Catch2/issues/2178


psalz (approved):

PeterTh:
Note: the 2 failures of the hipsycl:HEAD (forward-looking) CI builds are completely unrelated to this change, so I'll ignore that for the purpose of this PR.

PR #89, opened 2022-02-09
Support empty buffers and buffer requirements with empty ranges

fknorr:
Currently, Celerity does not support buffers or buffer accesses where `get_range().size() == 0`. This edge case can be encountered by using range mappers that produce empty ranges for some chunks. This PR enables creating empty buffers and empty buffer requirements while ensuring that no unnecessary dependencies or data transfers are generated.

The ComputeCpp and DPC++ backends do not support empty buffers or empty-range accessors themselves. As a workaround, we allocate unit-sized backing buffers on the device when necessary and verify that these allocations are not included in the bounding box of subsequent backing buffer resizes:

- `detail::device_buffer_storage::make_device_buf_effective_range()` ensures that `device_buf.get_range().size()` is always `>= 1` 
- `detail::get_effective_sycl_accessor_subrange()` re-maps empty-ranged accesses to unit-ranged accesses starting at the beginning of the backing buffer

### DIFF
--- a/include/buffer_manager.h
+++ b/include/buffer_manager.h
@@ -206,27 +206,27 @@ namespace detail {
 #endif
 			assert((range_cast<3>(offset + range) <= buffer_infos.at(bid).range) == cl::sycl::range<3>(true, true, true));
 
-			auto& old_buffer = buffers[bid].device_buf;
-			backing_buffer new_buffer;
+			auto& existing_buf = buffers[bid].device_buf;
+			backing_buffer replacement_buf;
 
-			if(!old_buffer.is_allocated()) {
-				new_buffer = backing_buffer{std::make_unique<device_buffer_storage<DataT, Dims>>(range_cast<Dims>(range), queue.get_sycl_queue()), offset};
+			if(!existing_buf.is_allocated()) {
+				replacement_buf = backing_buffer{std::make_unique<device_buffer_storage<DataT, Dims>>(range_cast<Dims>(range), queue.get_sycl_queue()), offset};
 			} else {
 				// FIXME: For large buffers we might not be able to store two copies in device memory at once.
 				// Instead, we'd first have to transfer everything to the host and free the old buffer before allocating the new one.
 				// TODO: What we CAN do however already is to free the old buffer early iff we're requesting a discard_* access!
 				// (AND that access request covers the entirety of the old buffer!)
-				const auto info = is_resize_required(old_buffer, range, offset);
+				const auto info = is_resize_required(existing_buf, range, offset);
 				if(info.resize_required) {
-					new_buffer = backing_buffer{
+					replacement_buf = backing_buffer{
 					    std::make_unique<device_buffer_storage<DataT, Dims>>(range_cast<Dims>(info.new_range), queue.get_sycl_queue()), info.new_offset};
 				}
 			}
 
-			audit_buffer_access(bid, new_buffer.is_allocated(), mode);
+			audit_buffer_access(bid, replacement_buf.is_allocated(), mode);
 
-			if(test_mode && new_buffer.is_allocated()) {
-				auto device_buf = static_cast<device_buffer_storage<DataT, Dims>*>(new_buffer.storage.get())->get_device_buffer();
+			if(test_mode && replacement_buf.is_allocated()) {
+				auto device_buf = static_cast<device_buffer_storage<DataT, Dims>*>(replacement_buf.storage.get())->get_device_buffer();
 
 				// We need two separate approaches here for hipSYCL and ComputeCpp, as hipSYCL currently (0.9.1) does
 				// not support reinterpreting buffers to other dimensionalities, while ComputeCpp (2.5.0) does not
@@ -251,17 +251,9 @@ namespace detail {
 				    .wait();
 			}
 
-			if(range.size() != 0) {
-				backing_buffer& target_buffer = new_buffer.is_allocated() ? new_buffer : old_buffer;
-				const backing_buffer empty{};
-				const backing_buffer& previous_buffer = new_buffer.is_allocated() ? old_buffer : empty;
-				make_buffer_subrange_coherent(bid, mode, target_buffer, {offset, range}, previous_buffer);
-			}
+			existing_buf = make_buffer_subrange_coherent(bid, mode, std::move(existing_buf), {offset, range}, std::move(replacement_buf));
 
-			if(new_buffer.is_allocated()) { buffers[bid].device_buf = std::move(new_buffer); }
-
-			return {dynamic_cast<device_buffer_storage<DataT, Dims>*>(buffers[bid].device_buf.storage.get())->get_device_buffer(),
-			    id_cast<Dims>(buffers[bid].device_buf.offset)};
+			return {dynamic_cast<device_buffer_storage<DataT, Dims>*>(existing_buf.storage.get())->get_device_buffer(), id_cast<Dims>(existing_buf.offset)};
 		}
 
 		template <typename DataT, int Dims>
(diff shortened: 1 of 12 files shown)


PeterTh (approved):
Seems fine overall, though I'm not perfectly happy about introducing this much complexity for this feature (that's not an implementation complaint though).

src/task_manager.cc
```diff
@@ -119,9 +119,9 @@ namespace detail {
 			if(std::any_of(modes.cbegin(), modes.cend(), detail::access::mode_traits::is_producer) || reduction.has_value()) {
 				auto write_requirements = get_requirements(tsk.get(), bid, {detail::access::producer_modes.cbegin(), detail::access::producer_modes.cend()});
 				if(reduction.has_value()) { write_requirements = GridRegion<3>::merge(write_requirements, GridRegion<3>{{1, 1, 1}}); }
-				assert(!write_requirements.empty() && "Task specified empty buffer range requirement. This indicates potential anti-pattern.");
-				const auto last_writers = buffers_last_writers.at(bid).get_region_values(write_requirements);
+				if(write_requirements.empty()) continue;
```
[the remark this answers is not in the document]
fknorr: I am torn on this as well. Specifying an empty requirement explicitly is probably a user error for which I would like to print a warning. If an empty requirement originates from generic code though, it might be a necessary corner case that is difficult to eliminate from the user's perspective - they can't simply make the creation and capture of an accessor conditional.

include/buffer_manager.h
```diff
@@ -291,10 +292,12 @@ namespace detail {
...
-			backing_buffer& target_buffer = new_buffer.is_allocated() ? new_buffer : old_buffer;
-			const backing_buffer empty{};
-			const backing_buffer& previous_buffer = new_buffer.is_allocated() ? old_buffer : empty;
-			make_buffer_subrange_coherent(bid, mode, target_buffer, {offset, range}, previous_buffer);
+			if(range.size() != 0) {
+				backing_buffer& target_buffer = new_buffer.is_allocated() ? new_buffer : old_buffer;
+				const backing_buffer empty{};
+				const backing_buffer& previous_buffer = new_buffer.is_allocated() ? old_buffer : empty;
+				make_buffer_subrange_coherent(bid, mode, target_buffer, {offset, range}, previous_buffer);
+			}
```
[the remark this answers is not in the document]
psalz: Have you taken a look at `make_buffer_subrange_coherent`? I don't think that function needs any additional complexity :P

src/task_manager.cc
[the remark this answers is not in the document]
psalz: I agree that while unlikely (especially since we pretty much only support `one_to_one` writing access right now), we cannot rule this out in the general case, and so I'd say such a diagnostic is best left for tooling (cc @facuMH).

test/runtime_tests.cc
```diff
@@ -2332,5 +2332,92 @@ namespace detail {
...
+		int init = 0;
+		SECTION("when the buffer is uninitialized") { buf = buffer<int>{1}; };
+		SECTION("when the buffer is host-initialized") { buf = buffer<int>{&init, 1}; };
+
+		test_empty_access<access_mode::discard_write>(q, *buf);
+		test_empty_access<access_mode::read_write>(q, *buf);
+		test_empty_access<access_mode::read>(q, *buf);
+	}
+
+	TEST_CASE_METHOD(test_utils::buffer_manager_fixture, "empty access ranges do not inflate backing buffer allocations", "[buffer_manager]") {
```
psalz, line 2383: Nice test case, I like the use of generators here!

test/graph_generation_tests.cc
```diff
@@ -1459,5 +1459,87 @@ namespace detail {
 		maybe_print_graphs(ctx);
 	}
 
+
+	TEST_CASE("buffer accesses with empty ranges do not generate pushes or data-flow dependencies", "[graph_generator][command-graph]") {
```
psalz, line 1463: I guess there is some overlap here with `"graph_generator allows chunks to require empty buffer ranges"`, but that's fine.

test/sycl_tests.cc
```diff
@@ -62,4 +62,90 @@ TEMPLATE_TEST_CASE_METHOD_SIG(dim_device_queue_fixture, "ranged_sycl_access work
...
+constexpr auto sycl_target_device = cl::sycl::access::target::global_buffer;
+#pragma GCC diagnostic pop
+#else
+constexpr auto sycl_target_device = cl::sycl::access::target::device;
+#endif
+
+template <access_mode AccessMode, bool UsingPlaceholderAccessor>
+static auto make_device_accessor(sycl::buffer<int, 1>& buf, sycl::handler& cgh, const subrange<1>& sr) {
+	if constexpr(UsingPlaceholderAccessor) {
+		sycl::accessor<int, 1, AccessMode, sycl_target_device, sycl::access::placeholder::true_t> acc{buf, sr.range, sr.offset};
```
psalz, line 81: I don't know if this was intentional since I assume not all implementations have caught up yet, but the `placeholder_t` template parameter is a no-op in SYCL 2020.

psalz (approved):
Very nice, seems like you covered all the bases here. LGTM!

include/buffer_manager.h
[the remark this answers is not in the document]
fknorr: Since each of the 3 callsites had effectively the same boilerplate code, I bit the bullet and chaged the API of `make_buffer_subrange_coherent`.

test/graph_generation_tests.cc
fknorr: You're right, that's basically a duplicate. I wasn't aware this was explicitly supported before at all. We even ran into the same issues with the range mappers while writing these tests!
```c++
// NOTE: It's important to construct range-mappers in such a way that passing the
// global size (during tdag generation) still returns the correct result!
```
Should we get rid of one of them, or are tests "the more the merrier"?

test/sycl_tests.cc
fknorr: That's duplicated from accessor.h lines 286 f. hipSYCL seems to be the only implementation to have removed `placeholder_t`.

BlackMark29A (approved):
Looks good to me.

test/graph_generation_tests.cc
```diff
@@ -1459,5 +1459,87 @@ namespace detail {
...
+			for(const auto cid : read_cids) {
+				const auto cmd = cdag.get(cid);
+				read_cmds[cmd->get_nid()] = cmd;
+			}
+		}
+
+		CHECK(!inspector.has_dependency(read_cmds[0]->get_cid(), write_cmds[0]->get_cid()));
+		CHECK(!inspector.has_dependency(read_cmds[1]->get_cid(), write_cmds[1]->get_cid()));
+
+		const abstract_command* push_cmd;
```
psalz, line 1521: This (and `await_push_cmd`) doesn't need to live outside the block.

test/graph_generation_tests.cc
```diff
@@ -1459,5 +1459,87 @@ namespace detail {
...
+		{
+			const auto pushes = inspector.get_commands(std::nullopt, std::nullopt, command_type::PUSH);
+			REQUIRE(pushes.size() == 1);
+			push_cmd = cdag.get(*pushes.begin());
+			CHECK(push_cmd->get_nid() == 0);
+			const auto push_dependencies = push_cmd->get_dependencies();
+			REQUIRE(std::distance(push_dependencies.begin(), push_dependencies.end()) == 1);
+			CHECK(push_dependencies.begin()->node == write_cmds[0]);
```
psalz, line 1529: What's the reason we need to check here that there's only one predecessor, and that it's the write command? That seems pretty unrelated to me and should already be covered by other tests. (Same below.)

psalz (changes requested):

test/graph_generation_tests.cc
psalz: Hmm I think having multiple tests is probably only confusing in the long run, so I'd say remove one of them. I'm fine with going with your new version, although I will say that I found the range mapper definition somewhat hard to understand (without running it).

test/graph_generation_tests.cc
fknorr: Not really necessary, you're right: If there's one push and one await push, each on different nodes, the rest is evident. I've simplified this.

test/graph_generation_tests.cc
fknorr: Fixed.

test/graph_generation_tests.cc
fknorr: I've removed the old test, replaced one of my RMs with `one_to_one` and made the other one more readable.

src/task_manager.cc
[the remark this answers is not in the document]
fknorr: I'm going to continue without the diagnostic then.

psalz (approved):
Ship it!

PR #92, opened 2022-02-14
Disallow submitting tasks with an empty execution range

fknorr:
Graceful handling of tasks with an empty execution range is non-trivial in the presence of reductions that must initialize their output buffer independently of the kernel. Since we cannot reliably submit empty chunks to SYCL, we throw on task creation instead, expecting the user to omit the submission instead. Proper support for empty execution ranges will be added in the future.


psalz (approved):
LGTM

PR #93, opened 2022-02-14
Suppress DPC++ deprecation warnings about sycl::atomic

fknorr:
DPC++ seems to have newly deprecated the entire type `sycl::atomic<T>`. We re-export and deprecate `atomic<T> accessor::operator[]` ourselves, so the deprecation warning needs to be disabled for the whole method definition.

Note that the newest DPC++ still produces warnings in its the SYCL RT (`-Wsycl-strict`), but these seem to be a bug in DPC++ so I consider them out of scope.


psalz (approved):
LGTM

BlackMark29A (approved):
Looks good to me.

PR #94, opened 2022-02-21
Capture buffer and host-object data on synchronization points

fknorr:
Currently, there is no native way to get results out of a Celerity buffer and back into the main thread at the end of a Celerity program. The current workaround is `celerity::allow_by_ref` and reference captures in host tasks, which is inelegant and error-prone.

SYCL offers `host_accessor` and explicit copy operations together with awaitable kernel events for that purpose. This is not a good fit for Celerity, since stalling the main thread is orders of magnitude more expensive in the distributed case.

This PR introduces Captures, a declarative API that allows requesting host objects and buffer subranges from `distr_queue::slow_full_sync()` and from runtime shutdown via the new `distr_queue::drain()` function. Specifying a capture adds the necessary dependencies and data transfers to the associated Epoch command and copies or moves the data out to the calling main thread once the epoch is reached, confining stalls to APIs which the user already expects to be slow.

## Example

Drains the queue on program exit to receive a verification result in the main thread.

```c++
int main() {
    distr_queue q;
    host_object<bool> verification_passed;

    q.submit([=](handler &cgh) {
        side_effect verify{verification_passed, cgh};
        cgh.host_task(on_master_node, [=] {
            *verify = ...;
        });
    });

    return q.drain(capture{verification_passed}) ? 0 : 1;
```


include/capture.h
```diff
@@ -0,0 +1,167 @@
...
+
+} // namespace celerity::detail
+
+namespace celerity::experimental {
+
+template <typename>
+class capture;
+
+template <typename T, int Dims>
+class buffer_data {
```
PeterTh, line 55: Maybe we can think of a more descriptive name for this.
Maybe `buffer_capture_data`?

include/capture.h
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	value_type exfiltrate_by_move() const { return exfiltrate_by_copy(); }
+};
+
+template <typename T, int Dims>
+capture(buffer<T, Dims>) -> capture<buffer<T, Dims>>;
+
+template <typename T, int Dims>
+capture(buffer<T, Dims>, subrange<Dims>) -> capture<buffer<T, Dims>>;
```
PeterTh, line 139: Are these necessary?
(Probably "yes" in C++17 and "no" in 20?)

include/distr_queue.h
```diff
@@ -86,6 +124,17 @@ class distr_queue {
 		}
 		tracker = std::make_shared<detail::distr_queue_tracker>();
 	}
+
+	detail::runtime::sync_guard slow_full_sync(detail::buffer_capture_map buffer_captures, detail::side_effect_map side_effects) {
+		check_not_drained(tracker);
+		return detail::runtime::get_instance().sync(std::move(buffer_captures), std::move(side_effects));
+	}
+
+	void drain(detail::buffer_capture_map buffer_captures, detail::side_effect_map side_effects) {
```
PeterTh, line 133: I think this should be called something like `drain_internal`, to make things easier when reading the public `drain`s.

src/graph_generator.cc
```diff
@@ -141,7 +151,8 @@ namespace detail {
 			reduce_execution_front_to(epoch);
 		}
 
-		min_epoch_for_new_commands = min_new_epoch;
+		min_epoch_for_new_commands = epochs[0]->get_cid();
```
PeterTh, line 154: Is this intentionally `epochs[0]` for all nodes? If so, why?

include/task_manager.h
```diff
@@ -207,6 +214,9 @@ namespace detail {
 		// The last epoch task that has been processed by the executor. Behind a monitor to allow awaiting this change from the main thread.
 		epoch_monitor latest_epoch_reached{initial_epoch_task};
 
+		// The last barrier epoch that was completed from the main thread by calling resume_after_barrier.
+		epoch_monitor latest_barrier_passed{initial_epoch_task};
```
PeterTh, line 218: Is it necessary to track this independently?

PeterTh (commented):
Some comments for now -- note that I haven't looked through the unit tests yet.

Also, I didn't add code comments for this, but I'm once again unhappy with the increasing signature bloat related to tasks ;)

include/capture.h
fknorr: How about `buffer_snapshot`?

include/capture.h
fknorr: Both, see https://godbolt.org/z/r6PcaTP4q

I believe there are no implicit CTAD guides for these constructors because they are not on the primary template.

src/graph_generator.cc
fknorr: This is the global minimum epoch command id (as opposed to the per-node epoch, which is set in `set_epoch_for_new_commands`). The vector is ordered by command id, as is verified by the assertion on the following line.

include/task_manager.h
fknorr: This synchronizes the main thread with the executor, but in the opposite direction of `latest_epoch_reached`. The main thread will advance the monitor once it has exited its "capture extraction" region. These events do not occur at the same time, so they can't re-use the monitor.

include/distr_queue.h
fknorr: Done.

fknorr:

psalz (changes requested):
Thanks! I've added some notes - better late than never, right?

Some additional remarks:
- As discussed in person, the `std::in_place` constructor of `host_object` currently forwards to the initializer-list constructor of `T`, if one exists, which is unexpected.
- I don't fully understand why the `capture` object exists, or rather, why can't we pass host objects / buffers into `drain` directly? Is it only for specifying the buffer range?
- I'm not sure about the naming of `drain`. To me this doesn't really imply that this is the last operation I'm allowed to do on a queue, essentially destroying it. It sounds more like a different kind of sync, imo.
- It is also not obvious in my opinion that `slow_full_sync` produces a copy, while `drain` moves the object. Didn't an earlier version of this patch require `host_object` to be moved into `drain`?

I'm also still concerned that the semantic difference between host objects and buffers will be confusing to people (one is local to each worker, the other distributed). This difference existed before, but is now exacerbated by the fact that capturing a buffer returns the same data on all workers, whereas capturing a host object does (potentially) not.

include/task_manager.h
```diff
@@ -214,6 +221,12 @@ namespace detail {
 		// The last epoch task that has been processed by the executor. Behind a monitor to allow awaiting this change from the main thread.
 		epoch_monitor m_latest_epoch_reached{initial_epoch_task};
 
+		// The last barrier epoch that was completed from the main thread by calling resume_after_barrier.
+		epoch_monitor m_latest_barrier_passed{initial_epoch_task};
+
+		// The last epoch that was used in task pruning after being reached. This allows skipping the pruning step if no new epoch was completed since.
+		task_id m_last_pruned_before{initial_epoch_task};
```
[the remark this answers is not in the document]
fknorr: Good catch!

include/runtime.h
```diff
@@ -40,6 +40,19 @@ namespace detail {
 		friend struct runtime_testspy;
 
 	  public:
+		class [[nodiscard]] sync_guard {
```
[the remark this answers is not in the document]
fknorr: Done.

test/graph_generation_tests.cc
```diff
@@ -792,5 +794,89 @@ namespace detail {
...
+		auto& inspector = ctx.get_inspector();
+		auto& cdag = ctx.get_command_graph();
+
+		auto buf = mbf.create_buffer(range<3>{5, 4, 7});
+		auto obj = mhof.create_host_object();
+
+		const auto tid_effect = test_utils::build_and_flush(ctx, num_nodes, test_utils::add_host_task(tm, experimental::collective, [&](handler& cgh) {
+			obj.add_side_effect(cgh, experimental::side_effect_order::sequential);
+		}));
+		const auto tid_effect_and_write = test_utils::build_and_flush(ctx, num_nodes,
```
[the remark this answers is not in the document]
fknorr: Applied.

test/graph_generation_tests.cc
```diff
@@ -792,5 +794,89 @@ namespace detail {
...
+			const auto deps = sync_cmd->get_dependencies();
+			CHECK(std::distance(deps.begin(), deps.end()) == 3 - static_cast<int>(sync_cmd->get_nid()));
+			for(const auto& d : deps) {
+				if(d.origin == dependency_origin::dataflow) {
+					if(sync_cmd->get_nid() == 0) {
+						REQUIRE(isa<task_command>(d.node));
+						const auto tid = static_cast<task_command*>(d.node)->get_tid();
+						CHECK((tid == tid_effect || tid == tid_effect_and_write));
+					} else {
+						if(!isa<await_push_command>(d.node)) {
```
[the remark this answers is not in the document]
fknorr: Done.

include/distr_queue.h
```diff
@@ -82,7 +90,37 @@ class distr_queue {
 	 * In production, it should only be used at very coarse granularity (second scale).
 	 * @warning { This is very slow, as it drains all queues and synchronizes accross the entire cluster. }
 	 */
-	void slow_full_sync() { detail::runtime::get_instance().sync(); } // NOLINT(readability-convert-member-functions-to-static)
+	void slow_full_sync() { (void)slow_full_sync({}, {}); }
+
+	template <typename T>
+	typename experimental::capture<T>::value_type slow_full_sync(const experimental::capture<T>& cap) {
```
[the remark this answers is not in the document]
fknorr: Done.

include/distr_queue.h
```diff
@@ -82,7 +90,37 @@ class distr_queue {
...
+		auto [buffer_captures, side_effects] = detail::capture_inspector::collect_requirements(std::tuple{cap});
+		const auto sync_guard = slow_full_sync(std::move(buffer_captures), std::move(side_effects));
+		return std::get<0>(detail::capture_inspector::exfiltrate_by_copy(cap));
```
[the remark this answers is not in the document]
fknorr: That was indeed broken! Fixed (and tested).

examples/matmul/matmul.cc
```diff
@@ -84,9 +76,6 @@ int main() {
 	multiply(queue, mat_a_buf, mat_b_buf, mat_c_buf);
 	multiply(queue, mat_b_buf, mat_c_buf, mat_a_buf);
 
-	bool verification_passed = true;
-	verify(queue, mat_a_buf, verification_passed);
-	queue.slow_full_sync(); // Wait for verification_passed to become available
-
-	return verification_passed ? EXIT_SUCCESS : EXIT_FAILURE;
+	auto mat_a_dump = queue.drain(celerity::experimental::capture{mat_a_buf});
```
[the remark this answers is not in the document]
fknorr: I've returned to distributed verification, but replaced the ref-capture / sync with a `drain(capture(host_object))`. I feel like `syncing` was made to demonstrate the old ref-capture + `slow_full_sync` workaround. Maybe we should drop this example altogether.

fknorr:
>  As discussed in person, the `std::in_place` constructor of `host_object` currently forwards to the initializer-list constructor of `T`, if one exists, which is unexpected.

I have replaced the "universal" initializer syntax with the `()`-syntax for non-aggregate types in places related to captures / host objects.
 
> I don't fully understand why the `capture` object exists, or rather, why can't we pass host objects / buffers into `drain` directly? Is it only for specifying the buffer range?

It is about the buffer subranges, and also to clarify the meaning of arguments to `slow_full_sync` in user code.

> I'm not sure about the naming of `drain`. To me this doesn't really imply that this is the last operation I'm allowed to do on a queue, essentially destroying it. It sounds more like a different kind of sync, imo.

IMO this is a pretty common term for "waiting for things to finish" (e.g. SLURM uses it to mark nodes that are not accepting more work in order to shut down).

> It is also not obvious in my opinion that `slow_full_sync` produces a copy, while `drain` moves the object. Didn't an earlier version of this patch require `host_object` to be moved into `drain`?

I agree that the distinction is subtle and can be confusing. It also does not serve the use case of capturing a non-copyable object at a `slow_full_sync` barrier. It could be worth investigating to have the copy/move distinction instead on the `capture` level, e.g. by having an additional `capture_by_move` wrapper (this consideration only applies to host objects anyway since buffer data is always trivially copyable).

One thing that's "pretty" about the current solution is that a capture-on-drain does not introduce an additional shutdown-epoch after the capture-epoch. This is going to be pretty irrelevant from a performance perspective though.

> I'm also still concerned that the semantic difference between host objects and buffers will be confusing to people (one is local to each worker, the other distributed). This difference existed before, but is now exacerbated by the fact that capturing a buffer returns the same data on all workers, whereas capturing a host object does (potentially) not.

I still agree on this, although I would perfer to revisit host objects after merging this (we're talking about experimental features anyway).

include/capture.h
fknorr: I have renamed the struct to `buffer_snapshot` for now.

examples/matmul/matmul.cc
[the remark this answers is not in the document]
psalz: I like it! Agreed on the syncing example.

src/task_manager.cc
```diff
@@ -47,12 +47,27 @@ namespace detail {
...
+		assert(tsk->get_type() == task_type::epoch);
 		assert(!m_latest_horizon_reached || *m_latest_horizon_reached < epoch_tid);
 		assert(m_latest_epoch_reached.get() < epoch_tid);
 
 		m_latest_epoch_reached.set(epoch_tid);
 		m_latest_horizon_reached = std::nullopt; // Any non-applied horizon is now behind the epoch and will therefore never become an epoch itself
+
+		if(tsk->get_epoch_action() == epoch_action::barrier) {
+			// Wait for the main thread to call resume_after_barrier. This ensures that no commands can be executed before all captures have been exfiltrated.
+			// It is not sufficient for the main thread to stop submitting work, since e.g. AWAIT PUSH commands do not depend on a task definition.
```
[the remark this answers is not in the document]
psalz: Ahh okay I think I get it now. I think I was confused because I thought there should be an anti-dependency from an await push onto the exfiltrating epoch that prevents this race... And there is, but that's not enough because the exfiltration happens in the main thread, not the epoch worker job. So we need a way for that job to wait until the main thread is done exfiltrating. Correct?

psalz (approved):
> I still agree on this, although I would perfer to revisit host objects after merging this (we're talking about experimental features anyway).

Fine with me!

Thanks for addressing all remarks. Also I like the new `buffer_snapshot` name!

PeterTh:
Offline discussion results:
* have a `move_capture` and `capture` (copy) [ bikeshed name @psalz ]
* get rid of `drain` entirely
* in the future, implement `fence` which does not require a full barrier / epoch

fknorr:
Superseded by #151.

PR #95, opened 2022-02-23
Remove printf from runtime test

BlackMark29A:
Currently there is one runtime test which makes use of `printf`, this is not necessary, however, and the same functionality can be better implemented using the `CAPTURE` feature of catch2.  

This PR eliminates the `printf` and slightly simplifies the test.

### DIFF
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -486,22 +486,16 @@ namespace detail {
 				}
 			}
 			memcpy_strided(source_buffer.get(), target_buffer.get(), sizeof(size_t), source_range, source_offset, target_range, target_offset, copy_range);
-			bool valid = true;
 			for(size_t i = 0; i < copy_range[0]; ++i) {
 				for(size_t j = 0; j < copy_range[1]; ++j) {
 					for(size_t k = 0; k < copy_range[2]; ++k) {
 						const auto id = target_offset + cl::sycl::id<3>{i, j, k};
 						const auto source_id = source_offset + cl::sycl::id<3>{i, j, k};
-						valid &= target_buffer[get_linear_index(target_range, id)] == source_id[0] * 10000 + source_id[1] * 100 + source_id[2];
-						if(!valid) {
-							printf("Unexpected value at %lu %lu %lu: %lu != %lu\n", id[0], id[1], id[2], target_buffer[get_linear_index(target_range, id)],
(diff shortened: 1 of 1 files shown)


fknorr (approved):

test/runtime_tests.cc
```diff
@@ -486,22 +486,16 @@ namespace detail {
 				}
 			}
 			memcpy_strided(source_buffer.get(), target_buffer.get(), sizeof(size_t), source_range, source_offset, target_range, target_offset, copy_range);
-			bool valid = true;
```
psalz, line 489: The reason it was done this way was to avoid counting each iteration of the loop as a separate assertion (reported by Catch2), because those numbers (IMO) quickly become meaningless. This is a problem we currently have with the `distr_tests`, where we have 58130 assertions in 5 test cases.

A compromise might look something like this:
```c++
const auto expected = source_id[0] * 10000 + source_id[1] * 100 + source_id[2];
const auto received = target_buffer[get_linear_index(target_range, id)];
CAPTURE(id, source_id, expected, received);
if(received != expected) {
  FAIL();
}
```

Of course we then don't count any assertion for this test case.

psalz (changes requested):

test/runtime_tests.cc
fknorr: `SUCCEED()` after the loop should report one passed assertion.

On the other hand I don't fully understand why the number of passed assertions even matters. We're only interested in the absolute number of failed assertions anyway, not the ratio.

BlackMark29A: I also don't see why having `58130 assertions in 5 test cases` is actually a problem. Sure, that's a lot of assertions, but how is artificially reducing the number of assertions any more meaningful than just listing the number of individual checks that were made?

psalz: Besides the points mentioned during the last meeting (large numbers lose all meaning for humans, number of assertions depends on mostly arbitrary loop bounds), a more practical downside I found is that running tests with `-s / --success` (which I do from time to time for sanity checking) will print a metric ton of assertions, making that flag downright unusable.

I have a proof-of-concept implementation for a `REQUIRE_LOOP` macro that only logs once when called in a loop (unless it fails). Maybe this is something worth pursuing.

BlackMark29A: I like the idea of a `REQUIRE_LOOP`, so I would be in favor of pursuing that.

PR #96, opened 2022-02-23
Experimental windows support

BlackMark29A:
This PR adds experimental windows support. Currently only DPC++ as the compiler, and running with an OpenCL backend on an Intel CPU has been tested and confirmed working. Using DPC++ with a CUDA backend has also been investigated, but does not work reliably in all configurations.  

Additionally, this PR includes a simple SYCL reductions test with 1024 elements, because using an Intel i7 4790k with DPC++ on windows results in an `CL_OUT_OF_RESOURCES` exception, when running even trivial reductions with more than 1023 elements. It is currently unclear whether this only happens on windows, and further testing with the same hardware on linux will be done.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,8 @@ target_link_libraries(celerity_runtime PUBLIC
   Threads::Threads
   MPI::MPI_CXX
   spdlog::spdlog
+  $<$<AND:$<BOOL:${WIN32}>,$<CONFIG:Debug>>:sycld>
+  $<$<AND:$<BOOL:${WIN32}>,$<CONFIG:Release>>:sycl>
 )
 
 # For debug builds, we set the CELERITY_DETAIL_ENABLE_DEBUG preprocessor flag,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,8 @@ target_include_directories(celerity_runtime PUBLIC
   $<INSTALL_INTERFACE:include/celerity/vendor>
 )
 
+# With DPC++ on windows the correct release (sycl) / debug (sycld) library needs to be linked,
+# because they are not ABI compatible.
 if(WIN32 AND CELERITY_SYCL_IMPL STREQUAL "DPC++")
   set(SYCL_LIB
     $<$<CONFIG:Debug>:sycld>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ if(CELERITY_SYCL_IMPL STREQUAL "ComputeCpp")
   endif()
 endif()
 
-if(MSVC)
+if(WIN32)
   list(APPEND CELERITY_CXX_FLAGS -D_CRT_SECURE_NO_WARNINGS)
 endif()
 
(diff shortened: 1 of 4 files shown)


test/sycl_tests.cc
```diff
@@ -148,4 +148,21 @@ TEST_CASE_METHOD(test_utils::device_queue_fixture, "SYCL can access empty buffer
 	}
 }
 
+#if CELERITY_FEATURE_SIMPLE_SCALAR_REDUCTIONS
+
+TEST_CASE_METHOD(test_utils::device_queue_fixture, "SYCL has working simple scalar reductions", "[sycl][reductions]") {
```
fknorr, line 153: Does this still fail anywhere? If so, maybe add a comment saying that if this test fails, Celerity can't reliably support reductions on the user's combination of backend and hardware.

CMakeLists.txt
```diff
@@ -220,6 +220,8 @@ target_link_libraries(celerity_runtime PUBLIC
   Threads::Threads
   MPI::MPI_CXX
   spdlog::spdlog
+  $<$<AND:$<BOOL:${WIN32}>,$<CONFIG:Debug>>:sycld>
+  $<$<AND:$<BOOL:${WIN32}>,$<CONFIG:Release>>:sycl>
```
fknorr, line 224: This looks DPC++ specific, maybe wrap in a `if(CELERITY_SYCL_IMPL STREQUAL "DPC++")`?

test/sycl_tests.cc
```diff
@@ -148,4 +148,21 @@ TEST_CASE_METHOD(test_utils::device_queue_fixture, "SYCL can access empty buffer
...
+	constexpr size_t N = 1024;
```
fknorr, line 154: I remember this test being sensitive to `N`. If that's correct we could run it for multiple values of `N` with  `GENERATE`+`CAPTURE`.

test/sycl_tests.cc
BlackMark29A: Yes, it does still fail. I added the comment.

CMakeLists.txt
BlackMark29A: I'm not sure if this is DPC++ or windows specific, and it's hard to test. But I wrapped it anyway, and this check can be extended later if need be.

CMakeLists.txt
```diff
@@ -216,10 +216,18 @@ target_include_directories(celerity_runtime PUBLIC
   $<INSTALL_INTERFACE:include/celerity/vendor>
 )
 
+if(WIN32 AND CELERITY_SYCL_IMPL STREQUAL "DPC++")
```
psalz, line 219: Maybe you could add a note here as to why this is even needed (most likely a bug in DPC++, right?)

psalz (approved):
LGTM, consider squashing some of the commits.

CMakeLists.txt
BlackMark29A: I'm not sure if it's a bug in DPC++. Linking against the correct debug/release library is a windows thing, because debug/release are not ABI compatible on windows. Maybe DPC++ could do this automatically with cmake, but I don't know if that's the intended behavior.

But I can of course add a comment regardless.

fknorr (approved):

PR #97, opened 2022-02-23
Fix cpu affinity runtime test permanently setting affinity of 1

BlackMark29A:
This PR is based on #96, so that one should be merged first.  

The cpu affinity runtime test currently sets the cpu affinity of the process to 1, but never restores the original affinity, leading to warnings in subsequent tests about not enough assigned cpu cores. This PR fixes this by introducing a test fixture that uses RAII to store the original affinity on test startup, and restore it after the test has been completed.

### DIFF
--- a/test/test_utils.h
+++ b/test/test_utils.h
@@ -475,19 +475,18 @@ namespace test_utils {
 
 	class affinity_fixture {
 	  public:
-		affinity_fixture() : process_handle(GetCurrentProcess()) {
+		affinity_fixture() {
 			[[maybe_unused]] DWORD_PTR system_mask;
-			auto ret = GetProcessAffinityMask(process_handle, &process_mask, &system_mask);
+			auto ret = GetProcessAffinityMask(GetCurrentProcess(), &process_mask, &system_mask);
 			REQUIRE(ret != FALSE);
 		}
 
 		~affinity_fixture() {
-			const auto ret = SetProcessAffinityMask(process_handle, process_mask);
+			const auto ret = SetProcessAffinityMask(GetCurrentProcess(), process_mask);
 			REQUIRE(ret != FALSE);
 		}
 
 	  private:
-		const HANDLE process_handle;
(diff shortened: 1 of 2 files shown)


test/test_utils.h
```diff
@@ -463,6 +471,47 @@ namespace test_utils {
 		std::unique_ptr<detail::buffer_manager> bm;
 	};
 
+#ifdef _WIN32
+
+	class affinity_fixture {
```
fknorr, line 476: I vote for moving this to `runtime_tests.cc` since that's the only place where the fixture is used.

Also IMO, `restore_process_affinity_fixture` or `reset_process_affinity_fixture` would be more descriptive names for this class.

Also, pedantic: copy ctors/`operator=`s should be deleted.

fknorr (changes requested):

test/test_utils.h
```diff
@@ -463,6 +471,47 @@ namespace test_utils {
...
+			REQUIRE(ret != FALSE);
+		}
+
+		~affinity_fixture() {
+			const auto ret = SetProcessAffinityMask(process_handle, process_mask);
+			REQUIRE(ret != FALSE);
+		}
+
+	  private:
+		const HANDLE process_handle;
```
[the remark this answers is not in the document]
BlackMark29A: I have no preference either way, so: done.

test/test_utils.h
BlackMark29A: > I vote for moving this to runtime_tests.cc since that's the only place where the fixture is used.

I just put this fixture with the other fixtures, I thought that's where they go. But I agree that it makes more sense in `runtime_tests.cc`, so I moved it there.

Also renamed it and deleted the unused ctors/operators.

fknorr (approved):
LGTM, thanks!

psalz (approved):
Quite a lot of ceremony, but makes sense. Thanks! Please squash!

PR #98, opened 2022-02-23
Named threads for better debugging

BlackMark29A:
This PR is based on #96, so that one should be merged first.  

This PR assigns names to all threads used by celerity for better and easier debugging. Because the `pthread` API limits thread names to only 15 characters, the names were kept relatively short. Especially the host task worker threads are only named `workerN` (where `N` is the thread number inside the thread pool), even though they could also include the MPI communicator name, as celerity uses a thread pool per communicator. This means that there will be worker threads belonging to different communicators, but sharing the same name.

### DIFF
--- a/include/host_queue.h
+++ b/include/host_queue.h
@@ -10,6 +10,7 @@
 #include <mpi.h>
 
 #include "config.h"
+#include "named_threads.h"
 #include "types.h"
 
 namespace celerity {
@@ -159,7 +160,12 @@ namespace detail {
 			MPI_Comm comm;
 			ctpl::thread_pool thread;
 
-			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {}
+			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {
+				for(size_t i = 0; i < n_threads; ++i) {
+					auto& worker = thread.get_thread(i);
+					set_thread_name(worker, "worker" + std::to_string(i));
+				}
+			}
 		};
 
 		std::unordered_map<collective_group_id, comm_thread> threads;
--- a/include/host_queue.h
+++ b/include/host_queue.h
@@ -114,7 +114,7 @@ namespace detail {
 
 		host_queue() {
 			// TODO what is a good thread count for the non-collective thread pool?
-			threads.emplace(std::piecewise_construct, std::tuple{0}, std::tuple{MPI_COMM_NULL, 4});
+			threads.emplace(std::piecewise_construct, std::tuple{0}, std::tuple{MPI_COMM_NULL, 4, id++});
 		}
 
 		void require_collective_group(collective_group_id cgid) {
@@ -122,7 +122,7 @@ namespace detail {
 			assert(cgid != 0);
 			MPI_Comm comm;
 			MPI_Comm_dup(MPI_COMM_WORLD, &comm);
-			threads.emplace(std::piecewise_construct, std::tuple{cgid}, std::tuple{comm, 1});
+			threads.emplace(std::piecewise_construct, std::tuple{cgid}, std::tuple{comm, 1, id++});
 		}
 
 		template <typename Fn>
@@ -160,15 +160,16 @@ namespace detail {
 			MPI_Comm comm;
 			ctpl::thread_pool thread;
 
-			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {
+			comm_thread(MPI_Comm comm, size_t n_threads, size_t id) : comm(comm), thread(n_threads) {
 				for(size_t i = 0; i < n_threads; ++i) {
 					auto& worker = thread.get_thread(i);
-					set_thread_name(worker.native_handle(), "worker" + std::to_string(i));
+					set_thread_name(worker.native_handle(), fmt::format("cy-worker-{}.{}", id, i));
 				}
 			}
 		};
 
 		std::unordered_map<collective_group_id, comm_thread> threads;
+		size_t id = 0;
 	};
 
 } // namespace detail
--- a/include/host_queue.h
+++ b/include/host_queue.h
@@ -163,7 +163,7 @@ namespace detail {
 			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {
 				for(size_t i = 0; i < n_threads; ++i) {
 					auto& worker = thread.get_thread(i);
-					set_thread_name(worker, "worker" + std::to_string(i));
+					set_thread_name(worker.native_handle(), "worker" + std::to_string(i));
 				}
 			}
 		};
--- a/include/named_threads.h
+++ b/include/named_threads.h
@@ -5,12 +5,10 @@
 
 namespace celerity::detail {
 
-void set_thread_name(std::thread& thread, const std::string& name);
+std::thread::native_handle_type get_current_thread_handle();
 
-void set_current_thread_name(const std::string& name);
+void set_thread_name(const std::thread::native_handle_type thread_handle, const std::string& name);
 
-std::string get_thread_name(std::thread& thread);
-
-std::string get_current_thread_name();
+std::string get_thread_name(const std::thread::native_handle_type thread_handle);
 
 } // namespace celerity::detail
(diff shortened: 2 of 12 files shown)


include/named_threads.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <string>
+#include <thread>
+
+namespace celerity::detail {
+
+void set_thread_name(std::thread& thread, const std::string& name);
```
fknorr, line 8: This API could be made more concise with
```c++
std::thread::native_handle_type get_current_thread();
void set_thread_name(std::thread::native_handle_type, const std::string&);
std::string get_thread_name(std::thread::native_handle_type);
```

Then you also wouldn't need the separate `static set/get_thread_name_unix/win` functions in the implementation.

src/platform_specific/named_threads.unix.cc
```diff
@@ -0,0 +1,45 @@
...
+	truncated_name.resize(PTHREAD_MAX_THREAD_NAME_LEN - 1); // -1 because of null terminator
+	[[maybe_unused]] const auto res = pthread_setname_np(thread_handle, truncated_name.c_str());
+	assert(res == 0 && "Failed to set thread name");
+}
+
+static inline std::string get_thread_name_unix(const pthread_t thread_handle) {
+	auto name = std::vector<char>(PTHREAD_MAX_THREAD_NAME_LEN, '\0');
+	[[maybe_unused]] const auto res = pthread_getname_np(thread_handle, name.data(), name.size());
+	assert(res == 0 && "Failed to get thread name");
+	return name.data(); // Strip null terminator
```
fknorr, line 26: What happens here would be more obvious when using
```c++
char name[PTHREAD_MAX_THREAD_NAME_LEN] = {};
```
instead of the char vector (also needs one fewer include)

src/platform_specific/named_threads.win.cc
```diff
@@ -0,0 +1,69 @@
...
+
+static_assert(std::is_same_v<std::thread::native_handle_type, HANDLE>, "Unexpected native thread handle type");
+
+template <typename SrcT, typename DstT>
+static inline DstT convert_string(const SrcT& str) {
+	static_assert(
+	    (std::is_same_v<SrcT, std::string> && std::is_same_v<DstT, std::wstring>) || (std::is_same_v<SrcT, std::wstring> && std::is_same_v<DstT, std::string>),
+	    "Unsupported string type");
+
+	constexpr auto converter = [](typename DstT::value_type* dst, const typename SrcT::value_type** src, const std::size_t len, std::mbstate_t* ps) {
```
fknorr, line 19: It looks like code de-duplication has gone a bit off the rails here. Wouldn't two overloads
```c++
std::string convert_string(const std::wstring&);
std::wstring convert_string(const std::string&);
```
with mostly identical implementations (up to `mbsrtowcs` vs `wcsrtombs`) have fewer total lines of code?

src/platform_specific/named_threads.win.cc
```diff
@@ -0,0 +1,69 @@
...
+}
+
+static inline std::string get_thread_name_windows(const HANDLE thread_handle) {
+	PWSTR wname = nullptr;
+	const auto res = GetThreadDescription(thread_handle, &wname);
+	assert(SUCCEEDED(res) && "Failed to get thread name");
+	std::string name;
+	if(SUCCEEDED(res)) {
+		name = convert_string<std::wstring, std::string>(wname);
+		LocalFree(wname);
```
fknorr, line 48: `wname` will leak when `convert_string` throws. This might not be a real concern as this method is only used in tests, but maybe add a comment.

test/runtime_tests.cc
```diff
@@ -56,6 +58,20 @@ namespace detail {
 		}
 	};
 
+	struct runtime_testspy {
```
fknorr, line 61: I wonder if testing the **actual** runtime thread names is necessary. Just testing `set_current_thread_name()` and `set_thread_name()` on artificial threads created in the tests should be enough and would avoid declaring all these testspies.

fknorr (commented):
Sane overall, but `get_thread_name` is only used in tests. Maybe it can be part of `runtime_tests.cpp` instead (precedence `SetProcessAffinityMask`).

test/runtime_tests.cc
PeterTh: While all those test spies aren't great, I feel like checking the actual thread names is valuable, since setting those could feasibly break independently of the utility functions.

include/named_threads.h
BlackMark29A: Yes, that is a very good idea, thank you.

src/platform_specific/named_threads.unix.cc
BlackMark29A: I don't necessarily agree about the regular array being more obvious, but I have no strong opinion about this either way, so I changed it to a normal array.

src/platform_specific/named_threads.win.cc
BlackMark29A: This is actually what I had originally. I set out to de-duplicate this thinking it would be a one-liner to select between `mbsrtowcs`/`wcsrtombs`, which turned out not to be the case. So now I reverted it to what I had before de-duplicating.

src/platform_specific/named_threads.win.cc
BlackMark29A: The only exception that `convert_string` could throw is `bad_alloc` (unless I'm missing something), and if that happens the situation is FUBAR anyway. I briefly thought about catching/rethrowing to properly free `wname`, but I don't think it matters in this case, so I just added the comment.

include/named_threads.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <string>
+#include <thread>
+
+namespace celerity::detail {
+
+std::thread::native_handle_type get_current_thread_handle();
+
+void set_thread_name(const std::thread::native_handle_type thread_handle, const std::string& name);
```
fknorr, line 10: Pedantic, just because clang-tidy complains about this: top-level cv-qualifiers on parameters are not part of the function signature and can be removed in prototypes since they are an implementation detail.

fknorr (approved):
Thanks, LGTM now.

> Yes, there is precedence, but I don't think it's good precedence. I think having the setters and getters in the same place makes more sense, it also follows the principle of least surprise.

Im fine with that.

test/runtime_tests.cc
fknorr: Sure, I don't feel to strongly about this.

include/host_queue.h
```diff
@@ -159,7 +160,12 @@ namespace detail {
 			MPI_Comm comm;
 			ctpl::thread_pool thread;
 
-			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {}
+			comm_thread(MPI_Comm comm, size_t n_threads) : comm(comm), thread(n_threads) {
```
psalz, line 163: I'd propose to pass an "id" into here (which is simply `threads.size()`) which can then be prefixed to the worker name, i.e., `fmt::format("worker-{}.{}", id, i)`.

psalz (changes requested):
Thanks, this is great! After trying it out I would wish however that all Celerity threads would be clearly distinguishable from other threads (for example I have several CUDA threads and some other unknown ones, presumably MPI). CUDA prefixes its threads with `cuda-`, so maybe we could do `celerity-`, or if that's too long, something like `clty` or `c6y-`?

test/runtime_tests.cc
```diff
@@ -2465,5 +2481,32 @@ namespace detail {
...
+		CHECK(executor_thread_name == "executor");
+
+		const auto main_thread_name = get_thread_name(get_current_thread_handle());
+		CHECK(main_thread_name == "main");
+
+		q.submit([](handler& cgh) {
+			cgh.host_task(experimental::collective, [&](experimental::collective_partition) {
+				const auto base_name = std::string("worker");
+				const auto worker_thread_name = get_thread_name(get_current_thread_handle());
+				CHECK(worker_thread_name.compare(0, base_name.size(), base_name) == 0); // Check thread name starts with "worker"
```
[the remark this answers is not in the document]
BlackMark29A: I didn't know that existed. Thank you, and done.

include/host_queue.h
BlackMark29A: Done.

psalz (approved):

Issue #85, opened 2022-02-05
Lack of cpu_set_t on macOS


maedoc:
macOS appears to lack cpu_set_t, though it seems the use is rather limited to counting cores, so some other API could stand in on macOS, e.g. [here](https://stackoverflow.com/a/5143867).

maedoc:
Otherwise by commenting out those core count usages I was able to build from source.

psalz:
Thanks for bringing this to our attention. While we don't "officially" support macOS (we neither have CI set up nor am I aware of anyone in our team having a Mac), I think we can certainly work around minor issues such as this. That being said, I'm not sure how important it would be to have that warning on macOS in the first place, as I would expect Macs to mostly be used during development (same as Windows, really). So we could just conditionally disable it. Unless you want to file a PR, of course :-).

I'm also interested in those test failures. What SYCL implementation and version are you building against?

maedoc:
> Macs to mostly be used during development 

That's my case, and I will likely switch to Docker in order to test multi "node" (i.e. containers) scenarios, where this issue doesn't arise.

> What SYCL implementation and version are you building against?

hipSYCL, develop branch at 209a7aac3ec, with the OMP backend.  But it could have as much to do with the hacks to get celerity to build, right? 
```diff
index 7d1d245..8aa311d 100644
--- a/src/platform_specific/affinity.unix.cc
+++ b/src/platform_specific/affinity.unix.cc
@@ -5,17 +5,20 @@
 #include <pthread.h>
 #include <sched.h>

-#include "affinity.h"
+// #include "affinity.h"

 namespace celerity {
 namespace detail {

        uint32_t affinity_cores_available() {
+               return 4;
+               /*
                cpu_set_t available_cores;
                const auto affinity_error = pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &available_cores);
                assert(affinity_error == 0 && "Error retrieving affinity mask.");
                return CPU_COUNT(&available_cores);
+               */
        }

 } // namespace detail
-} // namespace celerity
\ No newline at end of file
+} // namespace celerity
diff --git a/test/runtime_tests.cc b/test/runtime_tests.cc
index 4051b76..c914f99 100644
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -15,7 +15,7 @@

 #include <celerity.h>

-#include "affinity.h"
+// #include "affinity.h"
 #include "ranges.h"
 #include "region_map.h"

[...]

PR #99, opened 2022-02-28
Disable core count warning on macOS

facuMH:
Due to some incompatibilities mentioned in #85 we decided to exclude this check and test in OSX systems, since for now they are not supported.

This PR depends on #97

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,7 @@ set(SOURCES
 
 if(WIN32)
   set(SOURCES ${SOURCES} src/platform_specific/affinity.win.cc)
- elseif(UNIX)
+ elseif(UNIX AND NOT APPLE)
   set(SOURCES ${SOURCES} src/platform_specific/affinity.unix.cc)
 endif()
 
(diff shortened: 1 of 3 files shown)


BlackMark29A:
Looks good so far, but I assume this still won't compile on mac, will it? In cmake we have:

```cmake
if(WIN32)
  set(SOURCES ${SOURCES} src/platform_specific/affinity.win.cc)
 elseif(UNIX)
  set(SOURCES ${SOURCES} src/platform_specific/affinity.unix.cc)
endif()
```

and `UNIX` is also true on mac, but this translation unit then uses the `*_np` functions that don't work on mac, so this won't compile.

BlackMark29A (approved):
Looks good to me.

However, it would of course be a good idea to try to actually compile this on OSX, if at all feasible.

psalz (approved):
LGTM. @maedoc, @almightyvats could you give this a try on OSX?

maedoc:
> give this a try on OSX?

It configures & builds cleanly here (hipSYCL) and tests pass, except the two mentioned before.

PR #100, opened 2022-02-28
Replace std::list with small vector for dependency tracking

PeterTh:
This replaces the use of `std::list` for dependency tracking with `gch::small_vector`, and introduces some microbenchmarks for it.

The only other active use of `std::list` in celerity is in the buffer transfer manager, but all operations on those also seem to involve the network eventually, so micro-optimizing that doesn't appear to make much sense.

Microbenchmark results in graphical form are [available here](https://docs.google.com/spreadsheets/d/1n6uTvBkPvZROPfPZ7HZSH6yOL9E8yLszsQk8X79dg6I).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,7 @@
 [submodule "vendor/Catch2"]
 	path = vendor/Catch2
 	url = https://github.com/catchorg/Catch2
+[submodule "vendor/small_vector"]
+	path = vendor/small_vector
+	url = https://github.com/gharveymn/small_vector.git
+	branch = main
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,8 @@ target_include_directories(celerity_runtime PUBLIC
   $<INSTALL_INTERFACE:include/celerity>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/vendor>
   $<INSTALL_INTERFACE:include/celerity/vendor>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/vendor/small_vector/source/include>
+  $<INSTALL_INTERFACE:include/celerity/vendor/small_vector/include>
 )
 
 target_link_libraries(celerity_runtime PUBLIC
@@ -351,6 +353,10 @@ install(
 install(
   FILES ${PROJECT_SOURCE_DIR}/vendor/ctpl_stl.h
   DESTINATION include/celerity/vendor
+)
+install(
+  FILES ${PROJECT_SOURCE_DIR}/vendor/small_vector/source/include/gch/small_vector.hpp
+  DESTINATION include/celerity/vendor/small_vector/include/gch
 )
 install(
   TARGETS celerity_runtime
--- a/test/benchmarks.cc
+++ b/test/benchmarks.cc
@@ -1,0 +1,59 @@
+#include <catch2/catch.hpp>
+
+#include <intrusive_graph.h>
+
+using namespace celerity::detail;
+
+struct bench_graph_node : intrusive_graph_node<bench_graph_node> {};
+
+
+template <int N>
+void intrusive_graph_benchmark() {
+	BENCHMARK("creating nodes") {
+		bench_graph_node nodes[N];
+		return nodes[N - 1].get_pseudo_critical_path_length(); // trick the compiler
+	};
+
+	BENCHMARK("creating and adding dependencies") {
+		bench_graph_node n0;
+		bench_graph_node nodes[N];
+		for(int i = 0; i < N; ++i) {
+			n0.add_dependency({&nodes[i], dependency_kind::TRUE_DEP});
+		}
+		return n0.get_dependencies();
+	};
+
+	bench_graph_node n0;
+	bench_graph_node nodes[N];
+	BENCHMARK("adding and removing dependencies") {
+		for(int i = 0; i < N; ++i) {
+			n0.add_dependency({&nodes[i], dependency_kind::TRUE_DEP});
+		}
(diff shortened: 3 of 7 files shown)


PeterTh:
Note: as per to our previous team discussion, this should probably not be squashed, since it has the old benchmark results in the prior commit and the new ones with the change in the new commit.

test/benchmarks.cc
```diff
@@ -0,0 +1,44 @@
...
+
+struct bench_graph_node : intrusive_graph_node<bench_graph_node> {};
+
+TEST_CASE("benchmark intrusive graph dependency handling", "[benchmark]") {
+	// this is deliberately chosen to be relatively small, since that's what we expect to encounter in practice
+	int N = GENERATE(2, 7, 13);
+
+	BENCHMARK("creating and adding dependencies") {
+		bench_graph_node n0;
+		bench_graph_node nodes[N];
```
BlackMark29A, line 15: As discussed with @PeterTh, this only works because we have GNU extensions and should probably be a `std::vector`.

BlackMark29A (approved):
LGTM.

CMakeLists.txt
```diff
@@ -352,6 +354,10 @@ install(
   FILES ${PROJECT_SOURCE_DIR}/vendor/ctpl_stl.h
   DESTINATION include/celerity/vendor
 )
+install(
+  FILES ${PROJECT_SOURCE_DIR}/vendor/small_vector/source/include/gch/small_vector.hpp
+  DESTINATION include/celerity/vendor/small_vector/source/include/gch
```
fknorr, line 359: Since we are in complete control of the install directory layout, we should consider shortening this path by stripping the `source` subdirectory.

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "vendor/Catch2"]
 	path = vendor/Catch2
 	url = https://github.com/catchorg/Catch2
+[submodule "vendor/small_vector"]
+	path = vendor/small_vector
+	url = https://github.com/gharveymn/small_vector.git
+	branch = main
```
fknorr, line 10: I guess this was auto-generated, but what does `branch` do here exactly? The checked out revision is not stored in this file anyway. Just curious.

test/benchmarks.cc
```diff
@@ -0,0 +1,44 @@
+#include <catch2/catch.hpp>
+
+#include <intrusive_graph.h>
```
fknorr, line 3: We use quotes (`"intrusive_graph.h"`) for our own headers by convention everywhere else. We should align this.

test/benchmarks.cc
```diff
@@ -0,0 +1,44 @@
...
+	// this is deliberately chosen to be relatively small, since that's what we expect to encounter in practice
+	int N = GENERATE(2, 7, 13);
+
+	BENCHMARK("creating and adding dependencies") {
+		bench_graph_node n0;
+		bench_graph_node nodes[N];
+		for(int i = 0; i < N; ++i) {
+			n0.add_dependency({&nodes[i], dependency_kind::TRUE_DEP});
+		}
+		return n0.get_dependencies();
```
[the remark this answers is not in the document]
fknorr: Does it? From what I understand, the curly braces after `BENCHMARK` belong to a lambda expression, so they return from the `BENCHMARK`, not the `TEST_CASE`. This is probably done to stop the compiler from optimizing the previous side effects away.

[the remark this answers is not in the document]
fknorr: Oh, right. That's done correctly though in the benchmark in line 28. What's up with that?

CMakeLists.txt
PeterTh: I was thinking about this but decided against it since I thought it would be less surprising to have the same layout as in the repo. Of course this was from our perspective rather than user perspective.

.gitmodules
PeterTh: I don't know, as you assumed this is generated and I just kept it as-is.

test/benchmarks.cc
PeterTh: I actually checked in `intrusive_graph_tests.cc` which is the same type of file in the same folder with the same use case, and it uses `<>`, that's why I went with that. I can change both I guess.

test/benchmarks.cc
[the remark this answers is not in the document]
BlackMark29A: The idea behind doing it differently is that this is supposed to show the difference between including the construction of the node creation in the benchmark, and not doing that.

test/benchmarks.cc
PeterTh: Done.

CMakeLists.txt
PeterTh: Uses the same layout as if one were to install gch/small_vector now.

test/benchmarks.cc
```diff
@@ -0,0 +1,63 @@
...
+using namespace celerity::detail;
+
+struct bench_graph_node : intrusive_graph_node<bench_graph_node> {};
+
+template <int N>
+void intrusive_graph_benchmark() {
+
+	// note that bench_graph_nodes are created/destroyed *within* the BENCHMARK
+	// in the first two cases while the latter 2 cases only operate on already
+	// existing nodes -- this is intentional; both cases are relevant in practise
```
fknorr, line 14: typo: practise -> practice (noun)

fknorr (approved):

test/benchmarks.cc
PeterTh: It's ok, let's say it's BE.

psalz (approved):
LGTM!

test/benchmarks.cc
fknorr: Austrian English at most :wink:

PR #102, opened 2022-03-04
Unit test maintenance & refactoring

PeterTh:
This only touches unit tests, but does all the following:

- upgrade Catch2 to v3, and use individual headers
- semantically restructure unit tests to more translation units
- eliminate some code duplication
- improve naming & tagging of a few units test

I verified that the number of assertions executed on `master` and `unit-test-mainenance` is the same.

Full rebuild times (s):
| Type             | `master`        | This PR |
| ------------- | ------------: | ------------: |
|  Debug     | 65.992 | 51.792 |
|  Release | 183.796  | 105.901  |

Obviously, partial rebuilds of affected TUs benefit even more.


BlackMark29A (approved):
LGTM.

PR #104, opened 2022-03-09
Add all_tests unit test executable, linking all unit test object files

PeterTh:
Build system and CI only change, builds all tests as CMake "Object libraries" and links them together as required.


psalz (approved):
LGTM

PeterTh:
All of the CI runs got faster with this -- I looked into the logs and all assertions which should run seem to run.
It seems like half the reason is less startup/shutdown overhead for individual unit tests, and half is not running the benchmarks as part of the CI tests (I made a separate draft issue for that).

fknorr (approved):

PR #106, opened 2022-03-11
Do not instantiate runtime in tests that do not need it

fknorr:
Currently, "pure" tests such as the `task_graph_test` binary instantiate the runtime once at the beginning. This is unnecessary and slows down profiling of such "pure" test cases.

This PR

- Replaces the global Catch2 runtime re-instantiation hook with `test_utils::runtime_fixture` (and `test_utils::mpi_fixture` for those that need MPI, but no runtime instance)
- Triggers an assertion if the runtime is instantiated in a test that does not have a `runtime_fixture`
- Warns if a `runtime_fixture` is used on a test, but the runtime does not end up being instantiated
- Removes complexity around the `unit_test_suite` – it does not need multiple translation units anymore

Some tests might be refactored to avoid instantiating the runtime or initializing MPI. For example, `sycl_tests` currently require MPI transitively through `device_queue` and `config`. Also, we should move tests that do not involve the runtime out of `runtime_tests.cc`. I consider both these points to be out of scope for this PR, though.


psalz (approved):
Just yesterday I was talking to @PeterTh that we probably no longer need that `unit_test_suite` workaround. This is great! LGTM!

PR #109, opened 2022-03-24
Add custom CSV and Markdown reporters for benchmark results

psalz:
This adds a custom CSV reporter for producing machine-readable benchmarking results. You can see an example output in the updated results for gpuc1 (I noticed some of the numbers diverge quite wildly from the previous results, not sure how reliable those are - cc @PeterTh).

To use the reporter, simply call the benchmarks with `--reporter celerity-benchmark-csv`.


test/benchmark_csv_reporter.cc
```diff
@@ -0,0 +1,87 @@
...
+		active_sections.clear();
+	}
+
+  private:
+	// It is difficult to distinguish test case invocations for different generator values, since
+	// as of Catch2 v.3.0.0-preview4 there is unfortunately no real metadata available (for example,
+	// it is not possible to get the current value returned for each of the generators).
+	// The only piece of information we have is the "part number", which we emit to distinguish
+	// different generator values. Note however that sections also increment the part number.
+	uint64_t active_part;
```
fknorr, line 83: The default text reporter also doesn't show generator values, does it? Including the part number is not a bad idea, but it's more an argument for not using generators in benchmarks.

test/benchmark_csv_reporter.cc
```diff
@@ -0,0 +1,87 @@
...
+ * active in the current invocation will be printed as well (not just those directly
+ * surrounding a BENCHMARK).
+ */
+class benchmark_csv_reporter : public Catch::StreamingReporterBase {
+  public:
+	using StreamingReporterBase::StreamingReporterBase;
+
+	static std::string getDescription() { return "Reporter for benchmarks in CSV format"; }
+
+	void testRunStarting(Catch::TestRunInfo const& testRunInfo) override {
```
fknorr, line 35: We are west const everywhere else, so we should try to be consistent here (or change our policy, which I'm not opposed to either).

fknorr (commented):
Since the output has is for machine use anyway, IMO we should just export individual samples and leave all the statistics to the postprocessing script (e.g. scipy + matplotlib). That would be strictly more general.

test/benchmark_csv_reporter.cc
psalz: I'm never sure how to deal with vendor interfaces that use different conventions. In this case I've opted to stick with Catch2's style for the function signatures (you'll notice the parameters are camelCased as well). But we can also mix it with our style. I don't have a strong preference, either feels weird to me.

Regarding our policy: The SYCL spec is west const, which sets a rather strong precedence I'd say.

test/benchmark_csv_reporter.cc
psalz: I agree that generators should better be avoided, but I also don't like emitting the same line twice in case somebody does use them anyway. I could try detecting generators being used and print a warning.

psalz:
I've changed the CSV reporter to also include the raw samples. I've also introduced a new Markdown reporter that can be used to produce human-readable results alongside the CSV. Simply run with `./test/benchmarks --reporter celerity-benchmark-csv::results.csv --reporter celerity-benchmark-md::results.md` to generate both simultaneously.

test/benchmark_csv_reporter.cc
psalz: I've implemented the warning.

fknorr:
Thanks! Most of this looks great, however for the Markdown output, we technically need to escape [all the special characters](https://www.markdownguide.org/basic-syntax/#escaping-characters) so test / benchmark names aren't interpreted as formatting. For example, the CDAG benchmarks have an underscore as in `wave_sim topology`.

This is less about nitpicking and more towards the question whether we actually need proper Markdown output, or if a `.txt` file with an ASCII table (which essentially looks the same source-wise) would be enough. Viewing the Markdown file "raw" makes it much easier to parse for my eyes, since the numbers are properly aligned and spacing is not as wasteful (on GitHub's renderer).

PeterTh:
It looks like `adding and removing dependencies` and `checking dependencies` got much slower compared to the currently-checked-in results for the low-dep-count tests. It's still very fast, but don't really have an explanation for that.

PeterTh (approved):

psalz:
> This is less about nitpicking and more towards the question whether we actually need proper Markdown output, or if a `.txt` file with an ASCII table (which essentially looks the same source-wise) would be enough. Viewing the Markdown file "raw" makes it much easier to parse for my eyes, since the numbers are properly aligned and spacing is not as wasteful (on GitHub's renderer).

As discussed during yesterday's meeting, there is value in GitHub's Markdown diff, so we'll keep it. I've added rudimentary escaping.

fknorr (approved):
The markdown report in `ci/perf/gpuc1_bench.md` has not been updated since escaping was added. Otherwise LGTM!

psalz:
> The markdown report in `ci/perf/gpuc1_bench.md` has not been updated since escaping was added. Otherwise LGTM!

I've manually added the escaping to the relevant benchmark names; rebased onto master.

PR #110, opened 2022-03-28
Do not create empty entries in command_buffer_reads

fknorr:
A read-only `operator[]` access to `graph_generator::command_buffer_reads` caused unnecessary insertion into the map, causing a weird test failure down the line in #86 . This replaces the access with an `if (it=find(), it != end())`.

The accompanying test can now use stricter bounds on the map size. Somewhat unrelated, it was using `SECTION`s where `GENERATOR`s are much more straightforward, so this was drive-by refactored.

### DIFF
--- a/test/graph_compaction_tests.cc
+++ b/test/graph_compaction_tests.cc
@@ -1,7 +1,5 @@
 #include <array>
 #include <optional>
-#include <set>
-#include <unordered_set>
 
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/generators/catch_generators.hpp>
@@ -28,12 +26,6 @@ namespace detail {
 		static size_t get_num_regions(const region_map<T>& map) {
 			return map.region_values.size();
 		}
-		template <typename T>
-		static void print_regions(const region_map<T>& map) {
-			for(auto& reg : map.region_values) {
-				fmt::print("{} -> {}\n", reg.first, reg.second);
-			}
-		}
 	};
 
 	struct graph_generator_testspy {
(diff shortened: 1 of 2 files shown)


test/graph_compaction_tests.cc
```diff
@@ -65,78 +66,51 @@ namespace detail {
...
+		CHECK(buf_a_region_map_size <= NUM_NODES * 2 * horizon_step_size);
+		const auto buf_a_last_writer_map_size = graph_generator_testspy::get_buffer_last_writer_num_regions(ctx.get_graph_generator(), buf_a.get_id());
+		CHECK(buf_a_last_writer_map_size <= NUM_NODES * 2 * horizon_step_size);
+		for(node_id n = 0; n < NUM_NODES; ++n) {
+			CHECK(inspector.get_commands(std::nullopt, n, command_type::HORIZON).size() <= NUM_TIMESTEPS / horizon_step_size + 1);
+			CHECK(inspector.get_commands(std::nullopt, n, command_type::HORIZON).size() >= NUM_TIMESTEPS / horizon_step_size - 1);
 		}
 
+		// also check that unused commands are deleted
+		CHECK(ctx.get_command_graph().command_count() <= NUM_NODES * 13 * horizon_step_size);
```
psalz, line 110: Do we now know where that 13 (and 6 below) is coming from? If so consider adding a comment.

psalz (approved):
LGTM!

PeterTh (approved):

test/graph_compaction_tests.cc
fknorr: I have now reverse engineered and documented those values.

fknorr: I have now reverse engineered and documented these bounds.

fknorr:
I have added another commit that removes unused includes and commented-out debug facilities from the same test.

test/graph_compaction_tests.cc
psalz: Wow, looks like this took a lot of work!

PeterTh (approved):
I like the test changes. I'm not certain about removing the debug utilities -- they are just in the test code and I assume they were useful while trying to understand the test?

fknorr:
Not for this fix, no, I had some ad-hoc functionality for printing buffer reads (instead of writes) which also didn't make the PR. `print_regions` is one of the few remaining implicit iostreams uses (through fmt/ostream), so I have to admit my ulterior motive here :)

PR #111, opened 2022-03-30
Improve scheduler performance by reducing lock contention

psalz:
This PR is based on #109 so that one should be merged first.

This improves the performance of the scheduler in two ways:
  - Address a long standing issue (we've had a TODO note about this for years...) where the scheduler thread would hold the lock on the `events` queue until a task was fully built and flushed, effectively causing it to operate in lockstep with the main thread. This has been fixed by double-buffering the event queue, with the scheduler thread only locking the mutex to swap the two.
  - Reduce contention around the `task_manager`'s mutex by passing a task pointer to the scheduler directly, instead of it repeatedly calling `task_manager::get_task` (while the main thread is trying to acquire the lock for producing new tasks).

Additionally I've reduced the number of tasks generated in the "soup" topology benchmark so that the timings are in about the same order of magnitude as for the other topologies (= fits better onto a chart).

Here is a comparison between the original "lockstep" scheduler and the double-buffered "decoupled" one, but still with frequent calls to `get_task`. There are two bars for each benchmark in each group (same colors), with the right bar representing the decoupled scheduler. The two "ST *" groups refer to the single-threaded baseline implementation that generates commands in the main thread, circumventing the scheduler altogether (numbers in these groups should not be affected by the changes).

image

We can see that this is an easy win overall, with numbers improving across the board, in particular for the "soup" and "wave_sim" topologies. Additionally jitter is greatly reduced.

Unfortunately numbers are not *quite* as clear for the second change. Here we now compare the decoupled scheduler against the task pointer being passed around (the left bar here is the right bar above):

image

There are again considerable improvements for "soup" and "wave_sim", which now also outperform the single-threaded variant in most cases. Jitter is again reduced across the board. However with 4 nodes, there are some numbers that are worse, in particular for the "Immediate" group. While I was able to replicate this several times, I've also encountered an instance where performance was better everywhere, with the "Immediate" group outperforming "ST Immediate" for most topologies. I therefore attribute this "worsening" to overall jitter in our benchmarking setup, and given the significance of the "soup" and "wave_sim" results, I would still consider this an improvement overall.


include/scheduler.h
```diff
@@ -51,13 +51,16 @@ namespace detail {
 		graph_generator& ggen;
 		graph_serializer& gsrlzr;
 
-		std::queue<scheduler_event> events;
+		std::queue<scheduler_event> available_events;
+		std::queue<scheduler_event> in_flight_events;
```
fknorr, line 55: `in_flight_events` is only used inside `schedule()`, so it could be a local variable. This would make it more obvious that it is not synchronized by `events_mutex`.

src/scheduler.cc
```diff
@@ -55,6 +65,7 @@ namespace detail {
 	void scheduler::shutdown() {
 		abstract_scheduler::shutdown();
 		if(worker_thread.joinable()) { worker_thread.join(); }
+		assert(available_events.empty() && in_flight_count == 0);
```
fknorr, line 68: This is missing a lock on `events_mutex` (can potentially race with `notify()`). It should be enough to take it in a `#ifndef NDEBUG' though.

include/graph_generator.h
```diff
@@ -77,18 +77,16 @@ namespace detail {
...
 		 * @param cdag The command graph this generator should operate on.
 		 */
-		graph_generator(size_t num_nodes, task_manager& tm, reduction_manager& rm, command_graph& cdag);
+		graph_generator(size_t num_nodes, reduction_manager& rm, command_graph& cdag);
 
 		void add_buffer(buffer_id bid, const cl::sycl::range<3>& range);
 
 		// Build the commands for a single task
-		void build_task(task_id tid, const std::vector<graph_transformer*>& transformers);
+		void build_task(const task& tsk, const std::vector<graph_transformer*>& transformers);
```
fknorr, line 87: I like the use of `task&` instead of `task*` for non-null parameters here. IMO this should be policy for all non-null `task` and `abstract_command` parameters, which are currently pointers everywhere else for some reason.

include/scheduler.h
```diff
@@ -51,13 +51,16 @@ namespace detail {
...
+		std::atomic_size_t in_flight_count = 0;
```
fknorr, line 56: This is only really needed by `scheduler::is_idle()`, which will be removed by #86 (only use case was the ad-hoc `slow_full_sync`). Depending on merge order, we can do without this variable entirely.

src/scheduler.cc
```diff
@@ -14,35 +14,45 @@ namespace detail {
...
-				return;
+			{
+				std::unique_lock lk(events_mutex);
+				events_cv.wait(lk, [this] { return !available_events.empty(); });
+				std::swap(available_events, in_flight_events);
+				in_flight_count = in_flight_events.size();
+			}
+
+			while(!in_flight_events.empty()) {
+				const auto event = std::move(in_flight_events.front());
```
fknorr, line 31: Clang-Tidy: `std::move` of the expression of the trivially-copyable type `__gnu_cxx::__alloc_traits<std::allocator<celerity::detail::scheduler_event>, celerity::detail::scheduler_event>::value_type` (aka `celerity::detail::scheduler_event`) has no effect; remove `std::move()`

src/scheduler.cc
```diff
@@ -14,35 +14,45 @@ namespace detail {
...
+				in_flight_events.pop();
+
+				if(event.type == scheduler_event_type::SHUTDOWN) {
+					assert(in_flight_events.empty());
+					in_flight_count -= 1;
+					return;
+				}
+
+				if(event.type == scheduler_event_type::TASK_AVAILABLE) {
```
fknorr, line 40: This could be an assertion instead of an if, which would make it more obvious that no event is skipped – and by extension `in_flight_count` never misses an update.

fknorr (commented):
Great improvement, thanks! Even if there is not always a benefit to the single-threaded setup, this is an important requirement for any parallel CDAG generation we might want to do in the future.

Could you add another bar chart for the TDAG times as a baseline to see what the best-case improvement of the separate scheduler thread would be?

psalz:
> Could you add another bar chart for the TDAG times as a baseline to see what the best-case improvement of the separate scheduler thread would be?

Yes! I've plotted the single-threaded TDAG+CDAG times against the dedicated scheduler thread. If we had perfect overlapping, the scheduler times would be equal to the single-threaded CDAG times. For 4 nodes we are not that far off!

image

include/scheduler.h
psalz: Good idea, done!

src/scheduler.cc
psalz: Right, although this is moot now, since the refactoring into `abstract_scheduler` doesn't allow me to access these members anyway. I've removed the assertion.

include/graph_generator.h
psalz: Agreed!

src/scheduler.cc
psalz: I got the same diagnostic and deliberately kept this so that if `scheduler_event` would for some reason no longer be trivially-copyable in the future, we're already set. As it says, it's a no-op right now so it doesn't really do any harm either, IMO.

src/scheduler.cc
psalz: Right, changed.

src/scheduler.cc
fknorr: Fine with me, but we could still document the intent with a `// NOLINT(performance-move-const-arg)`.

fknorr:
Thanks! LGTM up to the `in_flight_count` remark. Some notes for future investigation into scheduler-thread performance:

It's remarkable how far results differ between topologies. It seems intuitive that the benefit should be maximal when TDAG and CDAG generation are most similar in execution time (`soup` in both graphs, `wave_sim` in the single-node case). I assume that the positive result for `wave_sim` in the 4-node example is simply due to jitter, because in no other scenario can the TDAG generation latency be hidden completely.

To me it seems like the 1-node soup should be the perfect condition for parallelization since TDAG and CDAG can be almost fully overlapped and the total runtime is long enough to avoid any constant overhead that might be visible in faster runs (e.g. `chain`). Yet we only save about 23% instead of the theoretical 45%. My guess is that since each individual task is very quick to compute, the scheduler has to block on its queue after every task, so we add an OS scheduler round-trip to each iteration.

Also it is notable that for 1-node soup, TDAG generation is more expensive than CDAG. Both generate mostly identical graphs, but the CDAG is also serialized (and contains the infamous `debug_label`), so I would expect it to be the other way round. Possible causes that come to mind:
- Allocations for command group type erasure and range mappers
- `task_mutex`, locked and unlocked for every task – but it is not contended here, so locking should never need a syscall

fknorr (approved):

psalz:
I've rebased onto master, which now includes the task ring buffer. Due to this the second optimization (passing around a task-pointer) is moot, as `task_manager::get_task` is now lock-free. Nevertheless, I think it's still a good thing to do.

With the ring buffer times are lower in general and the results after decoupling the scheduler are a lot less dramatic. Improvements are most noticeable for tasks where computing the command graph isn't cheap, e.g. wave_sim.

image

image

Note besides decreasing the number of tasks for the "soup" benchmark (as described above), I've now also changed the benchmarks to use the task-manager callback to notify the scheduler of newly created tasks. This has the effect that now horizons will also be built by the benchmark (previously they were silently ignored).

fknorr (approved):
LGTM! Some day we should do a pass over both task and command graph code replacing DAG node pointers with references where they are guaranteed to be non-null.

include/scheduler.h
```diff
@@ -1,23 +1,23 @@
 #pragma once
 
+#include <atomic>
```
PeterTh, line 3: I assume this is needed now because it was transitive over "types.h" before?

include/task_manager.h
```diff
@@ -74,6 +75,7 @@ namespace detail {
 				prepass_handler cgh(tid, std::make_unique<command_group_storage<CGF>>(cgf), num_collective_nodes);
 				cgf(cgh);
 				task& task_ref = register_task_internal(std::move(reservation), std::move(cgh).into_task());
+				tsk_ptr = &task_ref;
 
 				compute_dependencies(tid);
```
PeterTh, line 80: Maybe this should also directly take tasks? The other call-site could do this as well (if `reduce_execution_front` return a task rather than an id, which it probably should).

PeterTh (approved):
LGTM, other than the comment on going a bit farther with the task_id -> task refactoring (could also be a separate PR).

include/scheduler.h
psalz: Actually this was only needed before the rebase onto epochs - removed.

include/task_manager.h
psalz: Right, I've changed the task manager to also pass around references instead of ids in most places.

src/task_manager.cc
```diff
@@ -237,12 +236,12 @@ namespace detail {
...
 			current_horizon_critical_path_length = max_pseudo_critical_path_length;
 			const auto previous_horizon = current_horizon;
-			current_horizon = reduce_execution_front(std::move(reserve), task::make_horizon_task(tid));
+			current_horizon = reduce_execution_front(std::move(reserve), task::make_horizon_task(tid)).get_id();
 			if(previous_horizon) { set_epoch_for_new_tasks(*previous_horizon); }
 		}
 
 		// it's important that we don't hold the lock while doing this
-		invoke_callbacks(tid);
+		invoke_callbacks(get_task(tid));
```
PeterTh, line 244: Could get rid of this `get_task` by storing the task returned by `reduce_execution_front`.

src/task_manager.cc
```diff
@@ -253,15 +252,15 @@ namespace detail {
...
 			compute_dependencies(new_epoch);
-			set_epoch_for_new_tasks(new_epoch);
+			set_epoch_for_new_tasks(new_epoch.get_id());
 			current_horizon = std::nullopt; // this horizon is now behind the epoch_for_new_tasks, so it will never become an epoch itself
 			current_horizon_critical_path_length = max_pseudo_critical_path_length; // the explicit epoch resets the need to create horizons
 		}
 
 		// it's important that we don't hold the lock while doing this
-		invoke_callbacks(tid);
+		invoke_callbacks(get_task(tid));
```
PeterTh, line 263: Same here.

PeterTh (approved):

src/task_manager.cc
psalz: Done!

src/task_manager.cc
psalz: Done!

PR #113, opened 2022-04-05
Add overload for runtime::init/distr_queue ctor that accepts a device selector

almightyvats:
This feature adds the option to pass a device selector to the `distr_queue`, according to which devices get selected. Previously, a `sycl::device` could be passed (now deprecated). The device or the selector is then passed to the `device_queue.h/pick_device` function, which accepts a variant for either automatic device selection or the device selector or the `sycl::device` itself. 

The `pick_device` function has in total 4 branches in which a device can be selected. The first one is choosing the provided `sycl::device`, the second one is, choosing the device with the `CELERITY_DEVICES` environment variable. If in case none of those two options are taken, we then have another branch that either chooses the device automatically or uses the selector to choose the device.

The logic for automatic device selection:
1. `try_find_device_per_node` function takes in the type of device which should be selected and checks if any platform has enough devices of the same type to assign a unique device to each local node. 
The automatic selection first checks for all the unique devices to be GPUs, if not, it checks for any type of enough devices. 
2.  If there's no platform that can provide enough unique devices, the automatic selection tries to find at least one device, which is preferably a GPU, if not then any type of device with `try_find_one_device`.

The logic for device selection with the selector:
The same two functions as above are called, however, the type of device doesn't matter here. 
1. For `try_find_device_per_node` all the devices are collected and are sorted using the device selector. If enough devices of the same type belong to one platform, they are selected.
2. Similarly, for `try_find_one_device`, all the devices are collected and sorted and the device with the highest score is selected. 
3. A device with a score of `-1` is excluded from the selection. 

In case, no device gets selected according to the selector, the selection does not fall back to any random device.

### DIFF
--- a/include/celerity.h
+++ b/include/celerity.h
@@ -1,6 +1,7 @@
 #ifndef RUNTIME_INCLUDE_ENTRY_CELERITY
 #define RUNTIME_INCLUDE_ENTRY_CELERITY
 
+#include "device_queue.h"
 #include "runtime.h"
 
 #include "accessor.h"
@@ -15,14 +16,24 @@ namespace runtime {
 	/**
 	 * @brief Initializes the Celerity runtime.
 	 */
-	inline void init(int* argc, char** argv[]) { detail::runtime::init(argc, argv, nullptr); }
+	inline void init(int* argc, char** argv[]) { detail::runtime::init(argc, argv, detail::auto_select_device{}); }
 
 	/**
 	 * @brief Initializes the Celerity runtime and instructs it to use a particular device.
 	 *
 	 * @param device The device to be used on the current node. This can vary between nodes.
 	 */
-	inline void init(int* argc, char** argv[], cl::sycl::device& device) { detail::runtime::init(argc, argv, &device); }
+	[[deprecated("Use the overload with device selector instead, this will be removed in future release")]] inline void init(
+	    int* argc, char** argv[], sycl::device& device) {
+		detail::runtime::init(argc, argv, device);
+	}
+
+	/**
+	 * @brief Initializes the Celerity runtime and instructs it to use a particular device.
+	 *
+	 * @param device_selector The device selector to be used on the current node. This can vary between nodes.
+	 */
+	inline void init(int* argc, char** argv[], const detail::device_selector& device_selector) { detail::runtime::init(argc, argv, device_selector); }
 } // namespace runtime
 } // namespace celerity
 
--- a/include/device_queue.h
+++ b/include/device_queue.h
@@ -59,9 +59,79 @@ namespace detail {
 		std::unique_ptr<cl::sycl::queue> sycl_queue;
 		bool device_profiling_enabled = false;
 
-		cl::sycl::device pick_device(const config& cfg, cl::sycl::device* user_device) const;
 		void handle_async_exceptions(cl::sycl::exception_list el) const;
 	};
+
+	template <typename DeviceT, typename PlatformT>
+	DeviceT pick_device(const config& cfg, DeviceT* user_device, const std::vector<PlatformT>& platforms) {
+		DeviceT device;
+		std::string how_selected = "automatically selected";
+		if(user_device != nullptr) {
+			device = *user_device;
+			how_selected = "specified by user";
+		} else {
+			const auto device_cfg = cfg.get_device_config();
+			if(device_cfg != std::nullopt) {
+				how_selected = fmt::format("set by CELERITY_DEVICES: platform {}, device {}", device_cfg->platform_id, device_cfg->device_id);
+				CELERITY_DEBUG("{} platforms available", platforms.size());
+				if(device_cfg->platform_id >= platforms.size()) {
+					throw std::runtime_error(fmt::format("Invalid platform id {}: Only {} platforms available", device_cfg->platform_id, platforms.size()));
+				}
+				const auto devices = platforms[device_cfg->platform_id].get_devices();
+				if(device_cfg->device_id >= devices.size()) {
+					throw std::runtime_error(fmt::format(
+					    "Invalid device id {}: Only {} devices available on platform {}", device_cfg->device_id, devices.size(), device_cfg->platform_id));
+				}
+				device = devices[device_cfg->device_id];
+			} else {
+				const auto host_cfg = cfg.get_host_config();
+
+				const auto try_find_device_per_node = [&host_cfg, &device, &how_selected, &platforms](cl::sycl::info::device_type type) {
+					// Try to find a platform that can provide a unique device for each node.
+					for(size_t i = 0; i < platforms.size(); ++i) {
+						auto&& platform = platforms[i];
+						const auto devices = platform.get_devices(type);
+						if(devices.size() >= host_cfg.node_count) {
+							how_selected = fmt::format("automatically selected platform {}, device {}", i, host_cfg.local_rank);
+							device = devices[host_cfg.local_rank];
+							return true;
+						}
+					}
+					return false;
+				};
+
+				const auto try_find_one_device = [&device, &platforms](cl::sycl::info::device_type type) {
+					for(auto& p : platforms) {
+						for(auto& d : p.get_devices(type)) {
+							device = d;
+							return true;
+						}
+					}
+					return false;
+				};
+
+				// Try to find a unique GPU per node.
+				if(!try_find_device_per_node(cl::sycl::info::device_type::gpu)) {
+					// Try to find a unique device (of any type) per node.
+					if(try_find_device_per_node(cl::sycl::info::device_type::all)) {
+						CELERITY_WARN("No suitable platform found that can provide {} GPU devices, and CELERITY_DEVICES not set", host_cfg.node_count);
+					} else {
+						CELERITY_WARN("No suitable platform found that can provide {} devices, and CELERITY_DEVICES not set", host_cfg.node_count);
+						// Just use the first available device. Prefer GPUs, but settle for anything.
+						if(!try_find_one_device(cl::sycl::info::device_type::gpu) && !try_find_one_device(cl::sycl::info::device_type::all)) {
+							throw std::runtime_error("Automatic device selection failed: No device available");
+						}
+					}
+				}
+			}
+		}
+
+		const auto platform_name = device.get_platform().template get_info<cl::sycl::info::platform::name>();
+		const auto device_name = device.template get_info<cl::sycl::info::device::name>();
+		CELERITY_INFO("Using platform '{}', device '{}' ({})", platform_name, device_name, how_selected);
+
+		return device;
+	}
 
 } // namespace detail
 } // namespace celerity
--- a/include/device_queue.h
+++ b/include/device_queue.h
@@ -3,12 +3,18 @@
 #include <memory>
 
 #include <CL/sycl.hpp>
+#include <type_traits>
+#include <variant>
 
 #include "config.h"
 #include "workaround.h"
 
 namespace celerity {
 namespace detail {
+
+	struct auto_select_device {};
+	using device_selector = std::function<int(const sycl::device&)>;
+	using device_or_selector = std::variant<auto_select_device, sycl::device, device_selector>;
 
 	class task;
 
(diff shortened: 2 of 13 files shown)


test/sycl_tests.cc
```diff
@@ -171,4 +171,46 @@ TEST_CASE_METHOD(test_utils::device_queue_fixture, "SYCL has working simple scal
 
 #endif
 
+TEST_CASE("SYCL implements by-value equality-comparison of device information", "[sycl][device-selection][!mayfail]") {
+	constexpr static auto get_devices = [] {
```
BlackMark29A, line 175: Is there a reason why this is `static`? (Same with the other lambda below)

BlackMark29A (changes requested):
Looks quite promising, however I have some remarks and questions.

include/device_queue.h
```diff
@@ -59,9 +64,149 @@ namespace detail {
 		std::unique_ptr<cl::sycl::queue> sycl_queue;
 		bool device_profiling_enabled = false;
 
-		cl::sycl::device pick_device(const config& cfg, cl::sycl::device* user_device) const;
 		void handle_async_exceptions(cl::sycl::exception_list el) const;
 	};
 
+	template <typename DevicePtrOrSelector, typename PlatformT>
+	auto pick_device(const config& cfg, const DevicePtrOrSelector& user_device_or_selector, const std::vector<PlatformT>& platforms) {
+		using DeviceT = typename std::result_of<decltype (&PlatformT::get_devices)(PlatformT, cl::sycl::info::device_type)>::type::value_type;
```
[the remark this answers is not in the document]
almightyvats: Changed it to a simple `decltype`, `using DeviceT = typename decltype(std::declval<PlatformT&>().get_devices())::value_type`.

include/device_queue.h
```diff
@@ -59,9 +64,149 @@ namespace detail {
...
+
+		constexpr bool user_device_provided = std::is_same_v<DevicePtrOrSelector, DeviceT>;
+		constexpr bool device_selector_provided = std::is_invocable_r_v<int, DevicePtrOrSelector, DeviceT>;
+		constexpr bool auto_select = std::is_same_v<auto_select_device, DevicePtrOrSelector>;
+		static_assert(user_device_provided ^ device_selector_provided ^ auto_select);
```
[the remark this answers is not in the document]
almightyvats: How about "pick_device requires either a device, a selector, or the auto_select_device tag"?

test/sycl_tests.cc
almightyvats: Delegated to @fknorr.

include/device_queue.h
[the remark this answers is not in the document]
BlackMark29A: Sure, this message is fine as well. I'm not too concerned with the exact wording, I just think it's a good idea to have some explicit message.

BlackMark29A (approved):
Looks good to me, I just found one minor detail which should be changed to be in-line with previous changes:

include/celerity.h
```diff
@@ -15,14 +16,26 @@ namespace runtime {
...
+	    int* argc, char** argv[], cl::sycl::device& device) {
+		detail::runtime::init(argc, argv, device);
+	}
+
+	/**
+	 * @brief Initializes the Celerity runtime and instructs it to use a particular device.
+	 *
+	 * @param device_selector The device selector to be used on the current node. This can vary between nodes.
+	 */
+	inline void init(int* argc, char** argv[], std::function<int(const cl::sycl::device&)> device_selector) {
```
psalz, line 36: Receive `device_selector` as `const &`; consider introducing an alias for it (and then also use it in the variant).

test/device_selection_tests.cc
```diff
@@ -0,0 +1,641 @@
...
+	mock_platform tp(68);
+	mock_device td(42, tp);
+	tp.set_devices({td});
+
+	auto device = pick_device(cfg, td, std::vector<mock_platform>{tp});
+	CHECK(device == td);
+}
+
+TEST_CASE_METHOD(celerity::test_utils::mpi_fixture,
+    "pick_device automatically selects a gpu device if available and otherwise falls back to the first device available", "[device-selection][gen]") {
```
psalz, line 86: What is `[gen]`?

test/device_selection_tests.cc
```diff
@@ -0,0 +1,641 @@
...
+	}
+
+	SECTION("pick_device couldn't find any device") {
+		CHECK_THROWS_WITH(
+		    pick_device(cfg, celerity::detail::auto_select_device{}, std::vector<mock_platform>{}), "Automatic device selection failed: No device available");
+	}
+}
+
+
+#if !defined(WORKAROUND_COMPUTECPP)
```
psalz, line 439: Please add a comment explaining why this is disabled for CCPP!

test/device_selection_tests.cc
```diff
@@ -0,0 +1,641 @@
...
+
+		auto device = pick_device(cfg, device_selector, std::vector<mock_platform>{tp_0, tp_1, tp_2});
+		std::vector<mock_device> devices = tp_2.get_devices();
+		INFO("Platform id" << device.get_platform().get_id() << " device id " << device.get_id());
+		CHECK_THAT(lc.get_log(), Catch::Matchers::ContainsSubstring("device selector specified: platform"));
+		CHECK(device == devices[3]);
+	}
+}
+
+TEST_CASE_METHOD(celerity::test_utils::mpi_fixture, "pick_device fails on returning a negative number for a device with device selector",
```
psalz, line 607: The test name is pretty unclear IMO, I'd suggest `"pick_device does not consider devices with a negative selector score"` or something.

include/device_queue.h
```diff
@@ -59,9 +64,166 @@ namespace detail {
...
+				}
+			}
+		}
+
+		std::sort(devices_with_platform_idx.begin(), devices_with_platform_idx.end(),
+		    [selector](const auto& a, const auto& b) { return selector(std::get<0>(a)) > selector(std::get<0>(b)); });
+		bool same_platform = true;
+		if(devices_with_platform_idx.size() >= host_cfg.node_count) {
+			auto [device_from_platform, idx] = devices_with_platform_idx[0];
+			auto platform_name = device_from_platform.get_platform().template get_info<cl::sycl::info::platform::name>();
```
psalz, line 90: Could be made `const`. Comparing based on name seems unnecessarily brittle; `sycl::platform` implements `operator==`.

include/device_queue.h
```diff
@@ -59,9 +64,166 @@ namespace detail {
...
+
+			for(size_t i = 1; i < host_cfg.node_count; ++i) {
+				auto [device_from_platform, idx] = devices_with_platform_idx[i];
+				if(device_from_platform.get_platform().template get_info<cl::sycl::info::platform::name>() != platform_name) { same_platform = false; }
```
psalz, line 94: So really the behavior here is that a selection is only made if the first N devices, according to their selector score, are from the same platform. That is fine in principle I guess, however should be tested explicitly. Also, devices that have the same score but belong to different platforms could cause problems here (while being difficult to test, depending on stability of `std::sort`).

psalz (changes requested):
Thanks, I've added some notes!

test/device_selection_tests.cc
```diff
@@ -0,0 +1,641 @@
...
+		size_t local_num_cpus = 1;
+
+		celerity::detail::host_config h_cfg{node_count, local_rank, local_num_cpus};
+		celerity::detail::config_testspy::set_mock_host_cfg(cfg, h_cfg);
+
+		auto device_selector = [](const mock_device& d) -> int { return d.get_type() == cl::sycl::info::device_type::accelerator ? 2 : 1; };
+
+		auto device = pick_device(cfg, device_selector, std::vector<mock_platform>{tp_0, tp_1, tp_2});
+		std::vector<mock_device> devices = tp_2.get_devices();
+		CHECK_THAT(lc.get_log(), Catch::Matchers::ContainsSubstring("device selector specified: platform"));
```
psalz, line 530: Why are you not checking for the full string here?

include/device_queue.h
almightyvats: Agreed and modified.

test/device_selection_tests.cc
```diff
@@ -0,0 +1,641 @@
...
+	tp_2.set_devices({td_2});
+
+	auto device = pick_device(cfg, celerity::detail::auto_select_device{}, std::vector<mock_platform>{tp_1, tp_2});
+	std::vector<mock_device> devices;
+	if(dv_type_1 == device_t::gpu || (dv_type_1 != device_t::gpu && dv_type_2 != device_t::gpu)) {
+		devices = tp_1.get_devices();
+	} else {
+		devices = tp_2.get_devices();
+	}
+	CHECK(device == devices[0]);
```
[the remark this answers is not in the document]
almightyvats: Agreed and modified.

test/device_selection_tests.cc
almightyvats: Since the test used the "generator" for random combinations. Do you think it's superfluous?

test/device_selection_tests.cc
almightyvats: At this point, I thought since I am anyway checking for the right device the whole message isn't necessary, but you're right it does look uncanny, I have now changed it to the full message.

include/device_queue.h
almightyvats: From my understanding, the test "pick_device doesn't choose devices across platform" tests the same thing. Also, we could use the `std::stable_sort` here, however, it is not memory efficient.

psalz: Okay yes, that test does indeed assert that behavior. However I'm wondering whether this is actually what we want. Because if we for example wanted to select 2 devices and had a score/platform array of the form `[(1, 0), (1, 1), (1, 1)]`, because platform `0` happened to only have one device with a non-negative score, we end up selecting nothing, even though platform `1` could supply two devices with the same score (`1`). While I don't think we should overcomplicate things, in this case the outcome of the selection only depends on the implementation-defined order in which platforms are enumerated, which also doesn't seem right.

And yes, I think we should be using `std::stable_sort` here!

test/device_selection_tests.cc
psalz: Yes, I'd say that's an implementation detail of the test case.

include/device_queue.h
PeterTh: Discussion today:
Let's select the top N devices, regardless of platform.
Warn the user iff (i) the devices are from different platforms or (ii) they are different device types.

include/device_queue.h
```diff
@@ -59,9 +65,170 @@ namespace detail {
...
+			const auto platform = device_from_platform.get_platform();
+			const auto device_type = device_from_platform.template get_info<sycl::info::device::device_type>();
+
+			for(size_t i = 1; i < host_cfg.node_count; ++i) {
+				auto [device_from_platform, idx] = devices_with_platform_idx[i];
+				if(device_from_platform.get_platform() != platform) { same_platform = false; }
+				if(device_from_platform.template get_info<sycl::info::device::device_type>() != device_type) { same_device_type = false; }
+			}
+
+			if(!same_platform) { CELERITY_WARN("Selected devices do not belong to the same platform"); }
```
psalz, line 101: I feel like having one warning for both would be sufficient here. We print the platform / device name anyways. So maybe `"Selected devices are of different type and/or do not belong to the same platform"`.

psalz (changes requested):

include/device_queue.h
almightyvats: Thank you for the comments. Yes, it makes sense.

include/device_queue.h
```diff
@@ -59,9 +65,170 @@ namespace detail {
...
 	};
 
+	// Try to find a platform that can provide a unique device for each node using a device selector.
+	template <typename DeviceT, typename PlatformT, typename SelectorT>
+	bool try_find_device_per_node(
+	    std::string& how_selected, DeviceT& device, const std::vector<PlatformT>& platforms, const host_config& host_cfg, SelectorT selector) {
+		std::vector<std::tuple<DeviceT, size_t>> devices_with_platform_idx;
+		for(size_t i = 0; i < platforms.size(); ++i) {
+			auto&& platform = platforms[i];
+			if(platform.get_devices().size() >= host_cfg.node_count) {
```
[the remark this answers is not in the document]
almightyvats: We don't need this check anymore, I have updated this piece of code and also added a test for the same.

PR #114, opened 2022-04-22
Replace single-use MPI datatypes with binary serialization

fknorr:
So far, we are creating one custom composite `MPI_Datatype` for each `MPI_Isend` to transfer variably-sized commands and buffer subranges. For reasons not investigated further, this yields abysmal performance at least on IBM Spectrum MPI.

This PR replaces those single-use MPI datatypes with a serialization into _frames_, contiguous `struct`s with a zero-sized array as the last member, a common idiom for typing a header together with variably-sized data. These are sent and received over MPI as simple byte streams. The smart pointer type `unique_frame_ptr` is introduced to allocate and manage frame types with arbitrary, variably-sized payloads.

As a downside, this change adds a semantically unnecessary allocation and copy of incoming buffer transfers: MPI receives `data_frame`s, whereas the buffer manager keeps uncommitted transfers as `raw_buffer_data`. This might be noticeable for very large transfers. The solution is likely to get rid of the owning `raw_buffer_data` altogehter, but this will affect a lot of code, so it should be addressed in a follow-up PR.


include/buffer_transfer_manager.h
```diff
@@ -32,19 +32,25 @@ namespace detail {
...
+			// variable-sized structure
+			data_frame() = default;
+			data_frame(const data_frame&) = delete;
+			data_frame& operator=(const data_frame&) = delete;
+
 			buffer_id bid;
 			reduction_id rid; // zero if this does not belong to a reduction
 			subrange<3> sr;
 			command_id push_cid;
+			alignas(std::max_align_t) payload_type data[0];
```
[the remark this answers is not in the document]
fknorr: Done (just realized it isn't technically needed here, but will be with #115 )

src/buffer_transfer_manager.cc
```diff
@@ -19,23 +20,28 @@ namespace detail {
...
-		transfer->header.rid = data.rid;
-		transfer->header.push_cid = pkg.cid;
-		transfer->data_type =
-		    mpi_support::build_single_use_composite_type({{sizeof(data_header), &transfer->header}, {transfer->data.get_size(), transfer->data.get_pointer()}});
+		CELERITY_TRACE("Ready to send {} of buffer {} ({} B) to {}", data.sr, data.bid, frame.get_payload_size(), data.target);
 
 		// Start transmitting data
-		MPI_Isend(MPI_BOTTOM, 1, *transfer->data_type, static_cast<int>(data.target), mpi_support::TAG_DATA_TRANSFER, MPI_COMM_WORLD, &transfer->request);
+		MPI_Request req;
+		MPI_Isend(frame.get_pointer(), static_cast<int>(frame.get_frame_size_bytes()), MPI_BYTE, static_cast<int>(data.target), mpi_support::TAG_DATA_TRANSFER,
```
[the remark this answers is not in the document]
fknorr: That is a real concern, added an assertion.

src/buffer_transfer_manager.cc
```diff
@@ -143,21 +144,22 @@ namespace detail {
...
 
 	void buffer_transfer_manager::commit_transfer(transfer_in& transfer) {
-		const auto& header = transfer.header;
-		size_t elem_size = transfer.data.get_size() / (header.sr.range[0] * header.sr.range[1] * header.sr.range[2]);
-		transfer.data.reinterpret(elem_size, cl::sycl::range<3>(header.sr.range[0], header.sr.range[1], header.sr.range[2]));
-		if(header.rid) {
+		const auto& frame = *transfer.frame;
+		const size_t elem_size = transfer.frame.get_payload_size() / (frame.sr.range[0] * frame.sr.range[1] * frame.sr.range[2]);
+		raw_buffer_data raw_data{elem_size, frame.sr.range};
+		memcpy(raw_data.get_pointer(), frame.data, raw_data.get_size());
```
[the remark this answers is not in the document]
fknorr: Yes, but getting rid of the copy is pretty involved and is addressed by the follow-up #115.

BlackMark29A (approved):
LGTM.

include/mpi_support.h
```diff
@@ -1,56 +1,97 @@
...
+	unique_frame_ptr() = default;
+
+	unique_frame_ptr(from_payload_size_tag, size_t payload_size) : impl(make_impl(frame_bytes_from_payload_size(payload_size))), payload_size(payload_size) {}
+
+	unique_frame_ptr(from_frame_bytes_tag, size_t frame_bytes) : impl(make_impl(frame_bytes)), payload_size(payload_size_from_frame_bytes(frame_bytes)) {}
+
+	unique_frame_ptr(unique_frame_ptr&& other) noexcept : impl(static_cast<impl&&>(other)), payload_size(other.payload_size) { other.payload_size = 0; }
+
+	unique_frame_ptr& operator=(unique_frame_ptr&& other) noexcept {
+		static_cast<impl&>(*this) = static_cast<impl&&>(other);
```
PeterTh, line 58: Please add a comment to note that this delegates to `unique_ptr` and correctly deletes (as per offline discussion).

include/mpi_support.h
```diff
@@ -1,56 +1,97 @@
...
+		try {
+			new(mem) Frame;
+		} catch(...) {
+			operator delete(mem);
+			throw;
+		}
+		return impl{static_cast<Frame*>(mem)};
+	}
+
+	static size_t frame_bytes_from_payload_size(size_t payload_size) { //
```
PeterTh, line 87: Might be better to name this something like `payload_count`. `payload_size` sounds like it's in bytes.

include/mpi_support.h
```diff
@@ -1,56 +1,97 @@
...
+		return sizeof(Frame) + sizeof(payload_type) * payload_size;
+	}
+
+	static size_t payload_size_from_frame_bytes(size_t frame_bytes) {
```
PeterTh, line 91: Same (`payload_count`).

src/buffer_transfer_manager.cc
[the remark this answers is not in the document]
PeterTh: We should work on resolving/mitigating this with high priority.

include/executor.h
```diff
@@ -80,13 +80,14 @@ namespace detail {
...
-		void create_job(const command_pkg& pkg, const std::vector<command_id>& dependencies, Args&&... args) {
+		void create_job(const unique_frame_ptr<command_frame>& frame, Args&&... args) {
+			const auto& pkg = frame->pkg;
 			jobs[pkg.cid] = {std::make_unique<Job>(pkg, std::forward<Args>(args)...), pkg.get_command_type(), {}, 0};
 
 			// If job doesn't exist we assume it has already completed.
 			// This is true as long as we're respecting task-graph (anti-)dependencies when processing tasks.
-			for(const command_id& d : dependencies) {
-				const auto it = jobs.find(d);
+			for(size_t i = 0; i < frame.get_payload_size(); ++i) {
```
PeterTh, line 89: Might be better here (and also for next `for` loop change) to have an interface that still allows range-based iteration. (As discussed offline)

PeterTh (changes requested):

include/command.h
```diff
@@ -220,5 +221,17 @@ namespace detail {
...
+	struct command_frame {
+		using payload_type = command_id;
+
+		// variable-sized structure
+		command_frame() = default;
+		command_frame(const command_frame&) = delete;
+		command_frame& operator=(const command_frame&) = delete;
+
+		command_pkg pkg;
+		payload_type dependencies[0];
```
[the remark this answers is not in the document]
fknorr: Done.

include/mpi_support.h
fknorr: Done.

include/mpi_support.h
fknorr: Done.

include/mpi_support.h
fknorr: Done.

include/executor.h
fknorr: Done.

fknorr:
As per offline discussion yesterday, I've decided to duplicate the dependency count inside `command_frame` to allow using `const command_frame &` without its owning `unique_frame_ptr`. The previous approach with "mixed queries" to `unique_frame_ptr` and `command_frame` will become worse with #101 anyway, when command frames have to transport both dependencies and conflicts.

PeterTh (approved):

psalz (changes requested):
One minor thing, otherwise LGTM!

include/mpi_support.h
```diff
@@ -1,56 +1,90 @@
...
+	unique_frame_ptr(from_payload_count_tag, size_t payload_count) : unique_frame_ptr(from_size_bytes, sizeof(Frame) + sizeof(payload_type) * payload_count) {}
+
+	unique_frame_ptr(from_size_bytes_tag, size_t size_bytes) : impl(make_frame(size_bytes)), size_bytes(size_bytes) {}
+
+	unique_frame_ptr(unique_frame_ptr&& other) noexcept : impl(static_cast<impl&&>(other)), size_bytes(other.size_bytes) { other.size_bytes = 0; }
+
+	unique_frame_ptr& operator=(unique_frame_ptr&& other) noexcept {
+		static_cast<impl&>(*this) = static_cast<impl&&>(other); // delegate to base class unique_ptr<Frame>::operator=() to delete previously held frame
+		size_bytes = other.size_bytes;
+		other.size_bytes = 0;
```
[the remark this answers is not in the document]
fknorr: Good catch!

psalz (approved):

PR #116, opened 2022-04-25
Add ComputeCpp 2.9.0 Mainline and Experimental

BlackMark29A:
This PR adds computecpp 2.9.0 mainline and experimental support.
I've looked at all computecpp workarounds in the codebase and they are all still necessary for this version.


psalz (approved):
LGTM

PR #119, opened 2022-06-08
Fix assertion checking which task types can have side effects

psalz:
The assertion logic was always backwards, but did not get triggered as the `side_effects` member variable was shadowed by the moved-from constructor parameter.


psalz:
(Fixed CI and changed assertion to be more comprehensible).

PeterTh (approved):
LGTM

fknorr (approved):

PR #120, opened 2022-06-13
Improve CI test system coverage

fknorr:
This PR updates the CI configuration to build the most recent version of each SYCL implementation against Ubuntu 22.04, while keeping Ubuntu 20.04 for older SYCL versions. For our pinned hipSYCL version (built with Ubuntu 20.04), the CUDA version is downgraded to 11.0.3, while for hipSYCL HEAD (built with Ubuntu 22.04) CUDA is upgraded to 11.7.

This also adds platform support documentation listing the systems Celerity intends to support together with a table of the precise configurations tested in CI.

**Question:** Do we want to list Windows / MSVC as a supported platform? If so, for which SYCL implementation?

### DIFF
--- a/.github/workflows/celerity_ci.yml
+++ b/.github/workflows/celerity_ci.yml
@@ -27,6 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # make sure to keep this in sync with docs/platform-support.md
         include:
           - sycl: "computecpp"
             sycl-version: "2.6.0"
--- a/.github/workflows/celerity_ci.yml
+++ b/.github/workflows/celerity_ci.yml
@@ -28,65 +28,84 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - sycl-version: "computecpp:2.6.0"
+          - sycl: "computecpp"
+            sycl-version: "2.6.0"
+            ubuntu-version: "20.04"
             platform: "intel"
             build-type: "Debug"
-          - sycl-version: "computecpp:2.7.0"
+          - sycl: "computecpp"
+            sycl-version: "2.7.0"
+            ubuntu-version: "20.04"
             platform: "intel"
             build-type: "Debug"
-          - sycl-version: "computecpp:2.8.0"
+          - sycl: "computecpp"
+            sycl-version: "2.8.0"
+            ubuntu-version: "20.04"
             platform: "intel"
             build-type: "Debug"
-          - sycl-version: "computecpp:2.9.0"
+          - sycl: "computecpp"
+            sycl-version: "2.9.0"
+            ubuntu-version: "22.04"
             platform: "intel"
             build-type: "Debug"
-          - sycl-version: "computecpp:2.9.0"
+          - sycl: "computecpp"
+            sycl-version: "2.9.0"
+            ubuntu-version: "22.04"
             platform: "intel"
             build-type: "Release"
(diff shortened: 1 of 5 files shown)


docs/platform-support.md
```diff
@@ -0,0 +1,35 @@
...
+| "          | 2.7.0                                                                               | "            | "          |
+| "          | 2.8.0                                                                               | "            | "          |
+| "          | 2.9.0                                                                               | Ubuntu 22.04 | "          |
+| "          | 2.9.0 (Experimental Compiler)                                                       | "            | "          |
+| "          | 2.9.0                                                                               | "            | Release    |
+| "          | 2.9.0 (Experimental Compiler)                                                       | "            | "          |
+| DPC++      | [`7735139b`](https://github.com/intel/llvm/commit/7735139b)                         | Ubuntu 20.04 | Debug      |
+| "          | [`HEAD`](https://github.com/intel/llvm/)                                            | Ubuntu 22.04 | "          |
+| "          | "                                                                                   | Ubuntu 22.04 | Release    |
```
BlackMark29A, line 32: `Ubuntu 22.04` is repeated here, without using `"`, which is usued consistently everywhere else.

.github/workflows/celerity_ci.yml
```diff
@@ -28,65 +28,84 @@ jobs:
       fail-fast: false
       matrix:
         include:
```
psalz, line 30: I think we should add a note here that whenever this is updated, `docs/platform-support.md` should also be updated!

psalz (changes requested):
Thanks! Could you also link the new document in [`sidebars.json`](https://github.com/celerity/celerity-runtime/blob/master/website/sidebars.json) for the website, as well as the [docs entrypoint](https://raw.githubusercontent.com/celerity/celerity-runtime/master/docs/README.md)?

docs/platform-support.md
```diff
@@ -0,0 +1,35 @@
+# Officially Supported Platforms
```
psalz, line 1: Ah, one more thing, please also add the YAML preamble for the website, e.g. something like:

```
---
id: platform-support
title: Officially Supported Platforms
sidebar_label: Platform Support
---
```

docs/platform-support.md
```diff
@@ -0,0 +1,35 @@
...
+  * with [stable](https://developer.codeplay.com/products/computecpp/ce/download) and [experimental](https://developer.codeplay.com/products/computecpp/ce/download?experimental=true) compilers
+* DPC++ ≥ revision [`7735139b`](https://github.com/intel/llvm/commit/7735139b)
+  * on Intel hardware
+
+## Continuously Tested Configurations
+
+We automatically verify Celerity's build process and test suites against a select number of system configurations.
+Those are:
+
+| SYCL       | SYCL version                                                                        | OS           | Build type |
```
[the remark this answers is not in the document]
fknorr: I've recombined some rows to make the table more human-readable.

.github/workflows/celerity_ci.yml
fknorr: Done.

docs/platform-support.md
fknorr: Done.

psalz (approved):
LGTM, maybe rebase to get the hipSYCL fix in?

PeterTh (approved):

PR #121, opened 2022-06-13
Add support for ComputeCpp 2.10.0

fknorr:
**Note:** This PR is based on #120 , so that one has to be merged first.

ComputeCpp 2.10.0 works with Celerity out of the box, however the SYCL 2020 support has improved significantly with the new version, rendering existing workarounds unnecessary. These are now deactivated for ComputeCpp ≥ 2.10.0.


include/item.h
```diff
@@ -550,6 +550,6 @@ T inclusive_scan_over_group(const group<Dims>& g, V x, BinaryOperation binary_op
 #endif
 }
 
-#endif // !WORKAROUND_COMPUTECPP
+#endif // !WORKAROUND_COMPUTECPP || WORKAROUND(COMPUTECPP, 2, 10)
```
PeterTh, line 553: Why is there an end of `WORKAROUND(COMPUTECPP, 2, 10)` indicated here? I don't see a corresponding opening (just `2, 9`.

PeterTh (changes requested):
It would be nice if the `WORKAROUND` macro

1. was prefixed, e.g. `CELERITY_WORKAROUND`
2. had some documentation on what exactly `WORKAROUND(BLA, X, Y)` means, semantically

Looks good other than this (and the one line comment).

include/handler.h
```diff
@@ -59,7 +59,7 @@ namespace detail {
 	template <typename KernelName>
 	constexpr bool is_unnamed_kernel = std::is_same_v<KernelName, unnamed_kernel>;
 
-#if WORKAROUND_COMPUTECPP
+#if WORKAROUND_COMPUTECPP // TODO only required for the stable, but not the experimental compiler
```
psalz, line 62: Can't you use `CELERITY_IS_OLD_COMPUTECPP_COMPILER` for this?

psalz (approved):
Thanks, LGTM!

include/item.h
fknorr: Removed.

include/handler.h
fknorr: Good idea, done.

fknorr:
I have refactored the `WORKAROUND` mechanism to introduce namespacing and clarify semantics of the version check.

Workaround guards for all versions of a SYCL implementation:
```
#if CELERITY_WORKAROUND(HIPSYCL)
...
#endif
```

Workaround for a range of older SYCL implementation versions:

```
#if CELERITY_WORKAROUND_LESS_OR_EQUAL(COMPUTECPP, 2, 6)
...
#endif
```

There was one test `pick_device supports passing a device selector function` which had been incorrectly disabled for all backends because it was guarded by `#if !defined(WORKAROUND_COMPUTECPP)` instead of `#if !WORKAROUND_COMPUTECPP`. The new syntax which always requires parantheses for invocation hopefully avoids mistakes like this in the future.

PeterTh (approved):

PR #122, opened 2022-07-13
Add clang-tidy configuration; check during CI

psalz:
This change is based on #120, so that one should be merged first.

This adds a clang-tidy configuration with customized settings for (hopefully) all of our naming conventions, including the newly agreed upon `m_` prefix for private class members. I've also enabled all `readability-`, `performance-`, `cppcoreguidelines-` and `mpi-` checks. We'll have to see if we need to disable / configure some of those if they are too annoying (in particular the cppcore guidelines seem to be quite opinionated).

CI now also runs clang-tidy on the lines that have been changed in a PR, with a bot adding code comments (including quick-fixes, where possible) for all warnings. Obviously this is a last resort and ideally you should update your IDE to use the new config file so you can catch those diagnostics while writing the code.

In an ideal world our entire codebase would pass the clang-tidy checks, but we're not there yet (by a long shot), so I think doing it incrementally when changing existing code is a decent approach for now.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -26,6 +26,9 @@ CheckOptions:
     value: lower_case
   - key: readability-identifier-naming.LocalVariableCase
     value: lower_case
+  - key: readability-identifier-naming.LocalVariableIgnoredRegexp
+    # Allow single-letter uppercase variable names
+    value: "[A-Z]"
   - key: readability-identifier-naming.ParameterCase
     value: lower_case
   - key: readability-identifier-naming.PrivateMemberCase
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -11,7 +11,19 @@
 #
 InheritParentConfig: false
 # See https://clang.llvm.org/extra/clang-tidy/checks/list.html for a full list of available checks.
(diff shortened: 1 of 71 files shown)


psalz:
I would suggest we only merge this once all other currently open PRs have been merged, to avoid having to adjust private class member naming everywhere. In the meantime please make sure to use the [clang-tidy config](https://github.com/celerity/celerity-runtime/blob/clang-tidy/.clang-tidy) when writing new code!

PeterTh:
Discussion result:
- compile a list of failing checks with instance counts
- investigate based on this list what we want to enable immediately and how much work that entails

psalz:
I've fixed all naming issues and disabled all other checks for now. I suggest we merge this now as I don't want to do the whole thing again (it wasn't automagic at all!) and decide on which other rules we want to enable in a separate issue/PR.

BlackMark29A (approved):
LGTM.

PeterTh (approved):
LGTM.

- The automated changes look fine.
- I don't like the capitalization change regarding enums, but I accept the reasoning.
- I don't claim to fully understand what's going on in the CI integration.

PR #123, opened 2022-07-13
Update authoritative version of clang-format to 14

psalz:
This thankfully is mostly a documentation change, except for one bug fix in connection with deduction guides.

Also change the `AllowShortFunctionsOnASingleLine` rule from `Inline` to `true`. This rule seems to be extremely buggy, and clang-format's behavior for `Inline` starting with version 13 does not make any sense. Earlier versions also didn't treat the rule correctly (basically all short functions were put on a single line, regardless of their position in the code), which is what `true` is supposed to do (at least to my understanding).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,8 @@
+# Clang-format coding style definitions for Celerity
+#
+# Due to some inconsistencies between releases we currently
+# use clang-format 14 as the authoritative version.
+
 BasedOnStyle: llvm
 
 AlignAfterOpenBracket: DontAlign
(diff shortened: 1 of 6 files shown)


psalz:
Looks like I missed 3 files, updated.

PeterTh (approved):
LGTM.

I don't think I like `AllowShortFunctionsOnASingleLine` at this point, but if we use it I agree that it should at least be consistent.

psalz:
> LGTM.
> 
> I don't think I like `AllowShortFunctionsOnASingleLine` at this point, but if we use it I agree that it should at least be consistent.

I'm not a big fan either; I think with the `Inline` setting it can make sense (for getters / setters), but that's not what we got. Changing it to `false` now unfortunately is a pretty big diff, not sure if worth.

fknorr (approved):
LGTM

PR #124, opened 2022-07-14
Fix explicit DAG flushing for hipSYCL

psalz:
hipSYCL [recently changed](https://github.com/illuhad/hipSYCL/pull/749) its internal APIs for explicitly flushing its DAG (which is a workaround we need to ensure progress). To maintain backwards compatibility with older hipSYCL versions, this introduces a compile-time check to determine which of the APIs is available.

Note that our CI containers haven't been rebuilt in a while due to backend changes required for #120, so `hipsycl:HEAD` does not actually contain the aforementioned change. See https://github.com/celerity/celerity-runtime/runs/7322401874?check_suite_focus=true for a run that failed due to that change.

### DIFF
--- a/include/device_queue.h
+++ b/include/device_queue.h
@@ -18,6 +18,25 @@ namespace detail {
 
 	class task;
 
+#if WORKAROUND_HIPSYCL
+	template <typename T, typename U = void>
+	struct hipsycl_is_old_dag_api : std::false_type {};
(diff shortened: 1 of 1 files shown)


fknorr (approved):
LGTM, thank you!

PeterTh (approved):
LGTM.

PR #125, opened 2022-07-20
Dry run

facuMH:
The idea behind this is to able to simulate a run to analyze behavior and how would celerity scale to larger amounts nodes, without necessarily having those nodes.
For this we are not flushing commands unless they are epochs ( and this ones only to the master node).

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -40,6 +40,8 @@ namespace detail {
 		 */
 		const std::optional<device_config>& get_device_config() const { return m_device_cfg; };
 		std::optional<bool> get_enable_device_profiling() const { return m_enable_device_profiling; };
+		bool is_dry_run() const { return m_dry_run_nodes > 0; };
+		int get_dry_run_nodes() const { return m_dry_run_nodes; }
 
 		size_t get_graph_print_max_verts() const { return m_graph_print_max_verts; };
 
@@ -49,6 +51,7 @@ namespace detail {
 		std::optional<device_config> m_device_cfg;
 		std::optional<bool> m_enable_device_profiling;
 		size_t m_graph_print_max_verts = 200;
+		int m_dry_run_nodes = 0;
 	};
 
 } // namespace detail
--- a/include/runtime.h
+++ b/include/runtime.h
@@ -77,6 +77,8 @@ namespace detail {
 
 		host_object_manager& get_host_object_manager() const;
 
+		bool is_dry_run() const { return m_cfg->is_dry_run(); }
+
 	  private:
 		inline static bool m_mpi_initialized = false;
 		inline static bool m_mpi_finalized = false;
--- a/src/config.cc
+++ b/src/config.cc
@@ -180,6 +180,16 @@ namespace detail {
 			const auto result = get_env("CELERITY_FORCE_WG");
 			if(result.first) { CELERITY_WARN("Support for CELERITY_FORCE_WG has been removed with Celerity 0.3.0."); }
 		}
+
+		// -------------------------------- CELERITY_DRY_RUN_NODES ---------------------------------
+		{
+			const auto [is_set, value] = get_env("CELERITY_DRY_RUN_NODES");
+			if(is_set) {
+				const auto [is_valid, num_nodes] = parse_uint(value.c_str());
+				if(!is_valid) { CELERITY_WARN("CELERITY_DRY_RUN_NODES contains invalid value - will be ignored"); }
+				m_dry_run_nodes = num_nodes;
+			}
+		}
 	}
 } // namespace detail
 } // namespace celerity
--- a/src/runtime.cc
+++ b/src/runtime.cc
@@ -102,13 +102,19 @@ namespace detail {
 		MPI_Comm_size(MPI_COMM_WORLD, &world_size);
 		m_num_nodes = world_size;
 
+		m_cfg = std::make_unique<config>(argc, argv);
+		if(m_cfg->is_dry_run()) {
+			if(m_num_nodes != 1) throw std::runtime_error("In order to run with CELERITY_DRY_RUN_NODES a single MPI process/rank must be used.\n");
+			m_num_nodes = m_cfg->get_dry_run_nodes();
+			CELERITY_WARN("Performing a dry run with {} simulated nodes", m_num_nodes);
+		}
+
 		int world_rank;
 		MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);
 		m_local_nid = world_rank;
 
 		spdlog::set_pattern(fmt::format("[%Y-%m-%d %H:%M:%S.%e] [{:0{}}] [%^%l%$] %v", world_rank, int(ceil(log10(world_size)))));
 
-		m_cfg = std::make_unique<config>(argc, argv);
 #ifndef __APPLE__
 		if(const uint32_t cores = affinity_cores_available(); cores < min_cores_needed) {
 			CELERITY_WARN("Celerity has detected that only {} logical cores are available to this process. It is recommended to assign at least {} "
@@ -257,6 +263,10 @@ namespace detail {
 	}
 
 	void runtime::flush_command(node_id target, unique_frame_ptr<command_frame> frame) {
+		if(is_dry_run()) {
+			// We only want to send epochs to the master node for slow full sync and shutdown.
+			if(target != 0 || frame->pkg.get_command_type() != command_type::epoch) return;
+		}
 		// Even though command packages are small enough to use a blocking send we want to be able to send to the master node as well,
 		// which is why we have to use Isend after all. We also have to make sure that the buffer stays around until the send is complete.
 		MPI_Request req;
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -12,6 +12,7 @@
 
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
 #include <catch2/matchers/catch_matchers_all.hpp>
 
 #include <celerity.h>
@@ -47,6 +48,8 @@ namespace detail {
 		static scheduler& get_schdlr(runtime& rt) { return *rt.m_schdlr; }
 
 		static executor& get_exec(runtime& rt) { return *rt.m_exec; }
+
+		static size_t get_command_count(runtime& rt) { return rt.m_cdag->command_count(); }
 	};
 
 	struct scheduler_testspy {
@@ -992,6 +995,31 @@ namespace detail {
 	}
 
 #endif
+
+	void dry_run_with_nodes(const size_t nodes) {
+		const std::string dryrun_envvar_name = "CELERITY_DRY_RUN_NODES";
+		const auto ste = test_utils::set_test_env(dryrun_envvar_name, std::to_string(nodes));
+
+		distr_queue q;
+
+		auto& rt = runtime::get_instance();
+		auto& tm = rt.get_task_manager();
+		tm.set_horizon_step(2);
+
+		REQUIRE(rt.is_dry_run());
+
+		q.submit([=](handler& cgh) { cgh.host_task(range<1>{nodes * 2}, [](partition<1>) {}); });
+		q.slow_full_sync();
+
+		// (intial epoch + task + sync epoch) per node.
+		CHECK(runtime_testspy::get_command_count(rt) == 3 * nodes);
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run generates commands for an arbitrary number of simulated worker nodes", "[dryrun]") {
+		const size_t nodes = GENERATE(values({4, 8, 16}));
+		dry_run_with_nodes(nodes);
+	}
 
 } // namespace detail
 } // namespace celerity
--- a/test/test_utils.h
+++ b/test/test_utils.h
@@ -6,6 +6,12 @@
 #include <string>
 #include <unordered_set>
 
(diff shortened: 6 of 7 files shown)


src/runtime.cc
```diff
@@ -102,13 +102,18 @@ namespace detail {
 		MPI_Comm_size(MPI_COMM_WORLD, &world_size);
 		num_nodes = world_size;
 
+		cfg = std::make_unique<config>(argc, argv);
+		if(is_dry_run()) {
+			if(num_nodes != 1) throw std::runtime_error("YOU CAN NOT MAKE A DRY RUN WITH MPI.\n");
```
psalz, line 107:
```suggestion
			if(num_nodes != 1) throw std::runtime_error("Cannot perform a dry run with more than one rank.\n");
```

src/config.cc
```diff
@@ -180,6 +180,16 @@ namespace detail {
 			const auto result = get_env("CELERITY_FORCE_WG");
 			if(result.first) { CELERITY_WARN("Support for CELERITY_FORCE_WG has been removed with Celerity 0.3.0."); }
 		}
+
+		// -------------------------------- CELERITY_DRY_RUN ---------------------------------
+		{
+			const auto result = get_env("CELERITY_DRYRUN_NODES");
+			if(result.first) {
+				dry_run = std::stoi(result.second) > 0;
```
psalz, line 188: You can use the `parse_uint` helper here to make this more robust!

src/config.cc
```diff
@@ -180,6 +180,16 @@ namespace detail {
...
+				dry_run_nodes = std::stoi(result.second);
+				CELERITY_WARN("Celerity Dry Run detected.");
```
psalz, line 190: As this is not a validation warning/error, I would suggest moving it to the runtime constructor! Also maybe rephrase to `"Performing a dry run with {} simulated nodes."`.

include/config.h
```diff
@@ -40,6 +40,8 @@ namespace detail {
 		 */
 		const std::optional<device_config>& get_device_config() const { return device_cfg; };
 		std::optional<bool> get_enable_device_profiling() const { return enable_device_profiling; };
+		std::optional<bool> is_dry_run() const { return dry_run; };
```
psalz, line 43: I think we could just default to `false` if it is not set and void the `optional`.

include/runtime.h
```diff
@@ -77,6 +77,10 @@ namespace detail {
 
 		host_object_manager& get_host_object_manager() const;
 
+		bool is_dry_run() const { return cfg->is_dry_run().value_or(false); }
+
+		int get_command_count() const;
```
psalz, line 82: I'd prefer not introducing to new members to the runtime's API for a single test case. Why not use `runtime_testspy` instead?

test/runtime_tests.cc
```diff
@@ -992,5 +992,39 @@ namespace detail {
 		});
 	}
 
+	void dry_run_with_nodes(const int nodes) {
+		const std::string dryrun_envvar_name = "CELERITY_DRYRUN_NODES";
+		setenv(dryrun_envvar_name.c_str(), std::to_string(nodes).c_str(), 1);
```
psalz, line 997: For the device selection tests we've used a `config_testspy` instead of modifying the environment, imo that's a cleaner approach.

test/runtime_tests.cc
```diff
@@ -992,5 +992,39 @@ namespace detail {
...
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "dry run", "[dryrun]") {
+		SECTION("with 2 nodes") { dry_run_with_nodes(2); }
+		SECTION("with 8 nodes") { dry_run_with_nodes(8); }
+		SECTION("with 16 nodes") { dry_run_with_nodes(16); }
+		SECTION("with 32 nodes") { dry_run_with_nodes(32); }
+		SECTION("with 64 nodes") { dry_run_with_nodes(64); }
+		SECTION("with 128 nodes") { dry_run_with_nodes(128); }
```
psalz, line 1025: Use a generator instead? Also, maybe 2 or 3 configs are sufficient?

test/runtime_tests.cc
```diff
@@ -992,5 +992,39 @@ namespace detail {
...
+			cgh.parallel_for<class UKN(dryrun)>(buff.get_range(), [=](item<1> item) { acc[item.get_id()] = item.get_id().get(0); });
+		});
+		q.slow_full_sync();
+
+		// (intial epoch + task + sync epoch) per node.
+		CHECK(rt.get_command_count() == 3 * nodes);
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "dry run", "[dryrun]") {
```
psalz, line 1019:
```suggestion
	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry runs generate graphs for an arbitrary number of simulated worker nodes", "[dryrun]") {
```

test/runtime_tests.cc
```diff
@@ -992,5 +992,39 @@ namespace detail {
...
+		auto& tm = rt.get_task_manager();
+		tm.set_horizon_step(2);
+
+		CHECK(rt.is_dry_run());
+
+		buffer<size_t, 1> buff{nodes * 2};
+		q.submit([=](handler& cgh) {
+			accessor acc{buff, cgh, celerity::access::one_to_one{}, write_only};
+			cgh.parallel_for<class UKN(dryrun)>(buff.get_range(), [=](item<1> item) { acc[item.get_id()] = item.get_id().get(0); });
+		});
```
psalz, line 1011:
```suggestion
		q.submit([=](handler& cgh) {
			cgh.host_task(range<1>{nodes * 2}, [] {});
		});
```

src/config.cc
```diff
@@ -180,6 +180,16 @@ namespace detail {
 			const auto result = get_env("CELERITY_FORCE_WG");
 			if(result.first) { CELERITY_WARN("Support for CELERITY_FORCE_WG has been removed with Celerity 0.3.0."); }
 		}
+
+		// -------------------------------- CELERITY_DRY_RUN ---------------------------------
```
psalz, line 184:
```suggestion
		// -------------------------------- CELERITY_DRYRUN_NODES ---------------------------------
```

psalz (changes requested):
Thanks! I've added some notes. Also, I'm wondering whether we should call the variable `CELERITY_DRY_RUN_NODES` instead? 🤔

include/config.h
PeterTh, line 43: I don't see why this needs to be a `std::optional<bool>`, when it could just be a `bool`. I don't think we'll ever need to distinguish "dry run not set" from "not a dry run".

test/runtime_tests.cc
PeterTh: I feel like using setenv in the unit tests is fine. It's slightly more integration- than unit-testy, but eh.

PeterTh (commented):
Looks pretty good, as long as you take care of the comments. Also, DON'T SCREAM AT THE USER :P

test/runtime_tests.cc
facuMH: not sure if we can do the same here, because in `device_selection_tests` the `config_testspy` is modifying a hand-made config, not the config that is initialized inside the runtime, which is what one of the things we want to test here.

psalz: Right, because it's setting `num_nodes` in the constructor. In any case, modifying the global environment from a test is always iffy. What happens when the test fails? You'd need at least a RAII wrapper to ensure that the environment is always restored.

But maybe the cleanest solution would be to change the runtime ctor, like so:
```diff
-	runtime::runtime(int* argc, char** argv[], device_or_selector user_device_or_selector) {
+	runtime::runtime(int* argc, char** argv[], config cfg = {argc, argv}, device_or_selector user_device_or_selector) {
```

test/test_utils.h
```diff
@@ -427,6 +427,17 @@ namespace test_utils {
 		}
 	}
 
+	struct set_test_env {
+		set_test_env(const std::string& evn, const std::string val) {
+			setenv(evn.c_str(), val.c_str(), 1);
```
psalz, line 432: It looks like `setenv` does not exist on Windows, so this is not portable unfortunately. Also if you want this to be a generic utility (which I assume as it is in `test_utils.h`) I think a more robust solution would be to restore the previous value upon destruction.

psalz (changes requested):

test/test_utils.h
```diff
@@ -427,6 +427,17 @@ namespace test_utils {
 		}
 	}
 
+	struct set_test_env {
+		set_test_env(const std::string& evn, const std::string val) {
```
psalz, line 431:
```suggestion
		set_test_env(const std::string& env, const std::string val) {
```

test/runtime_tests.cc
```diff
@@ -992,5 +995,30 @@ namespace detail {
...
+
+		q.submit([=](handler& cgh) { cgh.host_task(range<1>{nodes * 2}, [](partition<1>) {}); });
+		q.slow_full_sync();
+
+		// (intial epoch + task + sync epoch) per node.
+		CHECK(runtime_testspy::get_command_count(rt) == 3 * nodes);
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run generate grap", "[dryrun]") {
```
psalz, line 1018: Something went wrong with the name change here!

src/runtime.cc
```diff
@@ -102,13 +102,19 @@ namespace detail {
 		MPI_Comm_size(MPI_COMM_WORLD, &world_size);
 		num_nodes = world_size;
 
+		cfg = std::make_unique<config>(argc, argv);
+		if(is_dry_run()) {
```
psalz, line 106:
```suggestion
		if(cfg->is_dry_run()) {
```

A minor thing, but otherwise it's not clear that config must exist before `runtime::is_dry_run` can be called.

test/runtime_tests.cc
```diff
@@ -992,5 +995,30 @@ namespace detail {
...
+		q.slow_full_sync();
+
+		// (intial epoch + task + sync epoch) per node.
+		CHECK(runtime_testspy::get_command_count(rt) == 3 * nodes);
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run generates commands for an arbitrary number of simulated worker nodes", "[dryrun]") {
+		const size_t nodes = GENERATE(values({4, 8, 16}));
+		SECTION("with {} nodes", std::to_string(nodes).c_str()) { dry_run_with_nodes(nodes); }
```
psalz, line 1020:
```suggestion
		SECTION("with " << nodes << " nodes", dry_run_with_nodes(nodes); }
```

Catch2 does not support fmt-style formatting.

test/test_utils.h
```diff
@@ -427,6 +432,50 @@ namespace test_utils {
...
+		set_test_env(const std::string& env, const std::string val) {
+			name_size = GetEnvironmentVariableA(env.c_str(), nullptr, 0);
+			if(name_size > 0) {
+				original_value.resize(name_size);
+				GetEnvironmentVariableA(env.c_str(), &original_value[0], name_size + 1);
+			}
+			SetEnvironmentVariable(env.c_str(), val.c_str());
+			env_var_name = env;
+		}
+		// if it was not previously set, then original_value is 0. using setEnvVar with 0 deletes the env var.
```
psalz, line 452: This comment seems outdated?

test/test_utils.h
```diff
@@ -427,6 +432,50 @@ namespace test_utils {
...
+		~set_test_env() {
+			if(original_value.size()) {
```
psalz, line 454:
```suggestion
			if(!original_value.empty()) {
```

test/test_utils.h
```diff
@@ -427,6 +432,50 @@ namespace test_utils {
...
+
+	  public:
+		set_test_env(const std::string& env, const std::string val) {
+			const char* has_value = std::getenv(env.c_str());
+			if(has_value != nullptr) { original_value = has_value; }
+			setenv(env.c_str(), val.c_str(), 1);
+			env_var_name = env;
+		}
+		~set_test_env() {
+			if(!original_value.size()) {
```
psalz, line 470: The branches here are reversed, which is somewhat confusing.

psalz (changes requested):
I've added a few more notes. Also I vote in favor of renaming the env var to `CELERITY_DRY_RUN_NODES` - thoughts?

BlackMark29A (changes requested):
I've identified some spots where readability/documentation could be improvied, but there are also some bugs which crept in:

BlackMark29A (changes requested):

BlackMark29A (approved):
Thank you. Looks good to me now.

src/runtime.cc
```diff
@@ -102,13 +102,19 @@ namespace detail {
 		MPI_Comm_size(MPI_COMM_WORLD, &world_size);
 		m_num_nodes = world_size;
 
+		m_cfg = std::make_unique<config>(argc, argv);
+		if(m_cfg->is_dry_run()) {
+			if(m_num_nodes != 1) throw std::runtime_error("Cannot perform a dry run with more than one rank.\n");
```
fknorr, line 107: That error message might be confusing, it can be read as "The only allowed parameter value for CELERITY_DRY_RUN_NODES is 1". Maybe go with "Celerity must be launched with a single MPI process when specifying CELERITY_DRY_RUN_NODES" or similar.

fknorr (approved):
LGTM up to the error message, thanks!

psalz (approved):
LGTM!

PR #129, opened 2022-08-02
Update catch2 to v3.1.0

BlackMark29A:
Catch2 has introduced cmake options in v3.0.1 to more easily set configuration defines. This PR updates catch2 to the latest version to allow an upcoming PR to make use of them.

PR #130, opened 2022-08-02
Fix compilation and runtime issues on windows

BlackMark29A:
This PR is based on #129, so that one must be merged first.

Some code changes in past PRs have introduced code that no longer compiles on windows, this PR remedies that by adding missing headers and some minor code tweaks.

Additionally, PR #106 refactored the unit tests, specifically changing when `MPI_Init` is called relative to when Catch2 is initialized. This surfaced an issue where MSMPI changes the top-level exception filter, after Catch2 has already changed it, resulting in a sanity check inside Catch2 failing, where it checks that the previous filter was restored correctly. This PR disables windows SEH for Catch2, mitigating the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,11 @@ macro(add_submodule_directory RELPATH)
 endmacro()
 
 add_submodule_directory(vendor/spdlog)
+if(WIN32)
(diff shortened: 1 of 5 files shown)


PeterTh (approved):
LGTM.

fknorr (approved):

PR #132, opened 2022-08-03
Add buffer debug names

facuMH:
For now this only shows the buffer names plus ids instead of just ids when printing the command graph, but in a future it could also be used for instrumenting with some tracer/profiler and facilitate getting a more relevant name than just the buffer id. 

Also it is the first of hopefully more debug functionalities.

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -1,0 +1,18 @@
+#include <string>
+
+#include "buffer.h"
+
+namespace celerity {
+namespace debug {
+	template <typename DataT, int Dims>
+	void set_buffer_name(const celerity::buffer<DataT, Dims>& buff, const std::string& debug_name) {
+		detail::set_buffer_name(buff, debug_name);
+		detail::runtime::get_instance().get_buffer_manager().set_debug_name(detail::get_buffer_id(buff), debug_name);
+	}
+	template <typename DataT, int Dims>
+	std::string get_buffer_name(const celerity::buffer<DataT, Dims>& buff) {
+		return detail::get_buffer_name(buff);
+	}
+
+} // namespace debug
+} // namespace celerity
--- a/src/print_graph.cc
+++ b/src/print_graph.cc
@@ -34,7 +34,16 @@ namespace detail {
 		}
 	}
 
-	void format_requirements(std::string& label, const task& tsk, subrange<3> execution_range, access_mode reduction_init_mode, const reduction_manager& rm) {
+	std::string get_buffer_label(const buffer_manager* bm, const buffer_id bid) {
+		// if there is no buffer manager or no name defined, the name will be the buffer id.
+		// if there is a name we want "id name"
+		std::string name;
+		if(bm != nullptr) { name = bm->get_debug_name(bid); }
+		return !name.empty() ? fmt::format("B{} \"{}\"", bid, name) : fmt::format("B{}", bid);
+	}
+
+	void format_requirements(std::string& label, const task& tsk, subrange<3> execution_range, access_mode reduction_init_mode, const reduction_manager& rm,
+	    const buffer_manager* bm) {
 		for(auto rid : tsk.get_reductions()) {
 			auto reduction = rm.get_reduction(rid);
 
@@ -43,15 +52,17 @@ namespace detail {
 
 			const auto bid = reduction.output_buffer_id;
 			const auto req = GridRegion<3>{{1, 1, 1}};
-			fmt::format_to(std::back_inserter(label), "<br/>(R{}) <i>{}</i> B{} {}", rid, detail::access::mode_traits::name(rmode), bid, req);
+			const std::string bl = get_buffer_label(bm, bid);
+			fmt::format_to(std::back_inserter(label), "<br/>(R{}) <i>{}</i> {} {}", rid, detail::access::mode_traits::name(rmode), bl, req);
 		}
 
 		const auto& bam = tsk.get_buffer_access_map();
 		for(const auto bid : bam.get_accessed_buffers()) {
 			for(const auto mode : bam.get_access_modes(bid)) {
 				const auto req = bam.get_requirements_for_access(bid, mode, tsk.get_dimensions(), execution_range, tsk.get_global_size());
+				const std::string bl = get_buffer_label(bm, bid);
 				// While uncommon, we do support chunks that don't require access to a particular buffer at all.
-				if(!req.empty()) { fmt::format_to(std::back_inserter(label), "<br/><i>{}</i> B{} {}", detail::access::mode_traits::name(mode), bid, req); }
+				if(!req.empty()) { fmt::format_to(std::back_inserter(label), "<br/><i>{}</i> {} {}", detail::access::mode_traits::name(mode), bl, req); }
 			}
 		}
 
@@ -60,7 +71,7 @@ namespace detail {
 		}
 	}
 
-	std::string get_task_label(const task& tsk, const reduction_manager& rm) {
+	std::string get_task_label(const task& tsk, const reduction_manager& rm, const buffer_manager* bm) {
 		std::string label;
 		fmt::format_to(std::back_inserter(label), "T{}", tsk.get_id());
 		if(!tsk.get_debug_name().empty()) { fmt::format_to(std::back_inserter(label), " \"{}\" ", tsk.get_debug_name()); }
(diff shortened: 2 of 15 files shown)


include/debug.h
```diff
@@ -0,0 +1,17 @@
+#include <string>
+
+#include "buffer.h"
+
+namespace celerity {
+namespace debug {
+	template <typename BuffType>
```
PeterTh, line 7: Should we constrain this template somehow?

src/graph_generator.cc
```diff
@@ -402,7 +402,11 @@ namespace detail {
 					if(auto rid_iter = buffer_reduction_map.find(bid); rid_iter != buffer_reduction_map.end()) {
 						cmd->debug_label += fmt::format("(R{}) ", rid_iter->second);
 					}
-					cmd->debug_label += fmt::format("{} {} {}\\n", detail::access::mode_traits::name(mode), bid, toString(req));
+
+					std::string buffer_debug_name{};
+					if(runtime::is_initialized()) buffer_debug_name = runtime::get_instance().get_buffer_manager().get_debug_name(bid);
+					if(buffer_debug_name.length() == 0) { buffer_debug_name = std::to_string(bid); }
+					cmd->debug_label += fmt::format("{} {}-{} {}\\n", detail::access::mode_traits::name(mode), buffer_debug_name, bid, toString(req));
```
PeterTh, line 409: Please move this extra complexity to a separate helper function.

test/buffer_manager_tests.cc
```diff
@@ -1098,6 +1098,14 @@ namespace detail {
 		}
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "set buffer debug name", "[buffer_debug_name]") {
```
PeterTh, line 1101: Maybe also add a test that checks that the name shows up in the generated graph.

PeterTh (changes requested):

test/graph_generation_tests.cc
```diff
@@ -792,5 +792,30 @@ namespace detail {
...
+		auto& tm = ctx.get_task_manager();
+		auto& ggen = ctx.get_graph_generator();
+		auto& inspector = ctx.get_inspector();
+		auto& cdag = ctx.get_command_graph();
+
+		test_utils::mock_buffer_factory mbf(ctx);
+		cl::sycl::range<3> range(16, 1, 1);
+		auto buff = mbf.create_buffer(range);
+		const std::string name{"my_buff_name"};
+		ggen.set_buffer_debug_name(0, name);
```
fknorr, line 807: Replace `0` with `buff.get_id()` here.

test/graph_generation_tests.cc
```diff
@@ -792,5 +792,30 @@ namespace detail {
...
+
+		auto tid = test_utils::build_and_flush(ctx, num_nodes,
+		    test_utils::add_compute_task<class UKN()>(
+		        tm, [&](handler& cgh) { buff.get_access<access_mode::discard_write>(cgh, one_to_one{}); }, range));
+
+		CHECK(graph_generator_testspy::get_buffer_debug_name(ggen, 0) == name);
```
fknorr, line 813: Replace `0` with `buff.get_id()` here.

fknorr (approved):
The tests have some variables that can be const. Also, `mock_buffer` has a method `get_id`, so we don't have to hard-code the expected buffer id.

Otherwise, LGTM!

include/debug.h
```diff
@@ -0,0 +1,17 @@
+#include <string>
+
+#include "buffer.h"
+
+namespace celerity {
+namespace debug {
+	template <typename DataT, int Dims>
+	void set_buffer_name(const celerity::buffer<DataT, Dims>& buff, const std::string& debug_name) {
+		detail::runtime::get_instance().set_buffer_debug_name(detail::get_buffer_id(buff), debug_name);
```
psalz, line 9: Why is there a dedicated function for setting the debug name, but not for getting it? In any case, I'm in favor of just using `get_buffer_manager`.

src/print_graph.cc
```diff
@@ -34,7 +34,19 @@ namespace detail {
 		}
 	}
 
-	void format_requirements(std::string& label, const task& tsk, subrange<3> execution_range, access_mode reduction_init_mode, const reduction_manager& rm) {
+	std::string get_buffer_debug_name(const buffer_manager* bm, const buffer_id bid) {
```
psalz, line 37:
```suggestion
	std::string get_buffer_label(const buffer_manager* bm, const buffer_id bid) {
```

src/print_graph.cc
```diff
@@ -43,15 +55,17 @@ namespace detail {
 
 			const auto bid = reduction.output_buffer_id;
 			const auto req = GridRegion<3>{{1, 1, 1}};
-			fmt::format_to(std::back_inserter(label), "<br/>(R{}) <i>{}</i> B{} {}", rid, detail::access::mode_traits::name(rmode), bid, req);
+			const std::string bdn = get_buffer_debug_name(bm, bid);
```
psalz, line 58:
```suggestion
			const std::string bl = get_buffer_label(bm, bid);
```

test/buffer_manager_tests.cc
psalz, line 1101:
```suggestion
	TEST_CASE_METHOD(test_utils::runtime_fixture, "buffer_manager allows to set debug names on buffers", "[buffer_manager]") {
```

I'm wondering whether this test should be using the buffer-manager APIs directly instead, given that its in "buffer_manager_tests" (right now it's more of an integration test).

test/print_graph_tests.cc
```diff
@@ -81,16 +85,46 @@ TEST_CASE("command graph printing is unchanged", "[print_graph][command-graph]")
...
+	CHECK(dot == expected);
+}
+
+TEST_CASE_METHOD(test_utils::runtime_fixture, "Buffer debug names show up in the generated graph", "[print_graph][buffer_names]") {
+	distr_queue q;
+	celerity::range<1> range(16);
+	celerity::buffer<int, 1> buff_a(range);
+	std::string buff_name{"my_buffer"};
+	celerity::debug::set_buffer_name(buff_a, buff_name);
+	CHECK(celerity::debug::get_buffer_name(buff_a) == buff_name);
```
psalz, line 108: Right now this is kind of redundant with the test in `buffer_manager_tests.cc`. However if you use the APIs on `buffer_manager` directly in the latter as suggested above, it would make sense to keep this, imo.

test/test_utils.h
```diff
@@ -46,6 +46,19 @@
...
+	struct runtime_testspy {
+		static scheduler& get_schdlr(runtime& rt) { return *rt.m_schdlr; }
+		static executor& get_exec(runtime& rt) { return *rt.m_exec; }
+		static size_t get_command_count(runtime& rt) { return rt.m_cdag->command_count(); }
+		static command_graph& get_cdag(runtime& rt) { return *rt.m_cdag; }
+		static std::string print_graph(runtime& rt) {
+			if(auto graph =
+			        rt.m_cdag.get()->print_graph(rt.m_cfg.get()->get_graph_print_max_verts(), *rt.m_task_mngr, *rt.m_reduction_mngr, rt.m_buffer_mngr.get()))
+				return graph.value();
+			return {};
```
psalz, line 58:
```suggestion
			if(auto graph =
			        rt.m_cdag.get()->print_graph(rt.m_cfg.get()->get_graph_print_max_verts(), *rt.m_task_mngr, *rt.m_reduction_mngr, rt.m_buffer_mngr.get())) {
				return graph.value();
			}
			return {};
```

I wonder why clang-tidy didn't catch this 🤔.

psalz: Alternatively you could just `return graph.value_or("");`.

psalz (changes requested):
Thanks, I've added a few minor notes.

Also please update the commit message to properly explain what this is doing.

src/print_graph.cc
```diff
@@ -108,21 +122,23 @@ namespace detail {
...
 			const auto req = GridRegion<3>{{1, 1, 1}};
-			fmt::format_to(std::back_inserter(label), "<b>reduction</b> R{}<br/>B{} {}", rrcmd->get_rid(), reduction.output_buffer_id, req);
+			const std::string bl = get_buffer_label(bm, reduction.output_buffer_id);
+			fmt::format_to(std::back_inserter(label), "<b>reduction</b> R{}<br/> {} {}", rrcmd->get_rid(), bl, req);
 		} else if(const auto hcmd = dynamic_cast<const horizon_command*>(&cmd)) {
 			label += "<b>horizon</b>";
 		} else {
 			assert(!"Unkown command");
-			label += "<b>unknown</b>";
+			label += "<b>unknown<b>";
```
fknorr, line 141: That doesn't seem right - maybe a search/replace error?

src/print_graph.cc
```diff
@@ -34,7 +34,19 @@ namespace detail {
...
-	void format_requirements(std::string& label, const task& tsk, subrange<3> execution_range, access_mode reduction_init_mode, const reduction_manager& rm) {
+	std::string get_buffer_label(const buffer_manager* bm, const buffer_id bid) {
+		// if there is no buffer manager or no name defined, the name will be the buffer id.
+		// if there is a name we want "id name"
+		std::string name{};
+		const std::string bid_str = fmt::format("B{}", bid);
+		if(bm != nullptr) { name = bm->get_debug_name(bid); }
+		if(!name.empty()) { name = fmt::format(" \"{}\"", name); }
+		name = fmt::format("{}{}", bid_str, name);
+		return name;
```
fknorr, line 45: I find the data flow around `name` a bit hard to follow. Suggestion:

```suggestion
		std::string name;
		if(bm != nullptr) { name = bm->get_debug_name(bid); /* empty if no name is set */ }
		return !name.empty() ? fmt::format("B{} \"{}\"", bid, name) : fmt::format("B{}", bid);
```

psalz:
A thought I just had: Right now the debug names are only really helpful when printing the graph, not when debugging a program itself (e.g. with GDB). This is because the names are not stored in the buffers themselves, but somewhere deep inside the runtime. We could however store a pointer to a `std::string` inside the buffer, which is returned by the buffer manager upon registration. What do you think?

src/print_graph.cc
psalz: This changes the output to always include quotes (even if the name is empty). That'd be fine with me though. Maybe even initialize it with `std::string name = "(unnamed)"`?

psalz: Nevermind the first part, didn't see the ternary operator.

psalz (approved):

PR #133, opened 2022-08-18
Remove abstract_command::debug_label

fknorr:
Currently, the DOT graph generation for `print_graph` happens partially in `graph_generator`, which stringifies the data requirements into the `abstract_command::debug_label`. In addition to the missing separation of concerns, this is surprisingly slow as it uses iostreams to print the `GridRegion` of each requirement.

This PR moves this entire printing logic to `print_graph.cc` and removes the `debug_label` member. It also simplifies the remaining printing code by replacing iostreams uses with fmt.


src/print_graph.cc
```diff
@@ -37,120 +35,136 @@ namespace detail {
...
-			ss << "[";
-			for(const auto& it : props) {
-				ss << " " << it.first << "=" << it.second;
-			}
-			ss << "];";
-
+			fmt::format_to(dot_append, "{}[label=\"{}\"];", tsk->get_id(), get_task_label(tsk));
 			for(auto d : tsk->get_dependencies()) {
-				ss << fmt::format("{} -> {} [{}];", d.node->get_id(), tsk->get_id(), dependency_style(d));
+				fmt::format_to(dot_append, "{}->{}[{}];", d.node->get_id(), tsk->get_id(), dependency_style(d));
```
PeterTh, line 44: Intentional output formatting change?
(Probably not too relevant, reading the dot files manually is not a common use case)

src/print_graph.cc
```diff
@@ -37,120 +35,136 @@ namespace detail {
...
 	}
 
-	std::string get_command_label(const abstract_command* cmd) {
-		std::string label = fmt::format("[{}] Node {}:\\n", cmd->get_cid(), cmd->get_nid());
-		if(const auto ecmd = dynamic_cast<const epoch_command*>(cmd)) {
+	std::string get_command_label(const abstract_command& cmd, const task_manager& tm, const reduction_manager& rm) {
+		const command_id cid = cmd.get_cid();
+		const node_id nid = cmd.get_nid();
+
+		std::string label = fmt::format("[{}] Node {}:\\n", cid, nid);
```
PeterTh, line 56: Maybe we should use this chance to make these prefixes a bit shorter (something like `{} N{}\\n`).

PeterTh (changes requested):
Given that there are formatting changes but no changes to any test, I assume we don't have any. Maybe it would be good to add at least a simple basic graph printing unit test (ideally in a commit before the change).

src/print_graph.cc
fknorr: The dot file is not for human consumption, so there's no reason to keep this whitespace.

src/print_graph.cc
fknorr: Done.

fknorr:
I have used the opportunity to improve readability and information density of the graphs overall:
TaskGraph
CommandGraph

Also, there are now smoke tests checking that task and command graphs remains unchanged for two example programs.

psalz (approved):
This is great! The graphs look much nicer now.

>  In addition to the missing separation of concerns, this is surprisingly slow as it uses iostreams to print the GridRegion of each requirement.

Have you tried running the microbenchmarks to see whether this is measurable?

src/print_graph.cc
```diff
@@ -24,133 +22,197 @@ namespace detail {
...
+				for(const auto mode : bam.get_access_modes(bid)) {
+					const auto req = bam.get_requirements_for_access(bid, mode, tsk.get_dimensions(), execution_range, tsk.get_global_size());
+					// While uncommon, we do support chunks that don't require access to a particular buffer at all.
+					if(!req.empty()) { fmt::format_to(std::back_inserter(label), "<br/><i>{}</i> B{} {}", detail::access::mode_traits::name(mode), bid, req); }
+				}
 			}
-			out << "];";
+
+			for(const auto& [hoid, order] : tsk.get_side_effect_map()) {
+				fmt::format_to(std::back_inserter(label), "<br/><i>affect</i> H{}", hoid);
```
psalz, line 151: Not sure if worthwhile, but there seems to be quite a bit of code duplication with `get_task_label`.

fknorr: You're right! I've deduplicated this.

fknorr:
Here's a comparison for CDAG generation benchmarks between master and this branch:

output1

output2

output3

This suggests a significant overall performance improvement, especially in the 4-node case - although it is curious why the same difference can't be seen for the 16-node case.

PeterTh:
LGTM. Re:16 threads - my guess would be that the improvement from getting rid of this string processing only scales linearly with the number of nodes, while many other overheads scale non-linearly, so the relative impact is reduced at increasing node counts.

PeterTh (approved):

PR #134, opened 2022-08-19
Introduce nightly builds to test against HEAD revisions

psalz:
This introduces a nightly workflow schedule for testing against the newest ("HEAD") revisions of DPC++ and hipSYCL. Only if these runs are successful, those versions are greenlit for normal CI testing, by tagging them as "latest".

This approach avoids situations where PRs fail CI because of unrelated upstream changes.

Additionally, this adds an input setting for the `workflow_dispatch` trigger (i.e., when manually launching a run through the GitHub UI) that determines whether to run against the HEAD revisions, normally only tested during nightly builds:

image

This is useful when fixing compatibility with HEAD revisions, so that they can be tagged as "latest". A second new setting determines whether the images should be tagged as well.

Note that the whole thing is a huge hack (due to limitations on GitHub's workflow syntax). But it works!

### DIFF
--- a/.github/workflows/celerity_ci.yml
+++ b/.github/workflows/celerity_ci.yml
@@ -96,56 +96,35 @@ jobs:
   # For normal CI runs we want to build and test against everything except the "HEAD" revisions, whereas during
   # nightly builds we *only* want those.
   #
-  # Current workaround is to build the matrix as a JSON object which is then deserialized in the next job.
-  construct-build-matrix:
+  # Current workaround is to represent the matrix as a JSON object, which is then deserialized in the next job.
+  read-build-matrix:
     needs: find-duplicate-workflows
     if: needs.find-duplicate-workflows.outputs.should_skip != 'true'
     runs-on: self-hosted
     outputs:
-      matrix: ${{ steps.finalize-json-matrix.outputs.matrix }}
+      matrix: ${{ steps.read-json-matrix.outputs.matrix }}
     steps:
-      # Make sure to keep this in sync with docs/platform-support.md.
-      # Also ensure that the "ubuntu-version" between "HEAD" and "latest" builds remains the same.
-      - if: github.event_name != 'schedule' && inputs.test-head == false
+      - uses: actions/checkout@v2
+      - id: read-json-matrix
+        name: Read build matrix from file
+        shell: python
         run: |
-          echo 'json_matrix<<EOF' >> $GITHUB_ENV
-          cat <<EOF >> $GITHUB_ENV
-            { "sycl": "computecpp", "sycl-version": "2.6.0", "ubuntu-version": "20.04", "platform": "intel", "build-type": "Debug" },\
-            { "sycl": "computecpp", "sycl-version": "2.7.0", "ubuntu-version": "20.04", "platform": "intel", "build-type": "Debug" },\
-            { "sycl": "computecpp", "sycl-version": "2.8.0", "ubuntu-version": "20.04", "platform": "intel", "build-type": "Debug" },\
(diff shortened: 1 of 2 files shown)


fknorr (commented):
This has to have been a horrific adventure. Thanks for looking into it!

.github/workflows/celerity_ci.yml
```diff
@@ -37,7 +54,7 @@ jobs:
     if: github.event.pull_request
     runs-on: [self-hosted, nvidia]
     env:
-      container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
+      container-workspace: <placeholder>
       build-dir: /root/build
     container:
```
fknorr, line 59: This container should also point to `:latest`.

.github/workflows/celerity_ci.yml
```diff
@@ -70,96 +92,80 @@ jobs:
...
+    runs-on: self-hosted
+    outputs:
+      matrix: ${{ steps.finalize-json-matrix.outputs.matrix }}
+    steps:
+      # Make sure to keep this in sync with docs/platform-support.md.
+      # Also ensure that the "ubuntu-version" between "HEAD" and "latest" builds remains the same.
+      - if: github.event_name != 'schedule' && inputs.test-head == false
+        run: |
+          echo 'json_matrix<<EOF' >> $GITHUB_ENV
+          cat <<EOF >> $GITHUB_ENV
```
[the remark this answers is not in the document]
psalz: Good idea, done!

.github/workflows/celerity_ci.yml
psalz: Fixed!

.github/workflows/celerity_ci.yml
```diff
@@ -197,16 +203,59 @@ jobs:
...
+        include:
+          - sycl: "dpcpp"
+            platform: "intel"
+          - sycl: "hipsycl"
+            platform: "nvidia"
+    env:
+      image-basename-dpcpp: celerity-build/dpcpp:ubuntu${{ needs.build-and-test.outputs.dpcpp-HEAD-ubuntu-version }}
+      image-basename-hipsycl: celerity-build/hipsycl:ubuntu${{ needs.build-and-test.outputs.hipsycl-HEAD-ubuntu-version }}
+    steps:
+      - if: matrix.sycl == 'dpcpp'
```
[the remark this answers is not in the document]
psalz: Unfortunately no, because the workflow syntax doesn't support nested expressions/substitutions.

.github/workflows/celerity_ci.yml
```diff
@@ -70,96 +92,58 @@ jobs:
...
+        name: Read build matrix from file
+        shell: python
+        run: |
+          import json
+          with open("${{ github.workspace }}/.github/workflows/build_matrix.json") as f:
+            matrices = json.load(f)
+            true_false = { 'true': True, 'false': False } # Hack for getting YAML booleans into python
+            is_regular_run = true_false['${{ github.event_name != 'schedule' && inputs.test-head == false }}']
+            default_matrix = json.dumps(matrices['default' if is_regular_run else 'nightly'])
+            print('::set-output name=matrix::{ "include":%s }' % default_matrix)
```
fknorr, line 118: This script could be a bit less convoluted:

```suggestion
          if '${{ github.event_name != 'schedule' && inputs.test-head == false }}' == 'true':
            matrix = matrices['default']
          else:
            matrix = matrices['nightly']
          endif
          print('::set-output name=matrix::{ "include":%s }' % json.dumps(matrix))
```

fknorr (approved):
One small stylistic comment on the Python script, otherwise LGTM!

.github/workflows/celerity_ci.yml
psalz: Thanks, done!

PR #135, opened 2022-08-22
Schedule all CI jobs through SLURM

fknorr:
Since #126 the `build-and-test` jobs are scheduled through SLURM, but linting and "meta-jobs" can still be assigned to non-slurm runners. This PR constrains all jobs to the SLURM runners.


psalz (approved):
LGTM!

PR #136, opened 2022-08-24
Run Celerity microbenchmarks via CI

fknorr:
This adds a new CI workflow that runs the `benchmarks` binary and creates `gpuc2_bench.*` as artifacts. slurmactiond is configured such that the corresponding SLURM job uses `--exclusive --cpu-freq=high --nodelist=gpuc2` for optimal results.

While attempting to run the benchmarks inside Docker for the first time, it became apparent that our `benchmark_reporters` rely on system-specific locales for number formatting which was not available in our Docker images. For robustness, I have replaced this use of locales with a hand-rolled loop that inserts the thousands separators we need.

**Note:** GitHub does not allow us to trigger `workflow_dispatch` for workflows that do not exist on the default branch. As a workaround I've added a `push` trigger to the new workflow and removed the `push` trigger for the existing CI - I will drop these commits before merging.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,36 @@
+name: Celerity Benchmarks
+
+on:
+  workflow_dispatch:
+
+jobs:
+  benchmark:
+    runs-on: [ self-hosted, "slurm-nvidia-benchmark" ]
+    env:
+      container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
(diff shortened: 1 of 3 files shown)


psalz (approved):
Looks like results aren't being uploaded for some reason. Otherwise LGTM!

You *could* also pipe the CSV report into `$GITHUB_STEP_SUMMARY` to [get a nice summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/).

PR #137, opened 2022-08-25
Remove task mutex

fknorr:
`task_manager` had a `m_task_mutex` left over from #112 that is now superfluous.

It was used in five places, all of which are only ever called from the main thread:
- task_manager::submit_command_group()
- task_manager::add_buffer()
- task_manager::generate_epoch_task()
- task_manager::generate_horizon_task()
- task_manager::print_graph()

All concurrency the task manager is involved in now is handled by the epoch monitor.

### DIFF
--- a/include/task_manager.h
+++ b/include/task_manager.h
@@ -68,7 +68,6 @@ namespace detail {
 			task_id tid;
 			const task* tsk_ptr = nullptr;
 			{
-				std::lock_guard lock(m_task_mutex);
 				auto reservation = m_task_buffer.reserve_task_entry(await_free_task_slot_callback());
 				tid = reservation.get_tid();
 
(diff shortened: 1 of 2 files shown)


psalz (approved):

include/task_manager.h
```diff
@@ -68,7 +68,6 @@ namespace detail {
 			task_id tid;
 			const task* tsk_ptr = nullptr;
 			{
-				std::lock_guard lock(m_task_mutex);
```
PeterTh, line 71: If you get rid of the lock here, you can also get rid of the surrounding compound statement block (which was only introduced for the lock I believe), and then you can get rid of the "forward definition" of `tid` and `tsk_ptr`, which is nice.

src/task_manager.cc
```diff
@@ -232,7 +230,6 @@ namespace detail {
 		task* new_horizon;
 		{
 			auto reserve = m_task_buffer.reserve_task_entry(await_free_task_slot_callback());
-			std::lock_guard lock(m_task_mutex);
```
PeterTh, line 235: Same thing regarding the extra compound statement + superfluous early definition applies here, and also in `generate_epoch_task`.

PeterTh (changes requested):

include/task_manager.h
fknorr: Good catch! There were also some outdated comments related to locking, removed those as well.

src/task_manager.cc
fknorr: Cleaned up.

PeterTh (approved):
LGTM.

PR #138, opened 2022-08-25
Communicate buffer creation to graph generator through scheduler event

fknorr:
Currently the graph generator holds a `m_buffer_mutex` for the sole purpose of synchronizing buffer addition between main thread and scheduler thread. This is contrary to the otherwise thread-unaware design of the graph generator and could cause scheduler stalls if buffers are constructed while tasks are scheduled.

This PR introduces a new scheduler event for buffer addition which makes this mutex unnecessary. It also transfers ownership of graph generator and graph serializer to the scheduler in order to prevent accidental concurrent access from a non-scheduler thread.

`test_utils::mock_buffer_factory` also gets a richer constructor interface to be able to deal with either a non-threaded graph generator or a threaded scheduler.

### DIFF
--- a/include/scheduler.h
+++ b/include/scheduler.h
@@ -20,7 +20,7 @@ namespace detail {
 	// Abstract base class to allow different threading implementation in tests
 	class abstract_scheduler {
 	  public:
-		abstract_scheduler(graph_generator& ggen, graph_serializer& gsrlzr, size_t num_nodes);
+		abstract_scheduler(std::unique_ptr<graph_generator> ggen, std::unique_ptr<graph_serializer> gsrlzr, size_t num_nodes);
 
 		virtual ~abstract_scheduler() = default;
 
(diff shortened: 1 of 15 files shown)


psalz (approved):
Neat, LGTM!

include/scheduler.h
```diff
@@ -42,18 +42,28 @@ namespace detail {
...
+			const task* tsk;
+		};
+		struct event_buffer_registered {
+			buffer_id bid;
+			celerity::range<3> range;
+		};
+		using event = std::variant<event_shutdown, event_task_available, event_buffer_registered>;
+
+		std::unique_ptr<graph_generator> m_ggen;
+		std::unique_ptr<graph_serializer> m_gsrlzr;
```
BlackMark29A, line 56: The IMHO very terrible name of `gsrlzr` could have benefitted from some drive-by-renaming.

BlackMark29A (approved):
Looks good to me, I just have one minor remark:

include/scheduler.h
fknorr: I agree. I've renamed to `gser` now, since all "sibling variables" are abbreviated in a similar manner.

IMO for new classes we should confine this kind of abbreviation to local variables and focus on readability for members.

PR #139, opened 2022-08-31
Refactor / clean up a few random things

psalz:
⬆

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <functional>
 #include <type_traits>
 #include <variant>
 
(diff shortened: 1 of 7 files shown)


include/utils.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <functional>
```
fknorr, line 4: How on earth did `std::hash` even end up in `<functional>`...

fknorr (approved):
This is missing a namespace qualification in `graph_generation_tests.cc:572` and `:557` (see CI failure).

Otherwise, LGTM.

psalz:
> This is missing a namespace qualification in `graph_generation_tests.cc:572` and `:557` (see CI failure).

...aw man, that's this annoying NFS issue again, where git doesn't pick up on all changes I've made through VSCode via SSH. Need to look into that.

PeterTh (approved):

PR #142, opened 2022-09-19
Coalesce command serialization into single MPI op per task and node

fknorr:
In addition to execution commands, scheduling a task will generate push / await-push / reduction cascades on participating nodes. So far, all of these implicit commands are `MPI_Sent` to workers individually, initiating _O(N²)_ MPI ops in case of an all-to-all communication pattern. As we have investigated previously, this will overwhelm the management of asynchronous ops at least in OpenMPI and degrade network performance.

### Frame Vectors

This PR introduces `frame_vector`, a contiguous data structure for serializing multiple `command_frame`s as they exist today. The memory layout is as follows:
```
[size_t num_frames][size_t frame_0_size_bytes][size_t frame_k_size_bytes...][frame_0][frame_k...]
```
A `frame_vector` can be allocated in two different manners:
- By constructing it `from_size_bytes` and filling it with a byte-wise copy (aka, `MPI_Recv`)
- By assembling it from scratch:
    1. Create a `frame_vector_layout`, calling `reserve_back` to compute the required allocation size
    2. Create ` frame_vector_builder` from the `layout` and constructing frames in-place by calling `emplace_back`
    3. Obtaining the final `frame_vector` from the `builder` by calling `into_vector`

This three-step program appears somewhat convoluted, but is easier to implement than an auto-resizing vector and also allows us to build the entire structure with a single call to `malloc`.

### Shared Frame Pointers

Since we are packing together commands only based on their task-id and destination node-id but are otherwise unrelated from the perspective of the Executor, we need to entertain the question how the frame memory should be freed.

The most elegant solution I came up with so far is to introduce a `shared_frame_ptr` akin to `unique_frame_ptr`, and create _N_ `shared_frame_ptr`s in the executor that collectively own the allocation of the receiving `frame_vector` that contained _N_ commands.

### Good Bye Virtual Dependencies

The `graph_serializer` has almost been re-written entirely to move from a _flush-as-you-go_ model to a _collect-and-flush-once_ aproach. This has the added benefit of requiring only around _2N_ allocations for _N_ nodes.

So far, we had the concept of a _virtual depencency_, aka an in-CDAG dependency to a command that exists implicitly and has not actually been flushed to its worker. This only applies to the initial epoch, which is created implicitly in TDAG and CDAG, and parts of the code (especially tests) weren't able to deal with a dependency to a command they did not receive. This _virtual dependency_ special casing became a nuisance when trying to compute command payload sizes in advance.

As a solution, this PR moves to serialize dependencies to init-epoch commands like any other dependency, and equips the (test) code in question to deal with those implicit commands. Technically this means we are inflating each affected command frame by one `size_t`, but I would argue the performance impact of this will not be measurable.

### What is Left to Figure Out

- [ ] This PR adds even more _frame_ management types than we already have. We could simplify this, e.g. by
   - getting rid of `unique_frame_ptr` and only ever transmitting `frame_vector`s (which would then have size 1 for `data_frames` and waste one `size_t` of bandwidth)
   - making shared-pointer semantics the default and getting rid of `unique_*_ptr` types
- [ ] Benchmark this PR against master to figure out if we have addressed the OpenMPI bottleneck
- [ ] Bikeshed terminology - does `frame_vector` make sense, or is the vector the actual _frame_ (because its being transmitted over the network) and its elements should have a different name?
- [x] ~~This PR seems to introduce a bug around reductions that is sometimes observable on DPC++ (see the [most recent CI failure](https://github.com/celerity/celerity-runtime/actions/runs/3081442005/jobs/4979998874)). See if this is caused by the modified transitive-dependency handling logic in `graph_serializer`, or if we depend on the order of commands being sent in an incorrect manner.~~ Was unrelated to this PR after all, and is fixed by #146.

### DIFF
--- a/src/graph_serializer.cc
+++ b/src/graph_serializer.cc
@@ -6,114 +6,118 @@
 #include "command_graph.h"
 #include "task_manager.h"
 
-namespace celerity {
-namespace detail {
+namespace celerity::detail {
 
-	void graph_serializer::flush(task_id tid) { flush(m_cdag.task_commands(tid)); }
+graph_serializer::graph_serializer(const size_t num_nodes, command_graph& cdag, flush_callback flush_cb)
+    : m_num_nodes(num_nodes), m_cdag(cdag), m_flush_cb(std::move(flush_cb)) {}
 
-	bool is_virtual_dependency(const abstract_command* const cmd) {
-		// The initial epoch command is not flushed, so including it in dependencies is not useful
-		// TODO we might want to generate and flush init tasks explicitly to avoid this kind of special casing
-		const auto ecmd = dynamic_cast<const epoch_command*>(cmd);
-		return ecmd && ecmd->get_tid() == task_manager::initial_epoch_task;
+void graph_serializer::flush(const task_id tid) {
+	node_command_map pending_node_cmds(m_num_nodes);
+	for(const auto tcmd : m_cdag.task_commands(tid)) {
+		assert(tcmd->get_tid() == tid);
+		collect_task_command(tcmd, pending_node_cmds);
 	}
+	for(node_id nid = 0; nid < m_num_nodes; ++nid) {
+		if(!pending_node_cmds[nid].empty()) { m_flush_cb(nid, serialize(pending_node_cmds[nid])); }
+	}
+}
 
-	void graph_serializer::flush(const std::vector<task_command*>& cmds) {
+void graph_serializer::collect_task_command(task_command* const cmd, node_command_map& pending_node_cmds) const {
+	cmd->mark_as_flushed();
+
+	const auto nid = cmd->get_nid();
+	assert(nid <= m_num_nodes);
+
+	// Iterate over first level of dependencies.
+	// These might either be commands from other tasks that have been flushed previously or generated data transfer / reduction commands.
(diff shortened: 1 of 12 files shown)


src/graph_serializer.cc
```diff
@@ -6,114 +6,118 @@
...
-		// Special casing for await_push commands: Also flush the corresponding push.
-		// This is necessary as we would otherwise not reach it when starting from task commands alone
-		// (unless there exists an anti-dependency, which is not true in most cases).
-		if(isa<await_push_command>(dep)) {
-			const auto pcmd = static_cast<await_push_command*>(dep)->get_source();
-			if(!pcmd->is_flushed()) flush_dependency(pcmd);
-		}
+	// Iterate over second level of dependencies. These will usually be flushed already. One notable exception are reduction dependencies, which generate
+	// a tree of push_await_commands and reduction_commands as a dependency.
+	// TODO: We could probably do some pruning here (e.g. omit tasks we know are already finished)
```
fknorr, line 69: Is this comment still relevant? It reads like a pre-horizons consideration.

PeterTh:
This will be made obsolete by distributed command generation in the future, which obviates the need for sending out commands.

PR #143, opened 2022-09-19
Remove reduction_manager from print_graph and test code

fknorr:
So far, `reduction_manager` is used both for rid -> bid mapping and for propagating reduction functors to the executor. This draws the class into `print_graph` and all non-runtime test code. It also lead to a bug being introduced to the runtime-shutdown graph printing as part of #133, where attempting to resolve the buffer id behind a reduction that had already finished executing would throw and abort in `~distr_queue`.

This PR replaces plain `reduction_id`s with the new `reduction_info` struct in TDAG and CDAG, replicating the cheap rid -> bid mapping in all places where it is needed. This allows us to move `reduction_manager` to the `runtime`-level of abstractions, keeping it out of all non-RT tests and disentangling quite a lot of related code.

In tests, reduction_manager is now replaced by the new `mock_reduction_factory` which, similar to `mock_buffer_factory` is mostly concerned with assigning reduction ids. Also, a smoke test `distr_test` that failed with graph printing enabled was added.


test/accessor_tests.cc
```diff
@@ -216,31 +216,32 @@ namespace detail {
 		auto& tm = runtime::get_instance().get_task_manager();
 		auto& rm = runtime::get_instance().get_reduction_manager();
```
psalz, line 217: Hmm... This is unused but doesn't seem to produce a warning? Weird.

test/system/distr_tests.cc
```diff
@@ -135,6 +135,24 @@ namespace detail {
 		});
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "runtime-shutdown graph printing does not break when encountering a finished reduction",
```
psalz, line 138: We won't notice if this test ever becomes obsolete when/if we change the way graph printing is handled (e.g. by not tying it directly to log level). You could assert that the graphs are being printed, see `log_capture` in device_selection_tests.cc.

psalz (approved):
Very nice!

test/accessor_tests.cc
fknorr: Weird indeed... removed it nonetheless.

test/system/distr_tests.cc
fknorr: Good point! I changed that test to use `log_capture` and prepended a commit that guarantees stable log captures independent of the `CELERITY_LOG_LEVEL` setting (this broke the `device_selection_tests` before when the global log level was set below `info`).

psalz (approved):

PR #145, opened 2022-09-20
Do not fmt::format log strings unless required

fknorr:
We have a two-stage log formatting scheme where we pass the user's formatting spec into `fmt::format` and then use our own format string `"{}{}"` to prepend the Celerity log context. We currently call `fmt::format` unconditionally, even if spdlog will drop the message due to the current global log level.

This PR makes the calls to fmt and spdlog conditional on the global log level.

### DIFF
--- a/include/log.h
+++ b/include/log.h
@@ -6,23 +6,25 @@
 #include <utility>
 #include <variant>
 
-// TODO: Make this configurable through CMake?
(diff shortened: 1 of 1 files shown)


BlackMark29A (approved):
LGTM.

psalz (approved):
Good change, thanks!

PR #146, opened 2022-09-21
Fix: Generate intra-task anti-dependencies for reduction pushes

fknorr:
`graph_generator` must generate anti-dependencies between push- and execution commands arising from the same task (known as _intra-task race conditions_ in the code). This works as intended when the push is generated from a buffer in `distributed_state` (the common case), but not when that buffer is in `pending_reduction_state` following a kernel that wrote (partial) reduction results.

missing-anti-dep

This PR fixes the bug by considering pushes arising from reductions for intra-task dependency analysis.

The bug surfaced in `distr_tests` (_multiple chained reductions produce correct results_) when testing #142 , although it is entirely unrelated to the contents of that PR.


psalz (approved):
Now that's what I call an *edge case* 😎.

PR #147, opened 2022-09-22
Mark intrusive_graph_node::has_[dependency|dependent] as const

psalz:
10chars

### DIFF
--- a/include/intrusive_graph.h
+++ b/include/intrusive_graph.h
@@ -54,24 +54,20 @@ namespace detail {
 		using dependent = dependency;
 
 	  public:
-		intrusive_graph_node() { static_assert(std::is_base_of<intrusive_graph_node<T>, T>::value, "T must be child class (CRTP)"); }
+		intrusive_graph_node() { static_assert(std::is_base_of_v<intrusive_graph_node<T>, T>, "T must be child class (CRTP)"); }
 
 	  protected:
 		~intrusive_graph_node() { // protected: Statically disallow destruction through base pointer, since dtor is not polymorphic
 			for(auto& dep : m_dependents) {
-				auto dep_it_end = static_cast<intrusive_graph_node*>(dep.node)->m_dependencies.end();
-				auto this_it =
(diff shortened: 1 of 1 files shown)


include/intrusive_graph.h
```diff
@@ -151,6 +151,13 @@ namespace detail {
 			if(it == deps.end()) return std::nullopt;
 			return it;
 		}
+
+		template <typename Dep>
+		std::optional<typename gch::small_vector<Dep>::const_iterator> maybe_get_dep(const gch::small_vector<Dep>& deps, const T* node) const {
```
fknorr, line 156: That's essentially a static method, so there should be no need for const/non-const versions:

```c++
		template <typename Dep>
		static auto maybe_get_dep(gch::small_vector<Dep>& deps, T* node) {
			auto it = std::find_if(deps.begin(), deps.end(), [&](auto d) { return d.node == node; });
			if(it == deps.end()) return std::nullopt; // not sure if this typechecks
			return std::optional{it};
		}
```

fknorr (commented):
The `const` qualifiers LGTM of course, but I can't help but scratch my head at the entire `maybe_get_dep` function. Why have an `optional<iterator>` when iterators are already nullable through `range.end()`?

If I were to design this from scratch it would look like so:

```c++
public:
    bool has_dependency(T* node, std::optional<dependency_kind> kind = std::nullopt) const {
        if (auto it = find_by_node(m_dependencies, node); it != m_dependencies.end()) {
            ...
        }
    }

private:
    template<typename Range>
    static auto find_by_node(Range &rng, T *const node) {
        using std::begin, std::end;
        return std::find_if(begin(rng), end(rng), [=](auto &d) { return d.node == node });
    }
```

Maybe this PR is the right place for a small refactoring.

psalz:
You are absolutely right! Refactored.

fknorr (approved):
LGTM, although there are a bunch of variables that could be const (multiple iterators `it` / `this_it`, and function parameters to `has_dependency` / `has_dependent`).

psalz:
Rebased onto current master, made everything const.

PR #148, opened 2022-09-26
CI: Disable buggy skipping of duplicate runs

fknorr:
Due to [a bug](https://github.com/fkirc/skip-duplicate-actions/issues/264), the fkirc/skip-duplicate-actions incorrectly identifies `pull_request` as duplicates to `push` triggers based on identical base branches, even though pull requests run will on an implicit merge commit.

This temporarily disables skipping of concurrent runs. We keep cancellation of known-outdated runs when a newer commit was pushed on the same branch.


psalz (approved):
Sorry for not looking at this earlier, LGTM

PR #151, opened 2022-11-07
Fences as a replacement for Captures

fknorr:
As discussed in #94, the current proposal attaching data extraction facilities to epochs is suboptimal.

- Extracting data on epochs prevents any concurrency between capture data transfers and other tasks
- Generating data transfers for epochs violates the command-id ordering property around epochs as it is implemented currently
- The distinction between captures on `slow_full_sync` and `drain` is subtle and confusing

This PR implements _fences_ in the form of the user-facing `celerity::experimental::fence(q, captures...)` and the new _fence_ task- and command types. Fences synchronize between executor and main threads for data extraction, but will allow independent tasks that have been submitted before to execute while the application is waiting on completion of the fence operation. To avoid serialization around multiple fences, a tuple of captures can be specified in the `fence()` call that will return a tuple of values.

A user can `fence` three different structures:

- A `buffer`, returning an `experimental::buffer_snapshot` which is just a `std::vector` with metadata
- An `experimental::buffer_subrange`, which combines a buffer with a subrange to extract. Likewise returns a `buffer_snapshot`.
- A `host_object`, whose interior needs to be copyable.

## Example

```c++
int main() {
    distr_queue q;
    buffer<int> buf_1;
    //...
    experimental::buffer_snapshot<int, 1> snap = experimental::fence(q,
        buffer_subrange(buf_1, subrange({1}, {1}});
    std::vector<int> data = std::move(snap.into_data());
}
```
```c++
int main() {
    distr_queue q;
    buffer<int> buf_1;
    buffer<float> buf_2;
    experimental::host_object<std::string> obj;
    //...
    auto [ints, floats, string] = experimental::fence(q, std::tuple(buf_1, buf_2, obj));
}
```

## DAGs for fencing a buffer subrange

buffers

## DAGs for fencing a host object

host-objects


fknorr:
I have updated this PR to use asynchronous fences through futures as discussed above.

The new interface is as follows:

```c++
namespace celerity::experimental {

class buffer_snapshot {
    ...
};

template <typename T>
std::future<T> fence(distr_queue&, const host_object<T>& obj);

std::future<buffer_snapshot<DataT, Dims>>
fence(distr_queue&, const buffer<DataT, Dims>& buf, const subrange<Dims>& sr);

std::future<buffer_snapshot<DataT, Dims>>
fence(distr_queue& q, const buffer<DataT, Dims>& buf);

}
```

The DPC++ CI failure is not related to these changes.

examples/matmul/matmul.cc
```diff
@@ -87,9 +87,10 @@ int main() {
...
-	bool verification_passed = true;
-	verify(queue, mat_a_buf, verification_passed);
-	queue.slow_full_sync(); // Wait for verification_passed to become available
+	celerity::buffer<bool> passed_buf(1);
+	verify(queue, mat_c_buf, passed_buf);
 
-	return verification_passed ? EXIT_SUCCESS : EXIT_FAILURE;
+	// The value of `passed` can differ between hosts if only part of the verification failed.
+	const auto passed = celerity::experimental::fence(queue, passed_buf).get();
+	return passed ? EXIT_SUCCESS : EXIT_FAILURE;
```
fknorr, line 95: Bug: I just noticed that this tests `buffer_snapshot::operator bool` instead of the snapshot contents! Looks like a footgun, maybe get rid of those pointer semantics?

src/graph_generator.cc
```diff
@@ -47,35 +47,33 @@ namespace detail {
...
 		case task_type::master_node: generate_independent_execution_commands(tsk); break;
+		case task_type::fence: generate_fence_commands(tsk); break;
 		}
 
 		for(auto& t : transformers) {
 			t->transform_task(tsk, m_cdag);
 		}
 
-		// Only execution tasks can have data requirements or reductions
-		if(tsk.get_execution_target() != execution_target::none) {
```
psalz, line 57: Do you think the overhead of running `process_task_data_requirements` on every horizon is negligible?

include/fence.h
```diff
@@ -0,0 +1,150 @@
...
+namespace celerity::experimental {
+
+/**
+ * Asynchronously captures the value of a host object by copy, introducing the same dependencies as a side-effect would.
+ *
+ * Waiting on the returned future in the application thread can stall scheduling of more work. To hide latency, either submit more command groups between
+ * fence and wait operations or ensure that other independent command groups are eligible to run while the fence is executed.
+ */
+template <typename T>
+std::future<T> fence(celerity::distr_queue&, const experimental::host_object<T>& obj) {
```
psalz, line 113: These should be `[[nodiscard]]`!

psalz, line 113: Creating a fence for a `host_object<T&>` currently doesn't work, and it probably shouldn't. The produced error is a bit obscure though - maybe add a static assertion here to catch this early?

test/graph_generation_tests.cc
```diff
@@ -793,5 +795,77 @@ namespace detail {
...
+		auto& inspector = ctx.get_inspector();
+
+		test_utils::mock_buffer_factory mbf(ctx);
+		auto buf = mbf.create_buffer<1>({1});
+
+		const auto tid_a = test_utils::build_and_flush(
+		    ctx, num_nodes, test_utils::add_host_task(tm, on_master_node, [&](handler& cgh) { buf.get_access<access_mode::discard_write>(cgh, all{}); }));
+		const auto tid_fence = test_utils::build_and_flush(ctx, num_nodes, test_utils::add_fence_task(tm, buf));
+		const auto tid_b = test_utils::build_and_flush(ctx, num_nodes,
+		    test_utils::add_host_task(tm, experimental::collective, [&](handler& cgh) { buf.get_access<access_mode::discard_write>(cgh, all{}); }));
```
psalz, line 844: Using overlapping writes here makes sense for the test but is of course only semi-legal. Maybe add a comment that this might break in the future, if we decide to explicitly forbid overlapping writes?

test/task_graph_tests.cc
```diff
@@ -690,5 +690,37 @@ namespace detail {
...
+		const auto tid_b = test_utils::add_host_task(
+		    tm, celerity::experimental::collective, [&](handler& cgh) { ho.add_side_effect(cgh, experimental::side_effect_order::sequential); });
+
+		CHECK(has_dependency(tm, tid_fence, tid_a));
+		CHECK(has_dependency(tm, tid_b, tid_fence));
+
+		test_utils::maybe_print_graph(tm);
+	}
+
+	TEST_CASE("fences introduce data dependencies and movement", "[task_manager][task-graph][fence]") {
```
psalz, line 710: What do you mean with "data movement" here?

examples/matmul/matmul.cc
```diff
@@ -45,27 +45,27 @@ void multiply(celerity::distr_queue queue, celerity::buffer<T, 2> mat_a, celerit
...
-		celerity::accessor result{mat_a_buf, cgh, celerity::access::one_to_one{}, celerity::read_only_host_task};
+void verify(celerity::distr_queue& queue, celerity::buffer<T, 2> mat_c, celerity::buffer<bool> passed_buf) {
+	queue.submit([=](celerity::handler& cgh) {
+		celerity::accessor c{mat_c, cgh, celerity::access::one_to_one{}, celerity::read_only_host_task};
+		celerity::accessor passed{passed_buf, cgh, celerity::access::all{}, celerity::write_only_host_task, celerity::no_init};
 
-		cgh.host_task(mat_a_buf.get_range(), [=, &verification_passed](celerity::partition<2> part) {
-			auto sr = part.get_subrange();
+		cgh.host_task(mat_c.get_range(), [=](celerity::partition<2> part) {
+			passed[0] = true;
```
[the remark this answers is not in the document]
fknorr: We should!

examples/matmul/matmul.cc
fknorr: Removed. There were quite a few more surface API bugs in the `buffer_snapshot` definition that I fixed as well.

include/fence.h
fknorr: Done.

fknorr: I've added a static assertion.

 `<T&>` and `<void>` are a bit of a stain anyway. We should overhaul host objects at some point.

test/graph_generation_tests.cc
fknorr: I've replaced it with a o2o range mapper - proves the same point.

src/graph_generator.cc
fknorr: No way to know without profiling, but I've added a shortcut branch around it for horizons and epochs.

examples/matmul/matmul.cc
```diff
@@ -87,9 +87,10 @@ int main() {
 	multiply(queue, mat_a_buf, mat_b_buf, mat_c_buf);
 	multiply(queue, mat_b_buf, mat_c_buf, mat_a_buf);
 
-	bool verification_passed = true;
-	verify(queue, mat_a_buf, verification_passed);
-	queue.slow_full_sync(); // Wait for verification_passed to become available
+	celerity::buffer<bool> passed_buf(1);
```
PeterTh, line 90: It might make sense to init this buffer to `false`, and also remove `no_init` in the accessor.  
That way the verification would fail for sure if for whatever reason the kernel were to not run at all.

include/command.h
```diff
@@ -213,7 +223,8 @@ namespace detail {
 			    [](const execution_data&) { return command_type::execution; },
 			    [](const push_data&) { return command_type::push; },
 			    [](const await_push_data&) { return command_type::await_push; },
-			    [](const reduction_data&) { return command_type::reduction; }
+			    [](const reduction_data&) { return command_type::reduction; },
+				[](const fence_data&) { return command_type::fence; }
```
PeterTh, line 227: Formatting broken?

include/fence.h
```diff
@@ -0,0 +1,151 @@
...
+
+	inline const T& operator[](const id<Dims> index) const { return m_data[detail::get_linear_index(m_sr.range, index)]; }
+
+	inline detail::subscript_result_t<Dims, const buffer_snapshot> operator[](const size_t index) const { return detail::subscript<Dims>(*this, index); }
+
+  private:
+	template <typename U, int Dims2>
+	friend class detail::buffer_fence_promise;
+
+	subrange<Dims> m_sr;
```
PeterTh, line 54: Super picky, but I feel like a member variable is sufficiently important/long lived to justify a longer name (i.e. `m_subrange` in this case).

include/ranges.h
```diff
@@ -5,6 +5,54 @@
 namespace celerity {
 namespace detail {
 
+	template <int TargetDims, typename Target, int SubscriptDim = 0>
+	class subscript_proxy;
```
PeterTh, line 9: Do we have any indication to what extent this will or will not affect the code quality produced by kernel compilers?  
(Since we recently saw that they can be exceedingly finicky)

include/task.h
```diff
@@ -23,6 +23,7 @@ namespace detail {
 		collective,     ///< host task with implicit 1d global size = #ranks and fixed split
 		master_node,    ///< zero-dimensional host task
 		horizon,        ///< task horizon
+		fence,
```
PeterTh, line 26: While some of the comments on the task_types above are meaningless, not having one for `epoch` and `fence` offends my sense of propriety.

include/task.h
```diff
@@ -147,37 +159,45 @@ namespace detail {
...
+		static std::unique_ptr<task> make_horizon(task_id tid) {
+			return std::unique_ptr<task>(new task(tid, task_type::horizon, collective_group_id{}, task_geometry{}, nullptr, {}, {}, {}, {}, {}, nullptr));
 		}
 
-		static std::unique_ptr<task> make_horizon_task(task_id tid) {
-			return std::unique_ptr<task>(new task(tid, task_type::horizon, collective_group_id{}, task_geometry{}, nullptr, {}, {}, {}, {}, {}));
+		static std::unique_ptr<task> make_fence(
+		    task_id tid, buffer_access_map access_map, side_effect_map side_effect_map, std::unique_ptr<fence_promise> fence_promise) {
+			return std::unique_ptr<task>(new task(tid, task_type::fence, collective_group_id{}, task_geometry{}, nullptr, std::move(access_map),
+			    std::move(side_effect_map), {}, {}, {}, std::move(fence_promise)));
```
PeterTh, line 200: I am still unhappy about the entire state of things in this general area.  
(Not an immediate call to action in this PR)

examples/matmul/matmul.cc
fknorr: Done.

include/command.h
fknorr: Fixed (just a hassle as always, indenting with tabs :P)

include/fence.h
fknorr: Renamed.

include/ranges.h
fknorr: Would be interesting to have some sort of microbenchmarking there for sure! Also regarding the `id` -> pointer offset computation that is happening within `operator[](id<>)`. I would suggest this for a separate PR though because the benchmark baseline is not an earlier Celerity version but a hand-rolled SYCL / CUDA kernel.

psalz (approved):

include/task.h
fknorr: Commented.

PR #152, opened 2022-11-07
Fix race between creating collective groups and submitting host tasks

fknorr:
Current Celerity has a race condition on worker nodes between submission of collective host tasks and the submission of their kernels to the host queue. Choosing a non-default collective group creates an additional worker thread from the main thread, and kernel submission requires that thread to be up and running.

- The executor busy-waits for the task definition to be available, but the collective group thread was created after task insertion into the ringbuffer. This has been reversed to ensure the worker thread is ready when the task becomes available to the executor.
- Even though the set of worker threads in `host_queue` was accessed by both main and executor thread, it was left unsynchronized. The class member is now protected by a mutex.

### DIFF
--- a/include/host_queue.h
+++ b/include/host_queue.h
@@ -119,7 +119,9 @@ namespace detail {
 		}
 
 		void require_collective_group(collective_group_id cgid) {
-			if(m_threads.find(cgid) != m_threads.end()) return;
+			const std::lock_guard lock(m_mutex); // called by main thread
+			if(m_threads.count(cgid) > 0) return;
+
 			assert(cgid != 0);
 			MPI_Comm comm;
 			MPI_Comm_dup(MPI_COMM_WORLD, &comm);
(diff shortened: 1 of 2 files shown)


include/host_queue.h
```diff
@@ -151,24 +153,26 @@ namespace detail {
 		 * @brief Waits until all currently submitted operations have completed.
 		 */
 		void wait() {
-			for(auto& ct : m_threads) {
-				ct.second.thread.stop(true /* isWait */);
+			const std::lock_guard lock(m_mutex); // called by main thread - never contended because the executor is shut down at this point
+			for(auto& [cgid, cp] : m_threads) {
+				cp.pool.stop(true /* isWait */);
```
psalz, line 158: Why `cp` instead of `ct`?

psalz (approved):
LGTM!

include/host_queue.h
fknorr: Im destructuring the `ct` that we iterated over before, and the right-hand side is **c**gid + **p**ool.

PR #153, opened 2022-11-08
Enable data transfers larger than 2 GiB

fknorr:
Currently, buffer data transfers are limited to 2 GiB by the `MPI_Isend`/`Irecv` API that takes 32-bit signed integer element counts. This PR increases the limit to 128GiB by transferring data as arrays of a 64-byte custom MPI type.

Since the receiver side cannot recover the byte-exact frame size, we also have to drop `unique_frame_ptr::get_payload_count()`, which was fortunately only used in one assertion.

### DIFF
--- a/include/buffer_transfer_manager.h
+++ b/include/buffer_transfer_manager.h
@@ -24,6 +24,8 @@ namespace detail {
 			bool complete = false;
 		};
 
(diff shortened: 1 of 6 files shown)


PeterTh (approved):
LGTM.

PR #154, opened 2022-12-12
Fix: Allocate memory for data frame padding

fknorr:
We send data frames in "packets" of 64 bytes (#153) to work around MPI's 32-bit-size limitation. The necessary rounding operation was only applied to the argument to `MPI_Isend`, but not the frame allocation. As a result, MPI overran the buffer  and sometimes read from unmapped memory.

This PR moves the padding computation to the `unique_frame_ptr` constructor through a `packet_size` parameter.


psalz (approved):
LGTM!

BlackMark29A (approved):
LGTM!

PR #155, opened 2022-12-14
Execute horizons on dry runs 

facuMH:
If no horizons are executed then task ring buffer doesn't update the counter of deleted tasks and gets full when it should not.

### DIFF
--- a/src/runtime.cc
+++ b/src/runtime.cc
@@ -264,7 +264,7 @@ namespace detail {
 
 	void runtime::flush_command(node_id target, unique_frame_ptr<command_frame> frame) {
 		if(is_dry_run()) {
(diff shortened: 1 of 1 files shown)


fknorr:
LGTM!

fknorr (approved):

src/runtime.cc
```diff
@@ -265,7 +265,7 @@ namespace detail {
 	void runtime::flush_command(node_id target, unique_frame_ptr<command_frame> frame) {
 		if(is_dry_run()) {
 			// We only want to send epochs to the master node for slow full sync and shutdown.
```
psalz, line 267:
```suggestion
			// Only flush epochs (for slow_full_sync / shutdown) and horizons (for deleting tasks from the ring buffer).
```

PR #156, opened 2022-12-14
Return const-reference instead of value from read-accessor operator[]

fknorr:
This becomes visible when trying to take the address of a buffer element.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -257,7 +257,7 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	}
 
 	template <access_mode M = Mode, int D = Dims>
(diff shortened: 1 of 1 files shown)


BlackMark29A (approved):
Returning a value instead of a const-reference from an index operator seems strange to me in the first place. So unless there was a particular reason why this was done I would say changing it to const-reference is the correct thing to do.

fknorr:
I believe it was simply a typo. SYCL also mandates references in the return type: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors

psalz (approved):

PR #157, opened 2022-12-22
cmake: Use FetchContent instead of submodules for dependencies

BlackMark29A:
Using submodules for dependencies is quite cumbersome when installing the library, as install targets have to be manually provided for each dependency. CMake FetchContent allows for much easier and more streamlined dependency handling, without having to have hacky installation targets. It also takes care of transitive dependencies.


BlackMark29A:
I didn't know it was possible to fetch from submodules, otherwise I would have done that to begin with. Thanks for pointing out that that works, I've changed it now.

CMakeLists.txt
```diff
@@ -150,20 +150,36 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # 3rdparty dependencies
-macro(add_submodule_directory RELPATH)
-  if(NOT EXISTS "${PROJECT_SOURCE_DIR}/${RELPATH}/CMakeLists.txt")
-    message(SEND_ERROR "${RELPATH} is missing, perform a git submodule update")
```
psalz, line 155: I feel like this is still useful to have, can we maybe abstract the call to `FetchContent` so that this check is performed first?

psalz: While we're at it, the message be improved a bit, imo. In the SYCL CTS we use

```
message(FATAL_ERROR "The git submodule ${RELPATH} is missing.\nTry running `git submodule update --init`.")`.
```

BlackMark29A: Done.

psalz: Perfect, thanks!

psalz (approved):

facuMH (approved):
LGTM.

PR #158, opened 2023-01-18
Add libenvpp

facuMH:
For multiple runtime parameters celerity uses environment variables.
This library makes the fetching of these variables easier while providing suggestions if miss spelling some of them.
It is also OS independent, hence we do not have to worry about OS specifics for fetching the env vars.

This PR depends on https://github.com/celerity/celerity-runtime/pull/155

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -2,6 +2,8 @@
 
 #include <cstddef>
 #include <optional>
+#include <string_view>
+#include <vector>
 
 namespace celerity {
 namespace detail {
@@ -47,7 +49,7 @@ namespace detail {
 		std::optional<device_config> m_device_cfg;
 		std::optional<bool> m_enable_device_profiling;
 		size_t m_graph_print_max_verts = 200;
-		int m_dry_run_nodes = 0;
+		size_t m_dry_run_nodes = 0;
 	};
 
 } // namespace detail
--- a/src/config.cc
+++ b/src/config.cc
@@ -5,7 +5,6 @@
 #include <sstream>
 #include <string_view>
 #include <thread>
-#include <vector>
 
 #include <mpi.h>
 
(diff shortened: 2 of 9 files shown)


src/config.cc
```diff
@@ -13,6 +13,18 @@
...
+	auto result = std::vector<std::string>{};
+	auto sstream = std::istringstream(std::string(str));
+	auto item = std::string{};
+	while(std::getline(sstream, item, delimiter)) {
+		result.push_back(std::move(item));
+	}
+	return result;
+}
+
 std::pair<bool, std::string> get_env(const char* key) {
```
BlackMark29A, line 28: `get_env` is now no longer required as this is handled by `libenvpp`.

src/config.cc
```diff
@@ -47,6 +59,61 @@ std::pair<bool, size_t> parse_uint(const char* str) {
 	return {false, 0};
 }
 
+using device_config_vector = std::vector<size_t>;
```
BlackMark29A, line 62: This does not create a new unique type, meaning that the specialization of `default_parser` and `default_validator` for this type applies to **all** `std::vector<size_t>` which is not a good idea, as other `vector<size_t>` might have different requirements.

For this use-case it probably makes the most sense to provide a custom parser and validator lambda to `register_variable` instead of specializing the `default_parser` and `default_validator`, but that would of course also be possible when wrapping the vector in a custom unique type.

src/config.cc
```diff
@@ -78,127 +145,85 @@ namespace detail {
...
+		    "LOG_LEVEL", {log_level::trace, log_level::debug, log_level::info, log_level::warn, log_level::err, log_level::critical, log_level::off});
+		const auto env_gpmv = pref.register_variable<size_t>("GRAPH_PRINT_MAX_VERTS");
+		const auto env_devs = pref.register_variable<std::vector<size_t>>("DEVICES");
+		const auto env_profile_ocl = pref.register_variable<int>("PROFILE_OCL");
+		const auto env_profile_kernel = pref.register_variable<int>("PROFILE_KERNEL");
+		const auto env_force_wg = pref.register_variable<bool>("FORCE_WG");
+		const auto env_dry_run_nodes = pref.register_variable<size_t>("DRY_RUN_NODES");
+
+		const auto parsed_and_validated_envs = pref.parse_and_validate();
+		if(parsed_and_validated_envs.ok()) {
```
BlackMark29A, line 159: Most if not all of the logic below should probably be a validation step of libenvpp, as that is what validations were designed for. The idea is that if there's a value that can be retrieved from libenvpp after parsing and validating that value can directly be used, without needing further checking.

BlackMark29A (changes requested):
Already looks much better than before, but there are still some issues.

One of them is that there are no unit tests to make sure environment variable parsing works as expected. To this end libenvpp supports the `set_for_testing` function on the `prefix`, which can be used in unit tests to specify which value a certain variable should have.

Further issues are:

src/config.cc
```diff
@@ -5,200 +5,198 @@
...
+		std::vector<size_t> devices;
+		const auto split_str = split(str, ' ');
+		// Delegate parsing of primitive types to the default_parser
+		for(size_t i = 0; i < split_str.size(); ++i) {
+			devices.push_back(env::default_parser<size_t>{}(split_str[i]));
+		}
+		if(devices.size() < 2)
+			throw env::validation_error{
+			    "Found " + std::to_string(devices.size())
+			    + "\nExpected the following format:CELERITY_DEVICES=\"<platform_id> <first device_id> <second device_id> ... <nth device_id>\""};
```
BlackMark29A, line 93: There should be a space after the colon and the error message could be written more cleanly using `fmt`.

src/config.cc
```diff
@@ -5,200 +5,198 @@
...
-			if(result.first) { m_enable_device_profiling = result.second == "1"; }
-		}
+			// ----------------------------- CELERITY_PROFILE_KERNEL ------------------------------
 
-		// -------------------------------- CELERITY_FORCE_WG ---------------------------------
+			const auto has_profile_kernel = parsed_and_validated_envs.get(env_profile_kernel);
+			if(has_profile_kernel) { m_enable_device_profiling = *has_profile_kernel; }
 
-		{
-			const auto result = get_env("CELERITY_FORCE_WG");
```
BlackMark29A, line 190: `CELERITY_FORCE_WG` is now completely missing, but it should still print the warning that support for it has been dropped if it is specified by the user.

src/config.cc
```diff
@@ -5,200 +5,198 @@
...
+			if(has_dry_run_nodes) { m_dry_run_nodes = *has_dry_run_nodes; }
+
+		} else {
+			for(const auto& warn : parsed_and_validated_envs.errors()) {
+				CELERITY_WARN(warn.what());
+			}
+			for(const auto& err : parsed_and_validated_envs.warnings()) {
+				CELERITY_ERROR(err.what());
 			}
+		throw std::runtime_error("Please make sure to use the environment variables correctly.");
```
BlackMark29A, line 199: Wrong indentation.
Also I'm not sure about this error message. I feel like telling the user to use the environment variables correctly is neither useful nor necessary. A much simpler `"Failed to parse/validate environment variables"` would probably be suffcient without sounding patronizing.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,50 @@ namespace detail {
 		dry_run_with_nodes(nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reads correctly environment varibles", "[env-vars]") {
```
BlackMark29A, line 1018: The test name is not proper english. Should probably be something like `"Config reads environment variables correctly"`.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,50 @@ namespace detail {
...
+		std::unordered_map<std::string, std::string> valid_test_env_vars{//
+		    {"CELERITY_LOG_LEVEL", "debug"},                             //
+		    {"CELERITY_GRAPH_PRINT_MAX_VERTS", "1"},                     //
+		    {"CELERITY_DEVICES", "0 0"},                                 //
+		    {"CELERITY_PROFILE_KERNEL", "1"},                            //
+		    {"CELERITY_DRY_RUN_NODES", "4"}};
```
BlackMark29A, line 1024: Adding a additional comma after the last entry (`{"CELERITY_DRY_RUN_NODES", "4"},`) will tell `clang-format` to format it like this, without the need to force this layout by adding empty comments.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,50 @@ namespace detail {
...
+		REQUIRE(cfg.get_graph_print_max_verts() == 1);
+		REQUIRE(config_testspy::get_device_list(cfg) == std::vector<size_t>{0, 0});
+		const auto has_prof = cfg.get_enable_device_profiling();
+		REQUIRE(has_prof.has_value());
+		REQUIRE((*has_prof) == true);
+	}
+
+	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reports incorrect environment varibles", "[env-vars]") {
+		{
+			std::unordered_map<std::string, std::string> valid_test_env_vars{{"CELERITY_LOG_LEVEL", "a"}};
```
BlackMark29A, line 1039: They can't be `valid_test_env_vars` if they are supposed to cause an error.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,50 @@ namespace detail {
...
+			const auto test_env = env::scoped_test_environment(valid_test_env_vars);
+			CHECK_THROWS_WITH((celerity::detail::config(nullptr, nullptr)), "Please make sure to use the environment variables correctly.");
+		}
+
+		{
+			std::unordered_map<std::string, std::string> valid_test_env_vars{{"CELERITY_DRY_RUN_NODES", "a"}};
+			const auto test_env = env::scoped_test_environment(valid_test_env_vars);
+			CHECK_THROWS_WITH((celerity::detail::config(nullptr, nullptr)), "Please make sure to use the environment variables correctly.");
+		}
+	}
```
BlackMark29A, line 1061: Why is there no fail test for `CELERITY_PROFILE_KERNEL`?

BlackMark29A (changes requested):

src/config.cc
```diff
@@ -5,200 +5,199 @@
...
-						m_device_cfg = device_config{pid_parsed.second, did_parsed.second};
-					}
-				}
-			}
-		}
+			// --------------------------------- CELERITY_DEVICES ---------------------------------
 
-		// ----------------------------- CELERITY_PROFILE_KERNEL ------------------------------
-		{
-			const auto result = get_env("CELERITY_PROFILE_OCL");
```
BlackMark29A, line 175: Why has this been dropped completely? It should either still exist and be parsed, but print the warning that it's been renamed, or it should just fail saying that it has been renamed in the error message.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,58 @@ namespace detail {
...
+
+		const auto test_env = env::scoped_test_environment(valid_test_env_vars);
+		auto cfg = config(nullptr, nullptr);
+
+		REQUIRE(spdlog::get_level() == spdlog::level::debug);
+		REQUIRE(cfg.get_graph_print_max_verts() == 1);
+		REQUIRE(config_testspy::get_device_list(cfg) == std::vector<size_t>{0, 0});
+		const auto has_prof = cfg.get_enable_device_profiling();
+		REQUIRE(has_prof.has_value());
+		REQUIRE((*has_prof) == true);
```
BlackMark29A, line 1035: There is no check inspecting that the number of dry run nodes was set correctly.

BlackMark29A (changes requested):

include/config.h
```diff
@@ -48,6 +51,16 @@ namespace detail {
...
+
+		// required for test
+		std::vector<size_t> m_devices_list;
+
+		// parsing functions
+		size_t parse_validate_graph_print_max_verts(const std::string_view str);
+		bool parse_validate_profile_kernel(const std::string_view str);
+		size_t parse_validate_dry_run_nodes(const std::string_view str);
+		std::vector<size_t> parse_validate_devices(const std::string_view str);
+		int parse_validate_forge_wg(const std::string_view str);
```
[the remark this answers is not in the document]
facuMH: I frist tried to implement it as `void` and came across that error. Hence I made it `int` because it does not get used. But I see your point. I will change it to `bool`

src/config.cc
```diff
@@ -5,200 +5,206 @@
...
+		std::vector<size_t> devices;
+		const auto split_str = split(str, ' ');
+		// Delegate parsing of primitive types to the default_parser
+		for(size_t i = 0; i < split_str.size(); ++i) {
+			devices.push_back(env::default_parser<size_t>{}(split_str[i]));
+		}
+		if(devices.size() < 2)
+			throw env::validation_error{fmt::format(
+			    "Found {} IDs.\nExpected the following format: CELERITY_DEVICES=\"<platform_id> <first device_id> <second device_id> ... <nth device_id>\"",
+			    std::to_string(devices.size()))};
```
BlackMark29A, line 91: This no longer requires `to_string` as `fmt` does that automatically.

src/config.cc
```diff
@@ -5,200 +5,206 @@
...
+		if(static_cast<long>(devices.size()) - 1 > static_cast<long>(m_host_cfg.node_count)) {
+			throw env::validation_error{fmt::format("CELERITY_DEVICES contains {} device indices, but only {} worker processes were spawned on this host",
+			    devices.size() - 1, m_host_cfg.node_count)};
+		}
+
+		return devices;
+	}
+
+	bool config::parse_validate_force_wg(const std::string_view str) {
+		throw env::validation_error{fmt::format("Support for CELERITY_FORCE_WG has been removed with Celerity 0.3.0.")};
```
BlackMark29A, line 106: There's no formatting going on here, so wrapping the string in `fmt::format` is pointless. Same goes for the function below.

src/config.cc
```diff
@@ -5,200 +5,206 @@
...
+	}
+
+	std::vector<size_t> config::parse_validate_devices(const std::string_view str) {
+		std::vector<size_t> devices;
+		const auto split_str = split(str, ' ');
+		// Delegate parsing of primitive types to the default_parser
+		for(size_t i = 0; i < split_str.size(); ++i) {
+			devices.push_back(env::default_parser<size_t>{}(split_str[i]));
+		}
+		if(devices.size() < 2)
```
BlackMark29A, line 88: Please add `{}` to this `if` block.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,59 @@ namespace detail {
...
+			std::unordered_map<std::string, std::string> invalid_test_env_var{{"CELERITY_DRY_RUN_NODES", "a"}};
+			const auto test_env = env::scoped_test_environment(invalid_test_env_var);
+			CHECK_THROWS_WITH((celerity::detail::config(nullptr, nullptr)), error_string);
+		}
+
+		{
+			std::unordered_map<std::string, std::string> invalid_test_env_var{{"CELERITY_PROFILE_KERNEL", "a"}};
+			const auto test_env = env::scoped_test_environment(invalid_test_env_var);
+			CHECK_THROWS_WITH((celerity::detail::config(nullptr, nullptr)), error_string);
+		}
```
BlackMark29A, line 1069: The two deprecated environment variables `CELERITY_PROFILE_OCL` and `CELERITY_FORCE_WG` should also be tested.

BlackMark29A (changes requested):

BlackMark29A (approved):
Looks good to me now.

However, before merging this, the libenvpp dependency should not be on something in the development branch, but rather the official release of version 1.1.0. I will wait with the release of that version until other people have also reviewed this PR, so that if further libenvpp features are required I can add them to version 1.1.0 without having to do another release.

src/config.cc
```diff
@@ -5,200 +5,207 @@
...
-		// -------------------------------- CELERITY_DRY_RUN_NODES ---------------------------------
-		{
-			const auto [is_set, value] = get_env("CELERITY_DRY_RUN_NODES");
-			if(is_set) {
-				const auto [is_valid, num_nodes] = parse_uint(value.c_str());
-				if(!is_valid) { CELERITY_WARN("CELERITY_DRY_RUN_NODES contains invalid value - will be ignored"); }
-				m_dry_run_nodes = num_nodes;
+		} else {
+			for(const auto& warn : parsed_and_validated_envs.errors()) {
+				CELERITY_WARN(warn.what());
```
psalz, line 203: It seems like you mixed up errors / warnings here!

src/config.cc
```diff
@@ -5,200 +5,207 @@
...
+			}
+			for(const auto& err : parsed_and_validated_envs.warnings()) {
+				CELERITY_ERROR(err.what());
 			}
+			throw std::runtime_error("Failed to parse/validate environment variables.");
```
psalz, line 208: Should we really be aborting everything if there's warnings?

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,70 @@ namespace detail {
...
+		    {"CELERITY_GRAPH_PRINT_MAX_VERTS", "1"},
+		    {"CELERITY_DEVICES", "0 0"},
+		    {"CELERITY_PROFILE_KERNEL", "1"},
+		    {"CELERITY_DRY_RUN_NODES", "4"},
+		};
+
+		const auto test_env = env::scoped_test_environment(valid_test_env_vars);
+		auto cfg = config(nullptr, nullptr);
+
+		REQUIRE(spdlog::get_level() == spdlog::level::debug);
```
psalz, line 1030: Prefer `CHECK` over `REQUIRE` unless the code afterwards would crash (e.g. if there's fewer than expected elements in array).

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,70 @@ namespace detail {
 		dry_run_with_nodes(nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reads environment variables correctly", "[env-vars]") {
```
psalz, line 1018: Consider adding `[config]` as a tag here.

include/config.h
```diff
@@ -48,6 +51,17 @@ namespace detail {
 		std::optional<bool> m_enable_device_profiling;
 		size_t m_graph_print_max_verts = 200;
 		int m_dry_run_nodes = 0;
+
+		// required for test
+		std::vector<size_t> m_devices_list;
```
psalz, line 56: This doesn't just store devices, it's a platform followed by a list of devices. But I don't see why this is required anyway, can't you just check that `m_device_cfg` is set correctly?

psalz (changes requested):
Thanks, I've added some notes.

src/config.cc
BlackMark29A: I would argue that we should, because warnings are also always some kind of user error that would result in unexpected behavior. Like a typo in an optional variable, yes it's safe, because the default will get used, but also the user meant to provide a value which won't actually get used. The warning might slip by unnoticed, while an aborted run will not.

psalz: Okay, fine, but then we should print them all as errors, because warnings shouldn't abort execution, IMO.

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,73 @@ namespace detail {
 		dry_run_with_nodes(nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reads environment variables correctly", "[env-vars][config]") {
+		std::unordered_map<std::string, std::string> valid_test_env_vars{
+		    {"CELERITY_LOG_LEVEL", "debug"},
+		    {"CELERITY_GRAPH_PRINT_MAX_VERTS", "1"},
+		    {"CELERITY_DEVICES", "0 0"},
```
psalz, line 1022: You could make the test a bit more robust by using something other than 0/0 here.

src/config.cc
BlackMark29A: Yes, I agree. They should all be printed as errors.

src/config.cc
```diff
@@ -5,200 +5,213 @@
 #include <sstream>
 #include <string_view>
 #include <thread>
-#include <vector>
```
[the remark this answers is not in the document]
facuMH: it's included by `config.h`

psalz (approved):
LGTM!

test/runtime_tests.cc
```diff
@@ -1013,5 +1015,73 @@ namespace detail {
 		dry_run_with_nodes(nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reads environment variables correctly", "[env-vars][config]") {
```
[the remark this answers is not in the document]
psalz: lol, we may have to disable the cognitive complexity check...

src/config.cc
```diff
@@ -5,200 +5,213 @@
...
+
+		return lvl;
+	}
+};
+} // namespace env
 
 namespace celerity {
 namespace detail {
+
+	size_t config::parse_validate_graph_print_max_verts(const std::string_view str) {
```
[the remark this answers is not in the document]
psalz: Maybe these should all just be free functions?

BlackMark29A:
I released v1.1.0 of libenvpp. So this PR can be updated to use the official release and then be merged.

src/config.cc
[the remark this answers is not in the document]
psalz: Ok, but we try to avoid relying on transitive includes usually

PR #160, opened 2023-01-24
Add constexpr for master node id

facuMH:
As per best practices, it is always better to have a `constexpr` than a magic number.


include/mpi_support.h
```diff
@@ -2,11 +2,14 @@
...
 #include <mpi.h>
 
+#include "types.h"
+
 namespace celerity::detail::mpi_support {
 
 constexpr int TAG_CMD = 0;
 constexpr int TAG_DATA_TRANSFER = 1;
 constexpr int TAG_TELEMETRY = 2;
+constexpr node_id MASTER_NODE_ID = 0;
```
fknorr, line 12: Constants should be in snake case (the `TAG` constants are "legacy format"). I suggest to renaming this simply to `master_node` (without the `_id`).

include/mpi_support.h
```diff
@@ -2,11 +2,14 @@
 
 #include <mpi.h>
 
+#include "types.h"
```
fknorr, line 5: We should move the constant definition up to types.h (and out of the `mpi_support` namespace) since it has nothing to do with MPI.

fknorr (changes requested):
Additional places where we should use the named constant:

- `graph_generation_tests:112`
- `graph_generation_tests:114`
- `graph_generation_tests:387`
- `graph_generation_tests:389`
- `graph_generation_tests:437`
- `graph_generation_tests:439`

include/mpi_support.h
facuMH: I will move it to `types.h` but on one hand, I don't fully agree that it has "nothing to do with MPI", on the other, since it is not a type, maybe we should have a separate header for all the constants.

fknorr (approved):

PeterTh (approved):
LGTM.

PR #162, opened 2023-03-08
Use native ("USM") pointers for backing buffer allocations

psalz:
This transitions Celerity away from using SYCL buffers for backing buffer allocations to essentially device-native memory allocations through `sycl::malloc_device` and `sycl::free`, which are part of SYCL 2020's USM APIs.

This has several advantages:
- It gives us much more control over device memory usage, as we don't have to guess what any given SYCL implementation will do (this is particularly important for multi-GPU support).
- It allows us to greatly simplify our accessor implementation, as we no longer need to wrap SYCL accessors internally.
- This in turn means that we completely bypass any dataflow tracking within the SYCL runtime. This was always an unnecessary overhead as we have much more precise tracking information anyway.
- It opens up the possibility for enabling special vendor-specific optimizations such as CUDA-aware MPI later down the road.

Unfortunately there is also a major downside, which is that SYCL's USM capabilities are very much 1-dimensional at the moment.
This is reflected across all USM APIs, and particularly problematic for us is that there is no way of doing 2D/3D rectangular (strided) copies. Dispatching a series of 1D copies in a loop is not a viable solution, as it is extremely slow. Instead, for now (until SYCL Next, hopefully), we have to resort back to doing copies manually using the underlying vendor APIs. This means we need specialized code paths for each of the backends that we want to support efficiently.

Currently I have only implemented a specialized backend for CUDA as well as a generic (and slow) fallback, however adding new backends should be relatively straightforward. Backends are selected dynamically based on the device used. Since SYCL does not yet officially support CUDA as a backend, we also need an implementation-specific mechanism for detecting whether a device is a CUDA device; this is currently supported for OpenSYCL and DPC++.

I wasn't quite sure what the best way of structuring the backend system would be. Loading specialized libraries dynamically at runtime (similar to what OpenSYCL does) seemed overkill, so I've instead opted to make enabling/disabling backends a compile-time option (`-DCELERITY_ENABLE_CUDA_BACKEND`).

This makes testing somewhat more difficult though, as the behavior of test cases is now affected by how they are compiled, which SYCL implementation is used, and on the hardware available during runtime. While I've added some unit tests that just opportunistically test whatever is available, to cover more of these combinations in our CI setup, I've also created our first *true* integration test (we previously used our examples as integration tests, which was always suboptimal) in the form of a Python script that compiles and runs different backend configurations.


include/backend/operations.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <stdexcept>
+
+#include "backend/type.h"
+
+namespace celerity::detail::backend_detail {
```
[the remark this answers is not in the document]
psalz: I'm not sure what is happening here, for some reason `clang-tidy` (and my local `clangd`) for that matter is confused by the backend system, maybe because it uses a nested CMakeLists...?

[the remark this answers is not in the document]
psalz: nvm, I forgot to set the `CXX_STANDARD` on `celerity_backends`.

include/handler.h
```diff
@@ -617,10 +614,10 @@ void handler::parallel_for_kernel_and_reductions(range<Dims> global_range, id<Di
 		} else if constexpr(!CELERITY_FEATURE_SCALAR_REDUCTIONS && sizeof...(reductions) > 1) {
 			static_assert(detail::constexpr_false<Kernel>, "DPC++ currently does not support more than one reduction variable per kernel");
 		} else if constexpr(std::is_same_v<KernelFlavor, detail::simple_kernel_flavor>) {
-			detail::invoke_sycl_parallel_for<KernelName>(cgh, chunk_range, detail::make_sycl_reduction(cgh, reductions)...,
-			    detail::bind_simple_kernel(kernel, global_range, global_offset, chunk_offset));
+			detail::invoke_sycl_parallel_for<KernelName>(
```
fknorr, line 617: Reminder for the future that we should get rid of this path entirely as it's fundamentally broken.

include/reduction_manager.h
```diff
@@ -36,9 +36,9 @@ namespace detail {
 				acc = m_op(acc, *static_cast<const DataT*>(data.get_pointer()));
 			}
 
-			auto host_buf = runtime::get_instance().get_buffer_manager().get_host_buffer<DataT, Dims>(
+			auto info = runtime::get_instance().get_buffer_manager().access_host_buffer(
```
fknorr, line 39: Can be const.

src/backend/cuda_backend.cc
```diff
@@ -0,0 +1,49 @@
...
+
+void memcpy_strided_device_cuda(sycl::queue& queue, const void* source_base_ptr, void* target_base_ptr, size_t elem_size, const sycl::range<1>& source_range,
+    const sycl::id<1>& source_offset, const sycl::range<1>& target_range, const sycl::id<1>& target_offset, const sycl::range<1>& copy_range) {
+	(void)queue;
+	const size_t line_size = elem_size * copy_range[0];
+	const auto ret = cudaMemcpy(static_cast<char*>(target_base_ptr) + elem_size * get_linear_index(target_range, target_offset),
+	    static_cast<const char*>(source_base_ptr) + elem_size * get_linear_index(source_range, source_offset), line_size, cudaMemcpyDefault);
+	if(ret != cudaSuccess) throw std::runtime_error("cudaMemcpy failed");
+	// Classic CUDA footgun: Memcpy is not always synchronous (e.g. for D2D)
+	cudaStreamSynchronize(0);
```
fknorr, line 17: `cudaStreamSynchronize` can also fail (especially if the `cudaMemcpy` is async). 

Suggestion / backport from distr-gather for all CUDA calls in this file:

```c++
// clang-format off
#define CELERITY_STRINGIFY2(f) #f
#define CELERITY_STRINGIFY(f) CELERITY_STRINGIFY2(f)
#define CELERITY_CUDA_CHECK(f, ...)                                                               \
	if (const auto cuda_check_result = (f)(__VA_ARGS__); cuda_check_result != cudaSuccess) {      \
		CELERITY_CRITICAL(CELERITY_STRINGIFY(f) ": {}", cudaGetErrorString(cuda_check_result));   \
		abort();                                                                                  \
	}
// clang-format on

CELERITY_CUDA_CHECK(cudaStreamSynchronize, 0);
```

src/backend/generic_backend.cc
```diff
@@ -0,0 +1,48 @@
...
+	evt.wait();
+}
+
+// TODO Optimize for contiguous copies?
+void memcpy_strided_device_generic(sycl::queue& queue, const void* source_base_ptr, void* target_base_ptr, size_t elem_size, const sycl::range<2>& source_range,
+    const sycl::id<2>& source_offset, const sycl::range<2>& target_range, const sycl::id<2>& target_offset, const sycl::range<2>& copy_range) {
+	const auto source_base_offset = get_linear_index(source_range, source_offset);
+	const auto target_base_offset = get_linear_index(target_range, target_offset);
+	const size_t line_size = elem_size * copy_range[1];
+	std::vector<sycl::event> wait_list;
```
fknorr, line 21: Consider explicitly giving the vector size as `copy_range[0]` here.

src/buffer_storage.cc
```diff
@@ -34,8 +34,9 @@ namespace detail {
 		const auto target_base_offset = get_linear_index(target_range, target_offset)
 		                                - get_linear_index(cl::sycl::range<2>{target_range[1], target_range[2]}, {target_offset[1], target_offset[2]});
 		for(size_t i = 0; i < copy_range[0]; ++i) {
-			const auto source_ptr = static_cast<const std::byte*>(source_base_ptr) + elem_size * (source_base_offset + i * (source_range[1] * source_range[2]));
-			const auto target_ptr = static_cast<std::byte*>(target_base_ptr) + elem_size * (target_base_offset + i * (target_range[1] * target_range[2]));
+			const auto* const source_ptr =
+			    static_cast<const std::byte*>(source_base_ptr) + elem_size * (source_base_offset + i * (source_range[1] * source_range[2]));
+			auto* const target_ptr = static_cast<std::byte*>(target_base_ptr) + elem_size * (target_base_offset + i * (target_range[1] * target_range[2]));
 			memcpy_strided(source_ptr, target_ptr, elem_size, cl::sycl::range<2>{source_range[1], source_range[2]}, {source_offset[1], source_offset[2]},
```
fknorr, line 40: Consider renaming this to `memcpy_strided_host`.

test/buffer_manager_test_utils.h
```diff
@@ -38,13 +41,13 @@ namespace test_utils {
 		}
 
 		template <typename DataT, int Dims>
-		cl::sycl::range<Dims> get_backing_buffer_range(detail::buffer_id bid, access_target tgt, cl::sycl::range<Dims> range, cl::sycl::id<Dims> offset) {
+		range<Dims> get_backing_buffer_range(detail::buffer_id bid, access_target tgt, cl::sycl::range<Dims> range, cl::sycl::id<Dims> offset) {
 			if(tgt == access_target::host) {
-				auto info = m_bm->get_host_buffer<DataT, Dims>(bid, cl::sycl::access::mode::read, detail::range_cast<3>(range), detail::id_cast<3>(offset));
-				return info.buffer.get_range();
+				auto info = m_bm->access_host_buffer<DataT, Dims>(bid, cl::sycl::access::mode::read, detail::range_cast<3>(range), detail::id_cast<3>(offset));
```
fknorr, line 46: Can be const.

test/buffer_manager_test_utils.h
```diff
@@ -38,13 +41,13 @@ namespace test_utils {
...
+				return detail::range_cast<Dims>(info.backing_buffer_range);
 			}
-			auto info = m_bm->get_device_buffer<DataT, Dims>(bid, cl::sycl::access::mode::read, detail::range_cast<3>(range), detail::id_cast<3>(offset));
-			return info.buffer.get_range();
+			auto info = m_bm->access_device_buffer<DataT, Dims>(bid, cl::sycl::access::mode::read, detail::range_cast<3>(range), detail::id_cast<3>(offset));
```
fknorr, line 49: Can be const (True for all `auto info` occurrences in this file)

test/integration/backend.cc
```diff
@@ -0,0 +1,90 @@
+#include <celerity.h>
+
+#define ASSERT(...)                                                                                                                                            \
+	if(!(__VA_ARGS__)) {                                                                                                                                       \
+		CELERITY_ERROR("ASSERT({}) on line {} failed\n", #__VA_ARGS__, __LINE__);                                                                              \
+		std::_Exit(EXIT_FAILURE);                                                                                                                              \
```
fknorr, line 6: This should be an `abort()` to trigger a core dump / debugger trap.

test/integration/backend.cc
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+template <int Dims>
+void test_copy(celerity::distr_queue& q) {
+	celerity::buffer<size_t, Dims> buf(celerity::detail::range_cast<Dims>(celerity::range<3>{5, 7, 9}));
+
+	// Initialize on device
+	q.submit([=](celerity::handler& cgh) {
+		celerity::accessor acc{buf, cgh, celerity::access::one_to_one<>{}, celerity::write_only, celerity::no_init};
+		cgh.parallel_for(buf.get_range(), [=](celerity::item<Dims> itm) { acc[itm] = itm.get_linear_id(); });
```
fknorr, line 16: Is this portable without a `class kernel_name`?

fknorr (commented):
Some nitpicks from my side, otherwise I would say this is true and tested by now!

fknorr:
Re the ComputeCpp CI failure which I'm also seeing in #163 : The backend library doesn't set the `CELERITY_CXX_FLAGS`, which leaves CCPP in pre-2020 SYCL mode and causes missing symbol declarations.

include/buffer_storage.h
```diff
@@ -92,109 +117,53 @@
...
-		device_buffer_storage(cl::sycl::range<Dims> range, cl::sycl::queue transfer_queue)
-		    : buffer_storage(range_cast<3>(range), buffer_type::device_buffer), m_transfer_queue(transfer_queue),
-		      m_device_buf(make_device_buf_effective_range(range)) {
-			// We never want SYCL to do any buffer write-backs. While we don't pass any host pointers to SYCL buffers,
-			// meaning there shouldn't be any write-back in the first place, it doesn't hurt to make sure.
-			// (This was prompted by a hipSYCL bug that did superfluous write-backs).
-			m_device_buf.set_write_back(false);
-		}
+		device_buffer_storage(range<Dims> range, sycl::queue owning_queue)
+		    : buffer_storage(range_cast<3>(range), buffer_type::device_buffer), m_owning_queue(std::move(owning_queue)), m_device_buf(range, m_owning_queue) {}
```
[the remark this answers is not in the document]
psalz: hipSYCL issue: queue should be move constructible.

test/backend_tests.cc
```diff
@@ -0,0 +1,157 @@
...
+
+using namespace celerity;
+using namespace celerity::detail;
+
+// NOTE: We currently test backends on a "best effort" basis, that is we test whatever configuration(s) are available.
+// To exhaustively test all backend configurations we'd need full control over the hardware and software (SYCL implementation)
+// as well as a script to generate the necessary build configurations.
+
+template <int Dims>
+void write_linear_ids(size_t* ptr, const range<Dims>& buffer_range) {
```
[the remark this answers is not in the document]
psalz: That's just wrong

include/backend/backend.h
```diff
@@ -0,0 +1,56 @@
...
+#include "backend/traits.h"
+#include "backend/type.h"
+
+// NOTE: These should not leak any symbols from the backend library (i.e. don't include it in the header)
+#if CELERITY_DETAIL_BACKEND_CUDA_ENABLED
+#include "backend/cuda_backend.h"
+#endif
+
+namespace celerity::detail::backend_detail {
+template <template <backend::type> typename Template, typename Callback>
```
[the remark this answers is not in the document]
psalz: Done!

include/device_queue.h
```diff
@@ -245,6 +246,21 @@ namespace detail {
 		const auto device_name = device.template get_info<sycl::info::device::name>();
 		CELERITY_INFO("Using platform '{}', device '{}' ({})", platform_name, device_name, how_selected);
 
+		if constexpr(std::is_same_v<DeviceT, sycl::device>) {
+			if(backend::get_effective_type(device) == backend::type::generic) {
+				if(backend::get_type(device) == backend::type::unknown) {
+					CELERITY_WARN("No backend specialization available for selected platform '{}', falling back to generic. Performance may be degraded.",
+					    device.get_platform().template get_info<sycl::info::platform::name>());
+				} else {
+					CELERITY_WARN("Selected platform '{}' is compatible with specialized {} backend, but it has not been compiled.",
```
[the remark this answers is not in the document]
psalz: Added!

test/backend_tests.cc
```diff
@@ -0,0 +1,157 @@
...
+	case copy_test_type::device_to_host: {
+		return std::pair{host_or_device{sycl::queue{devices[0]}}, host_or_device{std::nullopt}};
+	}
+	}
+}
+
+TEMPLATE_TEST_CASE_SIG("memcpy_strided_device allows to copy between the same device, different devices on the same platform and between host and device",
+    "[backend]", ((int Dims), Dims), 1, 2, 3) {
+	const copy_parameters<Dims> cp;
+	const size_t platform_id = GENERATE(Catch::Generators::range(size_t(0), sycl::platform::get_platforms().size()));
```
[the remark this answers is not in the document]
psalz: Done!

include/accessor.h
```diff
@@ -232,150 +226,71 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
-	    : m_eventual_sycl_cgh(std::get<0>(args)), m_sycl_accessor(std::get<1>(args)), m_index_offset(std::get<2>(args)), m_buffer_range(std::get<3>(args)) {
-		// SYCL 1.2.1 dictates that all kernel parameters must have standard layout.
-		// However, since we are wrapping a SYCL accessor, this assertion fails for some implementations,
-		// as it is currently unclear whether SYCL accessors must also have standard layout.
-		// See https://github.com/KhronosGroup/SYCL-Docs/issues/94
-		// static_assert(std::is_standard_layout<accessor>::value, "accessor must have standard layout");
+	// Constructor for tests, called through accessor_testspy.
+	accessor(DataT* ptr, id<Dims> index_offset, range<Dims> buffer_range) : m_device_ptr(ptr), m_index_offset(index_offset), m_buffer_range(buffer_range) {
+#if CELERITY_WORKAROUND_HIPSYCL
+		static_assert(std::is_trivially_copyable_v<accessor>);
```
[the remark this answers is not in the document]
psalz: Done!

include/buffer_manager.h
```diff
@@ -196,98 +210,25 @@ namespace detail {
 		void set_buffer_data(buffer_id bid, const subrange<3>& sr, unique_payload_ptr in_linearized);
 
 		template <typename DataT, int Dims>
-		access_info<DataT, Dims, device_buffer> get_device_buffer(
-		    buffer_id bid, cl::sycl::access::mode mode, const cl::sycl::range<3>& range, const cl::sycl::id<3>& offset) {
-			std::unique_lock lock(m_mutex);
+		access_info access_device_buffer(buffer_id bid, cl::sycl::access::mode mode, const cl::sycl::range<3>& range, const cl::sycl::id<3>& offset) {
```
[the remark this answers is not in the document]
psalz: Yes, good idea. Done!

include/reduction_manager.h
psalz: I wonder why I made these all non-const 🤔 Fixed!

src/backend/cuda_backend.cc
psalz: Good call, fixed!

src/backend/generic_backend.cc
psalz: Okay, sure.

src/buffer_storage.cc
psalz: Done!

test/integration/backend.cc
psalz: Done!

test/buffer_manager_test_utils.h
```diff
@@ -3,14 +3,17 @@
 #include "test_utils.h"
 
 namespace celerity {
+
 namespace detail {
+	struct accessor_testspy {
```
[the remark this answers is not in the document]
psalz: I think moving it once we need it is fine, in particular since I don't think we'll need this for anything other than testing the buffer manager. Usually we either use mock buffers without accessors, or we do a full integration-style test case where we construct accessors using their public APIs.

include/handler.h
fknorr: (This is now being addressed by #163)

test/backend_tests.cc
```diff
@@ -0,0 +1,161 @@
...
+			for(size_t i = 0; i < br3[2]; ++i) {
+				const auto linear_id = get_linear_index(buffer_range, id_cast<Dims>(id<3>(k, j, i)));
+				ptr[linear_id] = linear_id;
+			}
+		}
+	}
+}
+
+template <int Dims>
+void verify_copied_linear_ids(const size_t* host_buf, const range<Dims>& source_range, const id<Dims>& source_offset, const range<Dims>& target_range,
```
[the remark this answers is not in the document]
fknorr: Let's get rid of this lint, at least in tests.

test/integration/backend.cc
fknorr: Right, I didn't realize that this is hipSYCL-only code for our purposes.

test/backend_tests.cc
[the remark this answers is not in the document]
psalz: #166

include/accessor.h
```diff
@@ -445,12 +363,12 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 
 	template <access_mode M = Mode, int D = Dims>
 	std::enable_if_t<detail::access::mode_traits::is_producer(M) && (D > 0), DataT&> operator[](id<Dims> index) const {
-		return *(get_buffer().get_pointer() + get_linear_offset(index));
+		return *(m_host_ptr + get_linear_offset(index));
```
fknorr, line 366: While this line is not wrong, it could just use a subscript (see device accessor `operator[]`).

include/accessor.h
```diff
@@ -445,12 +363,12 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
 	}
 
 	template <access_mode M = Mode, int D = Dims>
 	std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M) && (D > 0), const DataT&> operator[](id<Dims> index) const {
-		return *(get_buffer().get_pointer() + get_linear_offset(index));
+		return *(m_host_ptr + get_linear_offset(index));
```
fknorr, line 371: Same here

fknorr (approved):
Looks good to me now!

PR #163, opened 2023-03-14
Support 0-dimensional buffers, accessors and kernels

fknorr:
This PR implements support for 0-dimensional buffers, 0-dimensional global and local accessors as well as 0-dimensional kernels as proposed by the SYCL standard without relying on the feature being present in the SYCL implementation.

Most of the changed lines simply replace `sycl::{range,id}` with `celerity::{range,id}`. We need our own implementation here to support the 0-dimensional case independently from SYCL (in fact the new Celerity coordinate types can be arbitrary-dimensional, even though buffers and kernels are limited to `Dims <= 3` as before).

0-dimensional accessors do not take range mappers on construction since every access must be an all-access anyway. To support the different signatures in CTAD the accessor type deduction logic was re-written from scratch.

The PR also gets rid of the broken backwards compatibility layer for `sycl::item` kernels.

### Open Question

In SYCL, 0-dimensional accessors do not have an `operator[](id<0>)`~~, but only `operator*()`.~~ Should we support this (and also maybe range-mappers on 0-dimensional accessors) as an extension to ease generic programming on buffer dimensionality?

**Edit:** Yes we should.

**Edit 2:** Actually SYCL does not have pointer semantics but implicit conversions to reference-type for 0-dim accessors. We are currently undecided if we want to go the same route.

**Edit 3:** We are going to implement both, our sane API and also SYCL's for porting compatibility.


include/handler.h
```diff
@@ -417,45 +417,33 @@ namespace detail {
...
-				} else {
-					// Explicit item constructor: ComputeCpp does not pass a sycl::item, but an implicitly convertible sycl::item_base (?) which does not have
-					// `sycl::id<> get_id()`
-					invoke_kernel_with_celerity_item(
-					    kernel, cl::sycl::item<Dims>{s_item_or_id}.get_id(), global_range, global_offset, chunk_offset, reducers...);
-				}
-			} else if constexpr(std::is_invocable_v<Kernel, cl::sycl::item<Dims>, decltype(reducers)...>) {
-				invoke_kernel_with_sycl_item(kernel, cl::sycl::item<Dims>{s_item_or_id}, reducers...);
+			static_assert(std::is_invocable_v<Kernel, celerity::item<Dims>, decltype(reducers)...>,
+			    "Kernel function must be invocable with celerity::item<Dims> (or cl::sycl::item<Dims>, "
```
psalz, line 439: Message is outdated

include/grid.h
```diff
@@ -11,11 +11,11 @@ namespace detail {
 
 	using namespace allscale::api::user::data;
 
-	inline GridPoint<1> sycl_id_to_grid_point(cl::sycl::range<1> range) { return GridPoint<1>(range[0]); }
+	inline GridPoint<1> sycl_id_to_grid_point(celerity::range<1> range) { return GridPoint<1>(range[0]); }
```
psalz, line 14: Should we take this opportunity to fix the argument type to be an actual id, instead of a range, and maybe just call it `id_to_grid_point`?

include/handler.h
```diff
@@ -438,11 +438,11 @@ namespace detail {
 			    "Kernel function must be invocable with celerity::item<Dims> (or cl::sycl::item<Dims>, "
 			    "deprecated) and as many reducer objects as reductions passed to parallel_for");
 			if constexpr(CELERITY_WORKAROUND(DPCPP) && std::is_same_v<id<Dims>, decltype(s_item_or_id)>) {
-				// CELERITY_WORKAROUND_LESS_OR_EQUAL: DPC++ passes a sycl::id instead of a sycl::item to kernels alongside reductions
+				// CELERITY_WORKAROUND_LESS_OR_EQUAL: DPC++ passes a celerity::id instead of a sycl::item to kernels alongside reductions
```
psalz, line 442: Wrong auto-replacement

include/handler.h
```diff
@@ -438,11 +438,11 @@ namespace detail {
...
 				invoke_kernel(kernel, s_item_or_id, global_range, global_offset, chunk_offset, reducers...);
 			} else {
 				// Explicit item constructor: ComputeCpp does not pass a sycl::item, but an implicitly convertible sycl::item_base (?) which does not have
-				// `sycl::id<> get_id()`
+				// `celerity::id<> get_id()`
```
psalz, line 446: Wrong auto-replacement

src/task.cc
```diff
@@ -46,7 +46,7 @@ namespace detail {
 			subrange<3> req;
 			switch(kernel_dims) {
 			case 0:
-				[[fallthrough]]; // cl::sycl::range is not defined for the 0d case, but since only constant range mappers are useful in the 0d-kernel case
+				[[fallthrough]]; // celerity::range is not defined for the 0d case, but since only constant range mappers are useful in the 0d-kernel case
```
psalz, line 50: Outdated comment

test/accessor_tests.cc
```diff
@@ -108,18 +108,18 @@ namespace detail {
 	class check_multi_dim_accessor;
 
 	TEMPLATE_TEST_CASE_METHOD_SIG(accessor_fixture, "accessor supports multi-dimensional subscript operator", "[accessor]", ((int Dims), Dims), 2, 3) {
-		// This test *used* to fill a buffer<sycl::id<Dims>> and check that the correct indices have been written. However, this caused the ComputeCpp 2.6.0
+		// This test *used* to fill a buffer<celerity::id<Dims>> and check that the correct indices have been written. However, this caused the ComputeCpp 2.6.0
```
psalz, line 111: That's rewriting history ;-)

include/ranges.h
```diff
@@ -84,17 +430,15 @@ struct chunk {
...
 
 	celerity::id<Dims> offset;
-	celerity::range<Dims> range = detail::range_cast<Dims>(celerity::range<3>(0, 0, 0));
+	celerity::range<Dims> range = detail::zero_range;
 
 	subrange() = default;
 
 	// Due to an apparent bug in the ComputeCpp 2.6.0 compiler, the implicit copy constructor of subrange instantiates a copy constructor of the "array" type
-	// underlying ComputeCpp's implementation of sycl::range. This generated array copy constructor receives a *non-const* lvalue instead of a const lvalue,
+	// underlying ComputeCpp's implementation of celerity::range. This generated array copy constructor receives a *non-const* lvalue instead of a const lvalue,
```
psalz, line 441: Wrong auto-replacement; also this workaround can presumably be removed now!

test/runtime_tests.cc
```diff
@@ -1004,5 +1004,65 @@ namespace detail {
 		dry_run_with_nodes(nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "kernels can access 0-dimensional buffers", "[buffer]") {
+		constexpr float value_a = 13.37f;
+		constexpr float value_b = 42.0f;
+
+		buffer<float, 0> buf_0 = value_a;
```
psalz, line 1011: That's an extension, right? Pretty cool!

include/sycl_wrappers.h
```diff
@@ -6,32 +6,48 @@
...
-inline constexpr detail::write_only_tag_t write_only;
-inline constexpr detail::read_only_host_task_tag_t read_only_host_task;
-inline constexpr detail::read_write_host_task_tag_t read_write_host_task;
-inline constexpr detail::write_only_host_task_tag_t write_only_host_task;
+} // namespace celerity
+
+namespace celerity::detail {
+
+template <access_mode Mode, access_mode NoInitMode, target Target>
+struct access_tag {};
```
psalz, line 21: Very elegant solution!

test/runtime_tests.cc
```diff
@@ -1064,5 +1064,40 @@ namespace detail {
 		});
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "0-dimensional kernels work as expected", "[buffer]") {
+		constexpr float value_a = 13.37f;
+		constexpr float value_b = 42.0f;
+
+		buffer<float, 0> buf_a(value_a);
```
psalz, line 1071: We should probably also have a test (or do this here?) of a 0-dimensional accessing a non-0-dimensional buffer!

include/workaround.h
psalz, line 1: Presumably this is for CCPP? Given #167 I would say lets drop this commit!

psalz (changes requested):
I did a first pass, looks good, thanks! Since you mentioned that you'll have to do the `s/sycl::/celerity::/` replacement again anyway, could you maybe do it as a separate commit to make reviewing easier?

include/handler.h
fknorr: Good catch, updated!

include/grid.h
fknorr: Done both, also fixed up the call sites.

src/task.cc
fknorr: I've opted to get rid of this fall-through and explicitly implement the path for RMs on 0-dimensional buffers or tasks. This a slight breaking change as it disallows `one_to_one` for accessing 1-dimensional buffers in master-node host tasks (which are 0-dimensional). This was used by one test, but IMO it was wrong to support this in the first place.

test/runtime_tests.cc
fknorr: I think so - but legitimate IMO since 0-dimensional buffers are just "managed variables". In that spirit, it's semantically somewhat unfortunate that SYCL defaults `Dims` to 1.

include/sycl_wrappers.h
fknorr: Thanks. It does remove the helpful `Invalid access tag, expecting one of celerity::{read_only,read_write,write_only}[_host_task]` error messages though.

include/handler.h
fknorr: Fixed.

include/handler.h
fknorr: Fixed.

test/accessor_tests.cc
fknorr: Fixed.

include/ranges.h
fknorr: Removed.

test/runtime_tests.cc
fknorr: I've added a test, and also moved accessor-related tests to `accessor_tests.cc`.

include/workaround.h
fknorr: This is reverting some changes from `access::mode` to `access_mode` I erroneously made in earlier commits. Since #167 is not "final" I would avoid breaking CCPP on purpose here.

test/accessor_tests.cc
```diff
@@ -34,7 +34,7 @@ namespace detail {
 	}
 
 	TEST_CASE_METHOD(test_utils::runtime_fixture, "accessors mode and target deduced correctly from SYCL 2020 tag types and no_init property", "[accessor]") {
```
psalz, line 36: This reminds me, I actually have a patch lying around that I wanted to upstream for some time; I think this test can be done purely at compile time: https://gist.github.com/psalz/bfa219820ba6f0e0ee6713d6d0e31a3c. Maybe you want to incorporate this?

include/accessor.h
```diff
@@ -261,44 +286,81 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
+
+	template <access_mode M = Mode>
+	std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M), const DataT&> operator*() const {
+		return *m_device_ptr;
+	}
 
-// Celerity Accessor Deduction Guides
-template <typename T, int D, typename Functor, typename TagT>
-accessor(const buffer<T, D>& buff, handler& cgh, Functor rmfn, TagT tag)
-    -> accessor<T, D, detail::deduce_access_mode<TagT>(), detail::deduce_access_target<std::remove_const_t<TagT>>()>;
```
psalz, line 293: Theres still some stray forward declarations for these functions left on lines 38-45.

include/buffer.h
```diff
@@ -38,14 +38,20 @@ class accessor;
 template <typename DataT, int Dims>
 class buffer {
```
psalz, line 39: `buffer::get_access` needs to be specialized for 0 dimensions!

include/accessor.h
```diff
@@ -469,17 +532,104 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
+
+	friend bool operator!=(const accessor& lhs, const accessor& rhs) { return !(lhs == rhs); }
+
+  private:
+	DataT* m_host_ptr = nullptr;
+};
+
+
+template <typename T, int D, access_mode Mode, access_mode ModeNoInit, target Target>
+accessor(const buffer<T, D>& buff, handler& cgh, const detail::access_tag<Mode, ModeNoInit, Target> tag) -> accessor<T, D, Mode, Target>;
```
psalz, line 600: I feel like it would be clearer if these wouldn't be templated on `D` at all (since `D = 0` anyway), and came after the guides for `D > 0`, in the same order the classes are implemented.

include/accessor.h
```diff
@@ -261,44 +286,81 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
 
 	template <typename Functor>
-	accessor(const ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, Functor rmfn) {
+	accessor(const ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, const Functor& rmfn) {}
+
+	size_t get_linear_offset(const id<Dims>& index) const { return detail::get_linear_index(m_buffer_range, index - m_index_offset); }
+};
+
+template <typename DataT, access_mode Mode>
+class accessor<DataT, 0, Mode, target::device> : public detail::accessor_base<DataT, 0, Mode, target::device> {
```
psalz, line 295: Not sure what your motivations were for doing this as a template specialization, and I'm also not sure how I feel about it. Maybe we can have a discussion about this. Some thoughts:

- It avoids having a ton of SFINAE everywhere, which is obviously nice. OTOH I think it might actually be useful to have the overloads receiving a range mapper be enabled, and then throw a static assertion to guide users towards the correct overload. Maybe we'd even want to accept it as long as the `rmfn` is an `all` range mapper? I'm thinking of generic programming contexts here.
- It saves space, at least two byte, by not including the range/offset, which is important for device accessors at least. Maybe we could achieve the same with some conditional inheritance hackery, though.
- It duplicates the multi-pass (or soon, hydration) logic, which is not great.

psalz (changes requested):
A few more notes!

include/accessor.h
fknorr: This was primarily motivated by the unnecessary data members, as you already recognized, and the fact that `operator[]` exists only for `Dims > 0` while `operator*` is defined only if `Dims == 0`. Other than constructors there is very little overlap between the two cases currently, but in case we decide to implement `operator[](id<0>)` as an extension (which I'm very much in favor of), this distinction becomes less clear.

include/accessor.h
fknorr: Done.

include/accessor.h
fknorr: Removed.

test/accessor_tests.cc
fknorr: I've included the patch and also added tests for the 0-dimensional case. The `property_list` constructors now produce `write` instead of `discard_write` modes, but this should be irrelevant as they always throw anyway.

include/accessor.h
```diff
@@ -307,61 +354,62 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
-	template <typename Functor, typename TagT>
-	accessor(const buffer<DataT, Dims>& buff, handler& cgh, Functor rmfn, TagT tag) : accessor(buff, cgh, rmfn) {}
+	template <typename Functor, access_mode TagModeNoInit>
+	accessor(const buffer<DataT, Dims>& buff, handler& cgh, const Functor& rmfn, const detail::access_tag<Mode, TagModeNoInit, target::host_task> /* tag */)
+	    : accessor(buff, cgh, rmfn) {}
 
 	/**
 	 * TODO: As of ComputeCpp 2.5.0 they do not support no_init prop, hence this constructor is needed along with discard deduction guide.
 	 *    but once they do this should be replace for a constructor that takes a prop list as an argument.
 	 */
```
PeterTh, line 392: We no longer support ComputeCpp as of 68367dd, can we just drop this?

include/accessor.h
PeterTh: I'm in favour of having less code duplication in this case (and offering `operator[](id<0>)`).

include/backend/cuda_backend.h
```diff
@@ -1,20 +1,26 @@
 #pragma once
 
-#include <sycl/sycl.hpp>
-
 #include "backend/operations.h"
 #include "backend/type.h"
+#include "ranges.h"
 
 namespace celerity::detail::backend_detail {
```
PeterTh, line 7: General readability question: do we actually need to qualify `celerity::range` and `celerity::id` in these files / inside the celerity namespace?

include/ranges.h
PeterTh, line 1: It would be nice to have basic unit tests for this. 
E.g. something that at least explicitly tests one operation from each category of the stuff defined by `CELERITY_DEFINE_COORDINATE_...`. (By explicitly I mean not recreating the macro machinery to test everything, but just manually testing one of the operations)

PeterTh (commented):
I only looked at 22/44 files so far, but I'll probably not get to the rest soon so I'll post that part for now.

include/accessor.h
```diff
@@ -261,44 +286,81 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	}
 
 	template <typename Functor>
-	accessor(const ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, Functor rmfn) {
+	accessor(const ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, const Functor& rmfn) {}
```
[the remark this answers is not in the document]
psalz: Yes, same thing for `host_task` accessors!

include/accessor.h
fknorr: We can re-think a large part of our `access_mode` logic now that USM buffers are merged and we do not need to pass on the `no_init` information through `access_mode::discard_write` anymore. This would also allow us to support accessor property lists for real (if that is even desirable). Since this line is not directly affected by this PR I would vote to postpone this though.

include/accessor.h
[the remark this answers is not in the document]
fknorr: I have moved the device-accessor constructor body back into the `internal_tag`ged overload. Since there's no `ctor_internal_tag` overloads for `host_task`, I'll leave that one as is.

include/accessor.h
fknorr: Implemented as such.

include/backend/cuda_backend.h
fknorr: I removed all unnecessary (and semantically not interesting) namespace qualifications on these types.

include/accessor.h
```diff
@@ -488,7 +548,8 @@ class local_accessor {
 	using const_reference = const DataT&;
 	using size_type = size_t;
 
-	local_accessor() : m_sycl_acc{make_placeholder_sycl_accessor()}, m_allocation_size(detail::zero_range) {}
```
psalz, line 491: Why did you remove this?

test/accessor_tests.cc
```diff
@@ -19,7 +19,7 @@ namespace detail {
 	TEST_CASE_METHOD(test_utils::runtime_fixture, "accessors behave correctly for 0-dimensional master node kernels", "[accessor]") {
```
psalz, line 19:
```suggestion
	TEST_CASE("accessors behave correctly for 0-dimensional master node kernels", "[accessor]") {
```

include/accessor.h
```diff
@@ -604,8 +605,9 @@ class local_accessor {
 
   private:
 	sycl_accessor m_sycl_acc;
-	range<Dims> m_allocation_size;
+	CELERITY_DETAIL_NO_UNIQUE_ADDRESS range<Dims> m_allocation_size;
 	cl::sycl::handler* const* m_eventual_sycl_cgh = nullptr;
+	// TODO after multi-pass removal: verify that sizeof(celerity::local_accessor) == sizeof(sycl::local_accessor) in accessor_tests
```
psalz, line 610: You could add a `[!shouldfail]` test for this now!

test/accessor_tests.cc
```diff
@@ -507,5 +507,13 @@ namespace detail {
 		});
 	}
 
+#if CELERITY_DETAIL_HAS_NO_UNIQUE_ADDRESS
```
psalz, line 510: Consider moving this into the test and conditionally `SKIP`!

include/accessor.h
fknorr: Done.

test/accessor_tests.cc
fknorr: Also done!

psalz (approved):
LGTM!

PeterTh (approved):
LGTM, but maybe run the GPUC2 benchmarks before merging and update the result files.

PR #164, opened 2023-03-21
Fix empty CELERITY_DPCPP_TARGETS for installed Celerity

psalz:
DPC++ crashes when `-fsycl-targets` is specified twice, and one of them is empty. This is the reason why we didn't have a successful `HEAD` build since October (the non-empty one came transitively from `libcelerity_runtime`).

Edit: [Here's](https://github.com/celerity/celerity-runtime/actions/runs/4488419101/jobs/7892998451) a CI run against HEAD that shows that we are getting past the "Build examples against installed Celerity" step. We now crash later on during a unit test - progress!


fknorr (approved):
LGTM

PR #169, opened 2023-03-27
[CI] Don't pin MPI ranks to CPU cores

psalz:
Currently each rank is limited to two physical cores, this removes the binding. This is not a performance consideration; limiting the number of concurrent threads may mask certain kinds of bugs.


fknorr (approved):
Especially relevant since multiple CI jobs are now scheduled to the same nodes concurrently with SLURM. LGTM!

PR #170, opened 2023-03-30
Add an option to use the mimalloc memory allocator

PeterTh:
Mimalloc is included as a submodule, used with `FetchContent` (as per convention).
~~Increases the required CMake version to 3.21 to enable `$<TARGET_OBJECTS:`.~~

MiMalloc generally seems to improve overall performance in the more realistic/larger microbenchmark scenarios by ~30% compared to the default allocator on GPUC2, which is quite significant for a simple allocator change:

It does not decrease performance in any microbenchmark (log-y):

Drive-by: 
* add a few common entries to gitignore
* enable `compile_commands.json` generation by default (we have discussed that we should do this in several F2F meetings before)


BlackMark29A:
Off-topic, but since you are already drive-by-ing some stuff and are touching the main cmake file, can you also drive-by fix the `chooosing` typo in the main cmake file in line 68?

PeterTh:
We've had an offline discussion regarding the bump in CMake min version. I'll look into whether we can link dynamically to avoid that, and what the perf impact (if any) is.

PeterTh:
The PR has been modified to use mimalloc as a shared library.

This has two consequences:
* it does not require us to bump the CMake version requirement,
* but only allows us to easily override **new**/**delete**-based allocations.

Benchmarks are very slightly different (both in positive and negative direction), but the overall picture and advantage compared to the default allocator remains the same.

fknorr (approved):
I like this, LGTM up to some minor details!

psalz (approved):
LGTM

PR #168, opened 2023-03-27
Update tested CUDA versions documentation

psalz:
In a failed attempt to make CMake find the "CUDA Toolkit" for #162, I've bumped the minimum CUDA version we test against to 11.3, as the corresponding Docker container (built by nvidia) is the first to install the `cuda-toolkit` package. I've then realized that it was actually a CMake version issue. Nevertheless, I think there's no harm in bumping to 11.3, which is about two years old at this point. While at it, I've also bumped the "bleeding edge" container, which we use to test against hipSYCL-HEAD, to use CUDA 12.1.

This just updates the documentation to reflect that.

PR #171, opened 2023-04-03
Build latest hipSYCL against CUDA 12.1 on Ubuntu 23.04

fknorr:
As noted in #168, Clang < 16 is incompatible with CUDA > 12. Docker images have been updated in the background so that with this PR, Ubuntu 22.04 / Clang 14 builds against CUDA 11.8, and a custom Ubuntu 23.04 / Clang 16 image is used to build against CUDA 12.1. To simplify workflow definitions, the hipSYCL version on 22.04 is now pinned.

With hipSYCL on CUDA 12, ptxas segfaults in the convolution example on `fmin(float3, float3)` for an unknown reason. Replacing the operation with a component-wise fmin seems to fix the issue.

### DIFF
--- a/.github/workflows/build_matrix.json
+++ b/.github/workflows/build_matrix.json
@@ -5,13 +5,15 @@
         { "sycl": "dpcpp", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "intel", "build-type": "Debug" },
         { "sycl": "dpcpp", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "intel", "build-type": "Release" },
         { "sycl": "hipsycl", "sycl-version": "24980221", "ubuntu-version": "20.04", "platform": "nvidia", "build-type": "Debug" },
-        { "sycl": "hipsycl", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "nvidia", "build-type": "Debug" },
-        { "sycl": "hipsycl", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "nvidia", "build-type": "Release" }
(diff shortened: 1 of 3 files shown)


.github/workflows/build_matrix.json
```diff
@@ -5,13 +5,15 @@
         { "sycl": "dpcpp", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "intel", "build-type": "Debug" },
         { "sycl": "dpcpp", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "intel", "build-type": "Release" },
         { "sycl": "hipsycl", "sycl-version": "24980221", "ubuntu-version": "20.04", "platform": "nvidia", "build-type": "Debug" },
-        { "sycl": "hipsycl", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "nvidia", "build-type": "Debug" },
-        { "sycl": "hipsycl", "sycl-version": "latest", "ubuntu-version": "22.04", "platform": "nvidia", "build-type": "Release" }
+        { "sycl": "hipsycl", "sycl-version": "373d3119", "ubuntu-version": "22.04", "platform": "nvidia", "build-type": "Debug" },
```
psalz, line 8: Consider using the same revision for 20.04 and 22.04 as there is no reason (that we know of) to distinguish the two, otherwise LGTM!

psalz (approved):

.github/workflows/build_matrix.json
fknorr: Both are now using the older of the two revisions.

keryell:
Just checking: `sycl::float3` has the size of `sycl::float4` which is different for CUDA.
But `sycl::mfloat3` has the size of CUDA float3.
Perhaps a reason?

PR #175, opened 2023-04-14
Lift restrictions on reductions for DPC++

psalz:
DPC++ now supports passing multiple reductions into a kernel. While it still only accepts 1-dimensional buffers (I guess the spec is simply ambiguous on this), that no longer matters to us since we now do a pointer-based reduction internally anyway.


fknorr (changes requested):
Nice! This means we can test full reduction support in CI for the first time.

We should however keep all feature flags intact (`CELERITY_FEATURE_SIMPLE_SCALAR_REDUCTIONS`, `CELERITY_FEATURE_SCALAR_REDUCTIONS`, `CELERITY_FEATURE_LOCAL_ACCESSOR`) since they part of the user-facing API (both CMake variables and compile_definitions).

psalz:
Done!

fknorr (approved):
LGTM

PR #176, opened 2023-04-24
Disallow in-source build to prevent issues

facuMH:
It is always recommended (and a good practice) to compile in a build directory. But considering a celerity user could potentially not be familiarized with good programming/debugging practices, this check should save them some time of headache.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.13)
 
+if (CMAKE_CURRENT_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  message(FATAL_ERROR "Celerity does not support in-source builds.\nPlease use a dedicated build directory and remove ${CMAKE_CURRENT_BINARY_DIR}/CMakeCache.txt and ${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles")
+endif()
+
 file(STRINGS "VERSION" Celerity_VERSION)
 string(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" _ "${Celerity_VERSION}")
 set(CELERITY_VERSION_MAJOR ${CMAKE_MATCH_1})


CMakeLists.txt
```diff
@@ -26,6 +26,16 @@ if(CXX_COMPILER_IS_DPCPP)
   message(STATUS "CXX compiler at ${CMAKE_CXX_COMPILER} detected as DPC++")
 endif()
 
+if ((CMAKE_CURRENT_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR) OR
+  (CMAKE_CURRENT_BINARY_DIR STREQUAL ${CMAKE_CURRENT_SOURCE_DIR}/src) OR
```
psalz, line 30: I don't think we should attempt to prevent *every* source directory from being used as a build directory. If a user really wants to build in `src`, I say let them...

psalz (changes requested):
Thanks, good idea. Unfortunately this doesn't stop `CMakeCache.txt` and `CMakeFiles` from being created (and there doesn't seem to be a way of doing that).

[Apparently](https://stackoverflow.com/a/10306476) there's also two built-in variables one can use to prevent in-source builds (undocumented but available since 2006). Maybe we should just use those instead?

CMakeLists.txt
facuMH: ok, allowing to build in sub dirs will be allowed, but I rather keep a custom error message, since the error message from the undocumented built-in vars is not as direct and clear as a custom one can be.

CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.13)
 
+if (CMAKE_CURRENT_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  message(FATAL_ERROR "Building in root or relevant directories will cause issues.\nPlease use a dedicated build directory and remove the CMakeCache.txt file and CMakeFiles dir created in the root.")
```
psalz, line 4:
```suggestion
  message(FATAL_ERROR "Celerity does not support in-source builds.\nPlease use a dedicated build directory and remove ${CMAKE_SOURCE_DIR}/CMakeCache.txt and ${CMAKE_SOURCE_DIR}/CMakeFiles.")
```

I think we should mention the term "in-source build" explicitly here. Also we can print the exact path, as "root" is ambiguous if Celerity is built as a subproject.

PeterTh (approved):
LGTM.

CMakeLists.txt
```diff
@@ -26,6 +30,7 @@ if(CXX_COMPILER_IS_DPCPP)
   message(STATUS "CXX compiler at ${CMAKE_CXX_COMPILER} detected as DPC++")
 endif()
 
+
```
fknorr, line 33: This blank line seems to be a left-over from multiple edits?

fknorr (approved):
When squashed, LGTM. Thanks!

PR #177, opened 2023-04-25
add print performance counters for buffer_manager and btm

facuMH:
Initial idea for reporting relevant performance counters:
- each relevant class keeps their own counters
- everyone with a relevant counter reports it through `print_performance_counters()` called from runtime and propagated. 

Adding @philippgs since he could come up with other metrics that could be interesting to keep an eye on.


psalz:
Closing as BM / BTM will be superseded by IDAG. Will revisit this idea afterwards.

PR #178, opened 2023-04-25
Accessor boundary check

facuMH:
This PR depends on #173 

This check has proven quite useful when debugging, hence adding only for debug builds since it can also have a non negligible impact on performance, after all it happens for all accessor subscriptor operators.

Not sure if just reporting it as a `CELERITY_ERROR` is enough or if it would also be expected for it to exit and report it as a fatal error or similar.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -254,12 +254,23 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	}
 
 	template <access_mode M = Mode>
-	inline std::enable_if_t<detail::access::mode_traits::is_producer(M), DataT&> operator[](const id<Dims>& index) const {
-		return m_device_ptr[get_linear_offset(index)];
-	}
-
-	template <access_mode M = Mode>
-	inline std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M), const DataT&> operator[](const id<Dims>& index) const {
+	inline std::conditional_t<detail::access::mode_traits::is_producer(M), DataT&, const DataT&> operator[](const id<Dims>& index) const {
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		// We currently don't support boundary checking for accessors created using accessor_testspy::make_device_accessor,
+		// which does not set m_oob_indices.
+		if(m_oob_indices != nullptr) {
+			const id<Dims> all_true = detail::id_cast<Dims>(id<3>(true, true, true));
+			const bool is_within_bounds_lo = (index >= m_accessed_virtual_subrange.offset) == all_true;
+			const bool is_within_bounds_hi = (index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range)) == all_true;
+			if((!is_within_bounds_lo || !is_within_bounds_hi)) {
+				for(int d = 0; d < Dims; ++d) {
+					sycl::atomic_ref<size_t, sycl::memory_order::relaxed, sycl::memory_scope::device>{m_oob_indices[0][d]}.fetch_min(index[d]);
+					sycl::atomic_ref<size_t, sycl::memory_order::relaxed, sycl::memory_scope::device>{m_oob_indices[1][d]}.fetch_max(index[d] + 1);
+				}
+				return m_oob_fallback_value;
+			}
+		}
+#endif
 		return m_device_ptr[get_linear_offset(index)];
 	}
 
@@ -311,6 +322,12 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	DataT* m_device_ptr = nullptr;
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS id<Dims> m_backing_buffer_offset;
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS range<Dims> m_backing_buffer_range = detail::zero_range;
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+	id<3>* m_oob_indices = nullptr;
+	subrange<Dims> m_accessed_virtual_subrange = {};
+	// This value (or a reference to it) is returned for all out-of-bounds accesses.
+	mutable DataT m_oob_fallback_value = DataT{};
+#endif
 
 	template <typename Functor>
 	accessor(const ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, const Functor& rmfn) {
@@ -340,6 +357,10 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 		m_device_ptr = other.m_device_ptr;
 		m_backing_buffer_offset = other.m_backing_buffer_offset;
 		m_backing_buffer_range = other.m_backing_buffer_range;
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		m_oob_indices = other.m_oob_indices;
+		m_accessed_virtual_subrange = other.m_accessed_virtual_subrange;
+#endif
 
 #if !defined(__SYCL_DEVICE_ONLY__)
 		if(detail::is_embedded_hydration_id(m_device_ptr)) {
@@ -352,6 +373,10 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 				m_device_ptr = static_cast<DataT*>(info.ptr);
 				m_backing_buffer_offset = detail::id_cast<Dims>(info.backing_buffer_offset);
 				m_backing_buffer_range = detail::range_cast<Dims>(info.backing_buffer_range);
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+				m_oob_indices = info.out_of_bounds_indices;
+				m_accessed_virtual_subrange = detail::subrange_cast<Dims>(info.accessed_virtual_subrange);
+#endif
 			}
 		}
 #endif
--- a/include/closure_hydrator.h
+++ b/include/closure_hydrator.h
@@ -62,6 +62,10 @@ class closure_hydrator {
 		range<3> backing_buffer_range;
 		id<3> backing_buffer_offset;
 		subrange<3> accessed_virtual_subrange;
+
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		id<3>* out_of_bounds_indices = nullptr;
+#endif
 	};
 
 	closure_hydrator(const closure_hydrator&) = delete;
--- a/include/worker_job.h
+++ b/include/worker_job.h
@@ -180,6 +180,10 @@ namespace detail {
 		cl::sycl::event m_event;
 		bool m_submitted = false;
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		std::vector<id<3>*> m_oob_indices_per_accessor;
+#endif
+
 		bool execute(const command_pkg& pkg) override;
 		std::string get_description(const command_pkg& pkg) override;
 	};
--- a/src/worker_job.cc
+++ b/src/worker_job.cc
@@ -213,7 +213,19 @@ namespace detail {
 				const auto [bid, mode] = access_map.get_nth_access(i);
 				const auto sr = grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
 				const auto info = m_buffer_mngr.access_device_buffer(bid, mode, sr);
+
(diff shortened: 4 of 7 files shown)


test/accessor_tests.cc
```diff
@@ -641,5 +656,38 @@ namespace detail {
 		sycl::free(result, q.get_sycl_queue());
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "out of bounds access gets reported.", "[accessor][oob]") {
```
psalz, line 659:
```suggestion
	TEST_CASE_METHOD(test_utils::runtime_fixture, "accessor reports out-of-bounds accesses"[accessor][oob]") {
```

test/accessor_tests.cc
```diff
@@ -534,6 +534,9 @@ namespace detail {
 
 	TEST_CASE("0-dimensional accessors are pointer-sized", "[accessor]") {
 		if(!CELERITY_DETAIL_HAS_NO_UNIQUE_ADDRESS) SKIP("[[no_unique_address]] not available on this compiler");
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		SKIP("no accessor size guarantees when accessor boundary check is enabled.");
```
psalz, line 538:
```suggestion
		SKIP("no accessor size guarantees when CELERITY_ACCESSOR_BOUNDARY_CHECK=1");
```

test/accessor_tests.cc
```diff
@@ -641,5 +656,38 @@ namespace detail {
 		sycl::free(result, q.get_sycl_queue());
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "out of bounds access gets reported.", "[accessor][oob]") {
+#if !CELERITY_ACCESSOR_BOUNDARY_CHECK
+		SKIP("accessor boundary check is disabled.");
```
psalz, line 661:
```suggestion
		SKIP("CELERITY_ACCESSOR_BOUNDARY_CHECK=0");
```

include/accessor.h
```diff
@@ -254,12 +254,17 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
 	template <access_mode M = Mode>
-	inline std::enable_if_t<detail::access::mode_traits::is_producer(M), DataT&> operator[](const id<Dims>& index) const {
-		return m_device_ptr[get_linear_offset(index)];
-	}
-
-	template <access_mode M = Mode>
-	inline std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M), const DataT&> operator[](const id<Dims>& index) const {
+	inline std::conditional_t<detail::access::mode_traits::is_producer(M), DataT&, const DataT&> operator[](const id<Dims>& index) const {
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		if(((index - m_backing_buffer_offset) >= m_backing_buffer_range) != detail::id_cast<Dims>(id<3>(false, false, false))) {
```
psalz, line 259: Should we be checking against the backing buffer range here, or against what the user has declared they are going to access? I'm thinking the latter! (This would require adding one additional field to the accessor).

test/accessor_tests.cc
```diff
@@ -641,5 +656,38 @@ namespace detail {
...
+
+		q.submit([&](handler& cgh) {
+			accessor acc1(buff_1, cgh, celerity::access::all(), celerity::read_only_host_task);
+			cgh.host_task(on_master_node, [=, &validation_values] {
+				validation_values[0] = acc1[0];
+				validation_values[1] = acc1[1];
+			});
+		});
+		q.slow_full_sync();
+		CHECK(validation_values[1] == original_values[0]);
```
psalz, line 688: As far as I can tell, this only works because of the current behavior of OOB-accesses returning a reference to `ptr[0]`. That is pretty arbitrary and IMO we shouldn't be testing it. Instead I think this test should explicitly check for the error message being emitted using a `log_capture`!

include/accessor.h
```diff
@@ -311,13 +316,20 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	DataT* m_device_ptr = nullptr;
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS id<Dims> m_backing_buffer_offset;
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS range<Dims> m_backing_buffer_range = detail::zero_range;
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+	detail::buffer_id m_accessed_buffer_id; // = nullptr;
```
psalz, line 320: This doesn't seem to get used anywhere, remnant of an earlier implementation?

include/accessor.h
```diff
@@ -311,13 +316,20 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
+	id<3>* m_attempted_index = nullptr;
```
psalz, line 321: `index` sounds like it's singular, when in reality it's two indices. Maybe call it something like `m_oob_indices`, or consider splitting it into two, e.g. `m_oob_idx_min` and `m_oob_idx_max`.

include/worker_job.h
```diff
@@ -179,6 +179,8 @@ namespace detail {
 		node_id m_local_nid;
 		cl::sycl::event m_event;
 		bool m_submitted = false;
+		std::vector<buffer_id> m_debug_bid_ptr;
+		std::vector<id<3>*> m_debug_indices_per_acc;
```
psalz, line 183:
```suggestion
		std::vector<id<3>*> m_oob_indices_per_accessor;
```

psalz (changes requested):
Thanks, this will make life a lot easier! I've added some (mostly minor) notes; I think my biggest concern would be that the current implementation only checks against the backing buffer size, not what the user declared (so whether or not the OOB check gets triggered could be dependent on scheduling decisions).

include/worker_job.h
psalz: Also this should be guarded with `#if`s!

include/accessor.h
```diff
@@ -254,12 +254,17 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
-	inline std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M), const DataT&> operator[](const id<Dims>& index) const {
+	inline std::conditional_t<detail::access::mode_traits::is_producer(M), DataT&, const DataT&> operator[](const id<Dims>& index) const {
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		if(((index - m_backing_buffer_offset) >= m_backing_buffer_range) != detail::id_cast<Dims>(id<3>(false, false, false))) {
+			for(int d = 0; d < Dims; ++d) {
+				sycl::atomic_ref<size_t, sycl::memory_order::relaxed, sycl::memory_scope::device>{m_attempted_index[0][d]}.fetch_min(index[d]);
+				sycl::atomic_ref<size_t, sycl::memory_order::relaxed, sycl::memory_scope::device>{m_attempted_index[1][d]}.fetch_max(index[d] + 1);
+			}
+			// TODO: make sure we always allocate at least one element in the backing buffer, if CELERITY_ACCESSOR_BOUNDARY_CHECK is enabled
+			return m_device_ptr[0];
```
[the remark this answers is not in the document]
facuMH: the reasoning here was to prevent the program from crashing (as it would if it were to run in Release build instead of Debug) and to give a cmdln error message to the user, warning that their data is incorrect because of the access. 
Also in some cases doing the out-of-bounds access would result in a hipsycl runtime crash with no info about why is the program crashing, that is why I opted for this instead.

include/host_object.h
```diff
@@ -198,6 +198,6 @@ explicit host_object(T&&) -> host_object<detail::assert_host_object_ctor_param_i
 template <typename T>
 explicit host_object(std::reference_wrapper<T>) -> host_object<T&>;
 
-explicit host_object()->host_object<void>;
+explicit host_object() -> host_object<void>;
```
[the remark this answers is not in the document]
facuMH: if you look at the `host_object` in line 199 you see that the arrow has spaces, when running clang format it added the spaces, so after talking about it with Peter we concluded that this line was probably unformatted when committed.

src/worker_job.cc
```diff
@@ -212,8 +216,20 @@ namespace detail {
 			for(size_t i = 0; i < access_map.get_num_accesses(); ++i) {
 				const auto [bid, mode] = access_map.get_nth_access(i);
 				const auto sr = grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
-				const auto info = m_buffer_mngr.access_device_buffer(bid, mode, sr);
+				[[maybe_unused]] const auto info = m_buffer_mngr.access_device_buffer(bid, mode, sr);
```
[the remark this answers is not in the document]
facuMH: you are right, it's unnecessary. I'll remove it

include/host_object.h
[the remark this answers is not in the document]
psalz: You are probably running a version of clangd > 14, which apparently changed this formatting. CI currently runs 14, and expects the version without spaces!

include/accessor.h
facuMH: I can't recall ATM which was the specific use case, but I recall in a few situations wondering why do accessors don't know their buffer's id, and yes, also most likely earlier impl.

include/accessor.h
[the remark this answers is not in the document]
psalz: I still think out-of-bounds writes shouldn't manifest anywhere in a buffer, that's confusing. A better way of going about this might be to have a "catch all" member of type `DataT` within each accessor, to which a reference is returned in this case.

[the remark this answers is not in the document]
facuMH: as a sort of default "there was a mistake" value of DataT ? but that would have to be provided either when constructing the `accessor` or the `buffer`, and either options seems more cumbersome. Alternatively, I would be willing to simply attempt the out-of-bounds access, provided @fknorr and @PeterTh agree on it too.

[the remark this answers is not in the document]
psalz: No, just a default-constructed value of type `DataT`, simply because we need *something* to return a reference to from `operator[]`.

psalz (changes requested):

include/accessor.h
```diff
@@ -254,12 +254,24 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
...
-	template <access_mode M = Mode>
-	inline std::enable_if_t<detail::access::mode_traits::is_pure_consumer(M), const DataT&> operator[](const id<Dims>& index) const {
+	inline std::conditional_t<detail::access::mode_traits::is_producer(M), DataT&, const DataT&> operator[](const id<Dims>& index) const {
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		// We currently don't support boundary checking for accessors created using accessor_testspy::make_device_accessor,
+		// which does not set m_oob_indices.
+		if(m_oob_indices != nullptr) {
+			const bool is_within_bounds_lo = (index >= m_accessed_virtual_subrange.offset) == detail::id_cast<Dims>(id<3>(true, true, true));
+			const bool is_within_bounds_hi =
+			    (index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range)) == detail::id_cast<Dims>(id<3>(true, true, true));
```
PeterTh, line 264: Maybe move `detail::id_cast<Dims>(id<3>(true, true, true)` to a `constexpr auto all_true` for readability (and same LoC).

include/closure_hydrator.h
```diff
@@ -62,6 +62,8 @@ class closure_hydrator {
 		range<3> backing_buffer_range;
 		id<3> backing_buffer_offset;
 		subrange<3> accessed_virtual_subrange;
+		// Only used when CELERITY_ACCESSOR_BOUNDARY_CHECK=1
+		id<3>* out_of_bounds_indices = nullptr;
```
PeterTh, line 66: Why is this not just `#if`'d`?

PeterTh (commented):
I'm a huge fan of this PR. Only some minor comments on top of what psalz mentioned.

include/closure_hydrator.h
facuMH: discussed offline, we keep it here for uniform handling in tests

psalz: This is based on an earlier suggestion of mine, since I thought having this included conditionally meant that we'd have to also have to branch every time we create a `closure_hydrator::accessor_info`. However since aggregate initialization allows to only partially initialize PODs we actually don't, so it would be fine to bring back the `#if`.

psalz (approved):

PR #179, opened 2023-05-30
Make buffer manager aware of current device memory usage

psalz:
- Limit amount of device memory that can be used (currently set to 95%)
- Detect out-of-memory errors and print helpful error message listing all currently allocated buffers and their sizes
  - As it turns out DPC++ (on LevelZero at least) swaps out GPU memory to system memory, and it'll just continue work (albeit slowly)
- If resizing large buffer would exceed device memory, attempt to go through host first
- If accessing part of large buffer would result in resize that exceeds device memory, attempt to spill currently unused parts to the host

### DIFF
--- a/include/buffer_manager.h
+++ b/include/buffer_manager.h
@@ -88,7 +88,7 @@ namespace detail {
 
 		using buffer_lifecycle_callback = std::function<void(buffer_lifecycle_event, buffer_id)>;
 
-		using device_buffer_factory = std::function<std::unique_ptr<buffer_storage>(const range<3>&, sycl::queue&)>;
+		using device_buffer_factory = std::function<std::unique_ptr<buffer_storage>(const range<3>&, device_queue&)>;
 		using host_buffer_factory = std::function<std::unique_ptr<buffer_storage>(const range<3>&)>;
 
 		struct buffer_info {
@@ -136,7 +136,7 @@ namespace detail {
 				std::unique_lock lock(m_mutex);
 				bid = m_buffer_count++;
 				m_buffers.emplace(std::piecewise_construct, std::tuple{bid}, std::tuple{});
-				auto device_factory = [](const celerity::range<3>& r, sycl::queue& q) {
+				auto device_factory = [](const celerity::range<3>& r, device_queue& q) {
 					return std::make_unique<device_buffer_storage<DataT, Dims>>(range_cast<Dims>(r), q);
 				};
 				auto host_factory = [](const celerity::range<3>& r) { return std::make_unique<host_buffer_storage<DataT, Dims>>(range_cast<Dims>(r)); };
@@ -211,10 +211,19 @@ namespace detail {
 		 */
 		void set_buffer_data(buffer_id bid, const subrange<3>& sr, unique_payload_ptr in_linearized);
 
+		// Set the maximum percentage of global device memory to be used, in interval (0, 1].
+		void set_max_device_global_memory_usage(const double max) {
+			assert(max > 0 && max <= 1);
+			m_max_device_global_mem_usage = max;
+		}
+
 		template <typename DataT, int Dims>
 		access_info access_device_buffer(buffer_id bid, access_mode mode, const subrange<Dims>& sr) {
 #if defined(CELERITY_DETAIL_ENABLE_DEBUG)
-			assert((m_buffer_types.at(bid)->has_type<DataT, Dims>()));
+			{
+				std::unique_lock lock(m_mutex);
+				assert((m_buffer_types.at(bid)->has_type<DataT, Dims>()));
+			}
 #endif
 			return access_device_buffer(bid, mode, subrange_cast<3>(sr));
 		}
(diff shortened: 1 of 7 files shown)


test/buffer_manager_tests.cc
```diff
@@ -1,5 +1,6 @@
 #include "sycl_wrappers.h"
 
+#include "catch2/matchers/catch_matchers_string.hpp"
```
fknorr, line 3: Should probably use angle brackets <>.

fknorr (commented):
Some minor remarks from my side, but looks good overall.

Another naming pet-peeve of mine: I would prefer if the code explicitly talked about raw allocation sizes and alignments in terms of bytes, e.g. `size_bytes`, `assume_bytes_freed` to highlight that it doesn't refer to element counts (or some sort of allocation granularity like pages).

PeterTh (approved):
LGTM

include/device_queue.h
```diff
@@ -68,6 +93,39 @@ namespace detail {
 			return evt;
 		}
 
+		template <typename T>
+		[[nodiscard]] device_allocation malloc(const size_t count) {
+			assert(m_sycl_queue != nullptr);
+			assert(m_global_mem_allocated + count * sizeof(T) < m_global_mem_size);
```
facuMH, line 99: `count * sizeof(T)* is used 5 times, it could also be a `const auto var`

include/device_queue.h
```diff
@@ -40,6 +40,29 @@ namespace detail {
 	}
 #endif
 
+	struct device_allocation {
```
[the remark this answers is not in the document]
psalz: I think having this as a move-only type makes sense semantically but given that we don't actually move allocations right now, and this will presumably be replaced by the instruction graph anyway, I've turned it into a POD!

include/buffer_storage.h
```diff
@@ -49,16 +45,18 @@ namespace detail {
...
 
-		DataT* get_pointer() { return m_device_ptr; }
+		DataT* get_pointer() { return static_cast<DataT*>(m_device_allocation.ptr); }
 
-		const DataT* get_pointer() const { return m_device_ptr; }
+		const DataT* get_pointer() const { return static_cast<DataT*>(m_device_allocation.ptr); }
 
-		bool operator==(const device_buffer& rhs) const { return m_device_ptr == rhs.m_device_ptr && m_queue == rhs.m_queue && m_range == rhs.m_range; }
+		bool operator==(const device_buffer& rhs) const {
+			return m_device_allocation == rhs.m_device_allocation && m_queue == rhs.m_queue && m_range == rhs.m_range;
```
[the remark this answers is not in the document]
psalz: Right, removed. Not sure why we needed it at some point.

test/buffer_manager_tests.cc
psalz: The one time I decide to use the clang-tidy auto fix...

src/buffer_manager.cc
```diff
@@ -78,16 +78,77 @@ namespace detail {
...
+					if(can_allocate(sr.range.size() * element_size, existing_buf.storage->get_range().size() * element_size)) {
+						spill_to_host = true;
+					} else {
+						// TODO: Same thing as above (could evict other buffers first)
+						die(allocation_size);
+					}
+				}
+
+				if(spill_to_host) {
+					CELERITY_WARN("Buffer {} cannot be resized to fit into device memory, spilling partially to host. Performance may be degraded.", bid);
```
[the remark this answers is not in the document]
psalz: Yes, good idea. Done!

include/buffer_storage.h
```diff
@@ -49,16 +45,18 @@ namespace detail {
...
 		range<Dims> get_range() const { return m_range; }
 
-		DataT* get_pointer() { return m_device_ptr; }
+		DataT* get_pointer() { return static_cast<DataT*>(m_device_allocation.ptr); }
 
-		const DataT* get_pointer() const { return m_device_ptr; }
+		const DataT* get_pointer() const { return static_cast<DataT*>(m_device_allocation.ptr); }
 
-		bool operator==(const device_buffer& rhs) const { return m_device_ptr == rhs.m_device_ptr && m_queue == rhs.m_queue && m_range == rhs.m_range; }
+		// bool operator==(const device_buffer& rhs) const {
```
psalz, line 52: ?

fknorr, line 52: //

fknorr (approved):
LGTM!

include/buffer_manager.h
```diff
@@ -222,7 +228,10 @@ namespace detail {
 		template <typename DataT, int Dims>
 		access_info access_host_buffer(buffer_id bid, access_mode mode, const subrange<Dims>& sr) {
 #if defined(CELERITY_DETAIL_ENABLE_DEBUG)
-			assert((m_buffer_types.at(bid)->has_type<DataT, Dims>()));
+			{
```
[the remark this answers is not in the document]
psalz: Turns out `access_device_buffer` didn't lock the mutex for the same check since #162, and as there is no `access_device_buffer_impl` I decided to keep them there for symmetry (it's only for debug builds anyway).

PR #183, opened 2023-06-14
CI: Basic Performance check reporting

PeterTh:
This PR adds a CI action which checks the microbenchmark results, as discussed at the last in-person meeting.

- [x] It generates a warning if the benchmark file was not updated.
- [x] It reports any significant deviations.
- [x] It generates and uploads box plots of all the benchmark results.
- [x] Box plots with significant changes are marked (red border).
- [x] If a new benchmark is added, or a benchmark is removed, this is reported.
- [x] A new comment is only added if the benchmark csv file changed since the previous version in the same PR.


fknorr:
Very neat! Do we have any historic PRs where we significantly improved / regressed performance to calibrate the THRESHOLDs?

PeterTh:
This is now done and fully functional, as far as I can tell.

fknorr:
I'm pretty happy with this! Two minor things:

- Could we have labels, or at least units on the y axis? Even though the scale doesn't matter, it's useful to see whether we're comparing rates or times.
- We're running all (?) other CI jobs on our own runners, should we do the same for performance reporting? We could then simply include the relevant ruby environment in our docker images and improve CI performance a bit.

PeterTh:
Sure, I can add the unit on the Y axis. I believe for all these benchmarks it's currently simply execution time in nanoseconds.

I purposefully didn't run this on our own runners since it has no dependencies on our own infrastructure, and so it made more sense to me not to add another dependency to our own stuff. 
I'm not sure what you mean in terms of CI performance. Since this runs on GitHub runners, it starts at the same time as everything else, and it takes less than a minute. So it will always be done long before everything else is ready.

PeterTh:
There is now a unit indicator on the Y axis.

fknorr (approved):

ci/check-perf-impact.rb
```diff
@@ -0,0 +1,248 @@
+require 'csv'
+require 'gruff'
+require 'json'
+require 'digest'
+
+# information regarding the benchmark file
+BENCH_FN = 'ci/perf/gpuc2_bench.csv'
+MEAN_COL = 5       # column containing the mean value
```
psalz, line 8: Unused

psalz (approved):
LGTM!

ci/check-perf-impact.rb
PeterTh: Thanks, removed.

PR #185, opened 2023-06-23
Fix a sporadic lifetime issue with log capturing in unit tests

PeterTh:
I believe that this lifetime issue is what caused the sporadic failures of some CI jobs in debug builds on some recent PRs.

The problem in the original code was actually rather easy to reproduce, it happened roughly one in 5 times. I've run the test in question 100 times with this version and there were no issues.

### DIFF
--- a/test/accessor_tests.cc
+++ b/test/accessor_tests.cc
@@ -654,29 +654,34 @@ namespace detail {
 #if !CELERITY_ACCESSOR_BOUNDARY_CHECK
 		SKIP("CELERITY_ACCESSOR_BOUNDARY_CHECK=0");
 #endif
-
 		buffer<int, Dims> buff(range_cast<Dims>(range<3>{10, 20, 30}));
 		const auto accessible_sr = subrange_cast<Dims>(subrange<3>{{5, 10, 15}, {1, 2, 3}});
 		const auto oob_idx_lo = id_cast<Dims>(id<3>{1, 2, 3});
 		const auto oob_idx_hi = id_cast<Dims>(id<3>{7, 13, 25});
-		distr_queue q;
 
-		celerity::test_utils::log_capture lc;
+		// we need to be careful about the orderign of the construction and destruction
+		// of the Celerity queue and the log capturing utility here
+		std::unique_ptr<celerity::test_utils::log_capture> lc;
+		{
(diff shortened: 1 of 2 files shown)


test/accessor_tests.cc
```diff
@@ -654,29 +654,34 @@ namespace detail {
...
 		buffer<int, Dims> buff(range_cast<Dims>(range<3>{10, 20, 30}));
 		const auto accessible_sr = subrange_cast<Dims>(subrange<3>{{5, 10, 15}, {1, 2, 3}});
 		const auto oob_idx_lo = id_cast<Dims>(id<3>{1, 2, 3});
 		const auto oob_idx_hi = id_cast<Dims>(id<3>{7, 13, 25});
-		distr_queue q;
 
-		celerity::test_utils::log_capture lc;
+		// we need to be careful about the orderign of the construction and destruction
+		// of the Celerity queue and the log capturing utility here
+		std::unique_ptr<celerity::test_utils::log_capture> lc;
```
psalz, line 664: Maybe it's because my brain is pretty fried atm but from the patch alone I can't really tell what the issue is. Could you elaborate? Also I found [this](https://github.com/celerity/celerity-runtime/blob/79f97c2b2b5b1442fec2668cfbf813aa50cb0225/test/system/distr_tests.cc#L140) - possibly related?

PeterTh: In debug, messages are printed to the log from another thread(s), so we need to ensure that the queue is dead before we read the capture.

psalz: Can't we just switch to a thread-safe `spdlog::sinks::ostream_sink_mt` then?

PeterTh: I just tested this, and it does not fix the segfaults.
I believe this is because, while the sink is MT safe, adding and removing sinks might not be. But this is just an educated guess. This seems likely though, because ultimately "sinks" is just a std vector. Since we have no way to control how it is accessed within the library (at least that I can think of) I believe we need to resort to controlling the lifetimes.

psalz (approved):
I hope we can find a solution that fixes this issue in general and not just for this test case (maybe some sort of `log_capture_fixture`?), but this will have to do for now.

PR #186, opened 2023-06-26
Switch to distributed scheduling model

psalz:
As discussed today this is an early draft of the new distributed scheduling model implementation. I apologize for the state of the codebase right now, but maybe you can still take a quick look if you can spot anything major. As I'm still in the process of migrating all existing command-graph-related tests to the new architecture (with the help of @fknorr - thanks!), the codebase currently has to support both the legacy `graph_generator` and the new `distributed_graph_generator` side by side, so there is quite a bit of code duplication and hackery.

The most interesting/relevant parts to review are probably `distributed_graph_generator` (obviously), as well as maybe `buffer_transfer_manager` and the new testing infrastructure. Figuring out all data dependencies is still quite the mess, although one could argue it's slightly better because we have one level of indirection less (no "per node" data maps anymore). I hear @fknorr has some ideas of how this could be improved at some point in the future.

This PR currently includes #184.


src/buffer_transfer_manager.cc
```diff
@@ -61,24 +61,35 @@ namespace detail {
 	}
 
 	std::shared_ptr<const buffer_transfer_manager::transfer_handle> buffer_transfer_manager::await_push(const command_pkg& pkg) {
+		// NOCOMMIT BTM Should have no notion of command_pkg...
 		assert(pkg.get_command_type() == command_type::await_push);
 		const auto& data = std::get<await_push_data>(pkg.data);
 
+		GridRegion<3> expected_region;
+		for(size_t i = 0; i < data.num_subranges; ++i) {
+			expected_region = GridRegion<3>::merge(expected_region, subrange_to_grid_box(data.region[i]));
```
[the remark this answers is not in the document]
PeterTh: This warning seems a bit extreme for an HPC (rather than security critical domain) codebase.

src/distributed_graph_generator.cc
```diff
@@ -0,0 +1,608 @@
...
+		}
+	}
+}
+
+void distributed_graph_generator::set_epoch_for_new_commands(const abstract_command* const epoch_or_horizon) {
+	// both an explicit epoch command and an applied horizon can be effective epochs
+	assert(isa<epoch_command>(epoch_or_horizon) || isa<horizon_command>(epoch_or_horizon));
+
+	for(auto& [bid, bs] : m_buffer_states) {
+		// TODO this could be optimized to something like cdag.apply_horizon(node_id, horizon_cmd) with much fewer internal operations
```
PeterTh, line 528: Is this comment still relevant? (I dimly recall adding it long ago, but the underlying mechanisms were very different)

test/accessor_tests.cc
```diff
@@ -651,7 +651,7 @@ namespace detail {
 	class acc_out_of_bounds_kernel {};
 
 	TEMPLATE_TEST_CASE_METHOD_SIG(oob_fixture, "accessor reports out-of-bounds accesses", "[accessor][oob]", ((int Dims), Dims), 1, 2, 3) {
-#if !CELERITY_ACCESSOR_BOUNDARY_CHECK
+#if 1 || !CELERITY_ACCESSOR_BOUNDARY_CHECK
```
PeterTh, line 654: Does the boundary check have an interaction with distributed scheduling?

test/graph_gen_transfer_tests_2.cc
```diff
@@ -0,0 +1,223 @@
...
+		// Assert that the number of pushes / await_pushes hasn't changed
+		CHECK(dctx.query().find_all(command_type::push, node_id(1)).count() == 1);
+		CHECK(dctx.query().find_all(command_type::await_push, node_id(0)).count() == 1);
+	}
+}
+
+// NOTE: This behavior changed between master/worker and distributed scheduling; we no longer consolidate pushes.
+//       In part this is because of the way data is being tracked (on a per-command last writer basis),
+//       however importantly it also enables better communication/computation overlapping in some cases.
+//       This behavior may change again in the future!
```
PeterTh, line 86: I assume the possible future change in behaviour is why we keep the test case around?

PeterTh (commented):
I have gone through everything (that isn't already covered by the r-tree PR), but not fully in-depth. I have purposefully not commented on most things already marked as TODO or NOCOMMIT in the code, since that is obviously already known anyway.

Overall, this looks good for the alpha release to me. `distributed_graph_generator::generate_distributed_commands` has lots of excellent comments, but it's still not easy to follow because it does so much. Post-0.4 it would be nice if we managed to somehow refactor or split that up.

include/distributed_graph_generator.h
```diff
@@ -0,0 +1,152 @@
...
+using node_bitset = std::bitset<max_num_nodes>;
+
+/**
+ * write_command_state is a command_id with two bits of additional information:
+ *   - Whether the data written by this command is globally still the newest version ("fresh" or "stale")
+ *   - Whether this data has been replicated from somewhere else, i.e., we are not the original producer
+ */
+class write_command_state {
+	constexpr static int64_t mask = 1ull << 63;
+	constexpr static int64_t mask2 = 1ull << 62;
```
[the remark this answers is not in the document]
psalz: Good idea!

src/distributed_graph_generator.cc
psalz: Yeah I don't think it's relevant anymore, removed.

test/accessor_tests.cc
psalz: No this was just to avoid the segfault.

test/graph_gen_transfer_tests_2.cc
psalz: Yes, and for documentation purposes, basically.

include/distributed_graph_generator.h
```diff
@@ -0,0 +1,152 @@
...
+ *   - Whether the data written by this command is globally still the newest version ("fresh" or "stale")
+ *   - Whether this data has been replicated from somewhere else, i.e., we are not the original producer
+ */
+class write_command_state {
+	constexpr static int64_t fresh_bit = 1ll << 63;
+	constexpr static int64_t replicated_bit = 1ll << 62;
+	static_assert(sizeof(fresh_bit) == sizeof(command_id));
+
+  public:
+	constexpr write_command_state() = default;
```
fknorr, line 33: Should this zero-initialize `m_cid` so that at least none of the flags are accidentally set?

include/executor.h
```diff
@@ -43,11 +44,16 @@ namespace detail {
...
 		// TODO: Try to decouple this more.
-		executor(
-		    node_id local_nid, host_queue& h_queue, device_queue& d_queue, task_manager& tm, buffer_manager& buffer_mngr, reduction_manager& reduction_mngr);
+		executor(const size_t num_nodes, const node_id local_nid, host_queue& h_queue, device_queue& d_queue, task_manager& tm, buffer_manager& buffer_mngr,
+		    reduction_manager& reduction_mngr);
 
 		void startup();
 
+		void enqueue(command_pkg&& pkg) {
+			std::scoped_lock lk(m_command_queue_mutex);
```
fknorr, line 53: That's the first time ever I've seen a `std::scoped_lock` being used, but sure, why not! :grinning:

fknorr (commented):
Unfortunately I do not have time at the moment to review this. One thing I would like to see for the final release is `distributed_graph_generator` being renamed to `command_graph_generator` (or similar, +bikeshedding) because we have two (and in the future three) different graph representations!

include/executor.h
psalz: Look. There's too many damn lock types in the standard library 😛

include/distributed_graph_generator.h
psalz: Yes it should!

src/executor.cc
```diff
@@ -110,30 +110,17 @@ namespace detail {
...
+			if(m_jobs.size() < MAX_CONCURRENT_JOBS) {
+				// TODO: Double-buffer command queue?
+				// FIXME: Don't hold lock while calling handle_command (it's annoying b/c handling can fail)
+				std::unique_lock lk(m_command_queue_mutex);
+				if(!m_command_queue.empty()) {
+					if(!handle_command(m_command_queue.front())) {
+						// In case the command couldn't be handled, don't pop it from the queue.
+						continue;
+					}
+					m_command_queue.pop();
```
fknorr, line 122: I have no intuition whether unlocking around `handle_command` would benefit us, but it could be implemented like this:
```suggestion
				std::unique_lock lk(m_command_queue_mutex);
				if(!m_command_queue.empty()) {
					auto pkg = std::move(m_command_queue.front());
					lk.unlock();
					const auto handled = handle_command(pkg);
					lk.lock();
					if (handled) {
						m_command_queue.pop();
					} else {
						// In case the command couldn't be handled, don't pop it from the queue.
						m_command_queue.front() = std::move(pkg);
						continue;
					}
```

src/print_graph.cc
```diff
@@ -154,18 +155,23 @@ namespace detail {
...
 
-			const auto name = cmd.get_cid();
-			const auto label = get_command_label(cmd, tm, bm);
-			const auto fontcolor = colors[cmd.get_nid() % (sizeof(colors) / sizeof(char*))];
-			const auto shape = isa<task_command>(&cmd) ? "box" : "ellipse";
-			return fmt::format("{}[label=<{}> fontcolor={} shape={}];", name, label, fontcolor, shape);
+			const auto id = local_to_global_id(cmd.get_cid());
+			const auto label = get_command_label(local_nid, cmd, tm, bm);
+			const auto* const fontcolor = colors[local_nid % (sizeof(colors) / sizeof(char*))];
+			const auto* const shape = isa<task_command>(&cmd) ? "box" : "ellipse";
```
fknorr, line 173: This `const auto` -> `const auto *const` substitution is a clang-tidy thing, isn't it? Because IMO it is actually worse for legibility and consistency.

src/print_graph.cc
```diff
@@ -186,25 +192,27 @@ namespace detail {
...
 			}
 
 			// Add a dashed line to the corresponding push
-			if(const auto apcmd = dynamic_cast<const await_push_command*>(cmd)) {
-				fmt::format_to(std::back_inserter(main_dot), "{}->{}[style=dashed color=gray40];", apcmd->get_source()->get_cid(), cmd->get_cid());
-			}
+			// if(const auto apcmd = dynamic_cast<const await_push_command*>(cmd)) {
+			// 	fmt::format_to(std::back_inserter(main_dot), "{}->{}[style=dashed color=gray40];", local_to_global_id(apcmd->get_source()->get_cid()),
+			// 	    local_to_global_id(cmd->get_cid()));
+			// }
```
fknorr, line 212: This is obsolete for single-node CDAG prints, isn't it?

src/runtime.cc
```diff
@@ -229,13 +215,34 @@ namespace detail {
...
+						MPI_Recv(&size, 1, MPI_UINT64_T, static_cast<int>(i), mpi_support::TAG_PRINT_GRAPH, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+						if(size > 0) {
+							std::string graph;
+							graph.resize(size);
+							MPI_Recv(graph.data(), static_cast<int>(size), MPI_BYTE, static_cast<int>(i), mpi_support::TAG_PRINT_GRAPH, MPI_COMM_WORLD,
+							    MPI_STATUS_IGNORE);
+							graphs.push_back(std::move(graph));
+						}
+					}
+					std::this_thread::sleep_for(std::chrono::milliseconds(500)); // Avoid racing stdout
```
fknorr, line 243: What would this race with?

src/scheduler.cc
```diff
@@ -33,14 +39,13 @@ namespace detail {
 
 				utils::match(
 				    event,
-				    [this](const event_task_available& e) {
+				    [this, &serializer](const event_task_available& e) {
```
fknorr, line 42: Nit, but I feel like there's no reason to use anything except `[&]` for the arms of this `match`.

test/benchmarks.cc
```diff
@@ -135,7 +138,7 @@ TEST_CASE("benchmark task handling", "[benchmark][task]") {
 struct task_manager_benchmark_context {
 	const size_t num_nodes = 1;
 	task_manager tm{1, nullptr};
-	test_utils::mock_buffer_factory mbf{tm};
+	std::unique_ptr<test_utils::mock_buffer_factory> mbf{std::make_unique<test_utils::mock_buffer_factory>(tm)};
```
fknorr, line 141: If I'm not mistaken, this can remain a `test_utils::mock_buffer_factory` without the `std::unique_ptr` indirection.

test/distributed_graph_generator_test_utils.h
```diff
@@ -0,0 +1,560 @@
...
+
+class task_builder {
+	friend class dist_cdag_test_context;
+
+	using action = std::function<void(handler&)>;
+
+	class step {
+	  public:
+		step(dist_cdag_test_context& dctx, std::deque<action> actions) : m_actions(std::move(actions)), m_dctx(dctx) {}
+		virtual ~step() noexcept(false);
```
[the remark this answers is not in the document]
fknorr:
```suggestion
		virtual ~step() noexcept(false); // NOLINT(bugprone-exception-escape)
```

test/distributed_graph_generator_test_utils.h
```diff
@@ -0,0 +1,560 @@
...
+#include "print_graph.h"
+#include "task_manager.h"
+#include "types.h"
+
+#include "test_utils.h"
+
+using namespace celerity;
+using namespace celerity::detail;
+
+class dist_cdag_test_context;
```
fknorr, line 25: All of these definitions should be in `namespace test_utils`.

test/distributed_graph_generator_test_utils.h
```diff
@@ -0,0 +1,560 @@
...
+
+		template <typename StepT>
+		StepT chain(action a) {
+			static_assert(std::is_base_of_v<step, StepT>);
+			m_actions.push_front(std::move(a));
+			return StepT{m_dctx, std::move(m_actions)};
+		}
+	};
+
+	class buffer_access_step : public step {
```
fknorr, line 58: This class hierarchy looks like the remains of some refactoring. All the functions on `buffer_access_step` could (should?) just be members of `step` as far as I can see. This would also address the `TODO misnomer` below.

test/distributed_graph_generator_test_utils.h
```diff
@@ -0,0 +1,560 @@
...
+  public:
+	// TODO Other ideas:
+	// - remove(filters...)						=> Remove all commands matching the filters
+	// - executes(global_size)					=> Check that commands execute a given global size (exactly? at least? ...)
+	// - writes(buffer_id, subrange) 			=> Check that commands write a given buffer subrange (exactly? at least? ...)
+	// - find_one(filters...)					=> Throws if result set contains more than 1 (per node?)
+
+	template <typename... Filters>
+	command_query find_all(Filters... filters) const {
+		assert_not_empty(__FUNCTION__);
```
fknorr, line 149: What does that assertion actually do?

test/distributed_graph_generator_test_utils.h
```diff
@@ -0,0 +1,560 @@
...
+};
+
+class dist_cdag_test_context {
+	friend class task_builder;
+
+  public:
+	dist_cdag_test_context(size_t num_nodes) : m_num_nodes(num_nodes) {
+		m_rm = std::make_unique<reduction_manager>();
+		m_tm = std::make_unique<task_manager>(num_nodes, nullptr /* host_queue */);
+		// m_gser = std::make_unique<graph_serializer>(*m_cdag, m_inspector.get_cb());
```
fknorr, line 401: This can be removed I suppose.

fknorr (commented):
Without reviewing the tests:

include/distributed_graph_generator.h
```diff
@@ -0,0 +1,152 @@
...
+	bool is_replicated() const { return (m_cid & replicated_bit) != 0u; }
+
+	void mark_as_stale() { m_cid |= fresh_bit; }
+
+	operator command_id() const { return m_cid & ~fresh_bit & ~replicated_bit; }
+
+	bool operator==(const write_command_state& other) const { return m_cid == other.m_cid; }
+
+	bool operator!=(const write_command_state& other) const { return !(*this == other); }
+
```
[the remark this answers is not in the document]
psalz: Done!

include/distributed_graph_generator.h
```diff
@@ -0,0 +1,152 @@
...
+		buffer_state(region_map<write_command_state> lw, region_map<std::bitset<max_num_nodes>> rr)
+		    : local_last_writer(std::move(lw)), replicated_regions(std::move(rr)), pending_reduction(std::nullopt) {}
+
+		region_map<write_command_state> local_last_writer;
+		region_map<node_bitset> replicated_regions;
+
+		// When a buffer is used as the output of a reduction, we do not insert reduction_commands right away,
+		// but mark it as having a pending reduction. The final reduction will then be generated when the buffer
+		// is used in a subsequent read requirement. This avoids generating unnecessary reduction commands.
+		std::optional<reduction_info> pending_reduction;
```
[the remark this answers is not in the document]
psalz: Well as you noted we actually do need to access the `local_last_writer` still, so that would then have had to either be replicated into the `pending_reduction_state` or moved to a separate map altogether. Other than that (and implementation order, as you note), I don't think so. In any case, this code isn't going to last anyway.

src/executor.cc
psalz: Thanks, I've added it. It's probably not going to do much, but I still think it's cleaner that way.

src/print_graph.cc
psalz: We'll remove this check, as per in-person discussion yesterday. I won't go back and undo all of these, though.

src/print_graph.cc
psalz: Well, kind of. At least its a lot more difficult to implement now. But I'm also not sure whether it's all that useful anyway, in particular as it makes layouting the graphs a lot more convoluted.

src/runtime.cc
psalz: Shutdown messages from other nodes. This was added because it actually happened quite frequently.

src/scheduler.cc
psalz: Well I was simply following precedent (aa0239bbf) but sure, changed 😛

test/benchmarks.cc
psalz: These need to be uniform across the different contexts so that the benchmarking code can always use `->` for member access!

test/distributed_graph_generator_test_utils.h
psalz: Agreed!

test/distributed_graph_generator_test_utils.h
psalz: Yes, good point!

test/distributed_graph_generator_test_utils.h
psalz: Calling for example `query_a.have_type(command_type::execution)` doesn't make much sense if `query_a` does not contain any commands. What should it return, true or false? I decided it'd be best to just throw in that case. For `find_all` it's actually somewhat debatable whether it should be allowed to further "narrow down" an already empty set of commands (maybe in some generic context it would make sense), but I decided to err on the side of caution for now.

test/distributed_graph_generator_test_utils.h
psalz: Whoops, yes.

include/distributed_graph_generator.h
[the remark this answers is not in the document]
fknorr: Sure, let's address this in the IDAG reduction overhaul.

src/print_graph.cc
fknorr: Fair enough!

src/runtime.cc
fknorr: So the race is happening in mpirun due to line buffering? We could mention that in the comment.

test/benchmarks.cc
fknorr: AFAICS the `mbf`s of all benchmark contexts can be values.

test/distributed_graph_generator_test_utils.h
fknorr: Ah, this is a member function. I assumed this would do something akin to `assert(__FUNCTION__ != "")`

src/scheduler.cc
fknorr: Whoops!

src/runtime.cc
psalz: Done!

PR #188, opened 2023-07-11
Update mimalloc to version 1.8.2

PeterTh:
Among other things, this gets rid of some warnings in mimalloc when compiling with dpcpp.


PeterTh:
Note: this is moving from 1.8.0 to 1.8.2, so there shouldn't be user-facing changes other than bugfixes.

PeterTh:
Average relative microbenchmark perf compared to master is 0.957. 
(I didn't want to introduce a dependency on the perf reporting commit so I did this manually one last time)

psalz (approved):

fknorr (approved):

PR #182, opened 2023-06-13
Update website and documentation for version 0.4.0

PeterTh:
This should be merged once distributed scheduling is in, before the 0.4 release.

PR #193, opened 2023-07-13
RFC: Implement access::components higher-order range mapper

fknorr:
RFC, based on #182.

A common parallelization pattern, e.g. in dense matrix-vector products, is to map 1D thread-ids to the rows of a 2D matrix while iterating over all columns on each item. This currently requires a custom range mapper:
```cpp
const auto rows = [](const chunk<1> &chnk, const range<2> &buffer_size) {
    return subrange<2>{
        {chnk.offset[0], 0},
        {chnk.range[0], buffer_size[1]}
    };
};
```
While we could just have `access::rows` and `access::columns` convenience range mappers with all the confusion about row-major vs column-major vs fastest-dimensions that would entail, I have come up with a generic solution for the entire class of these "component mappings".

Let me introduce the first **range-mapper combinator**, `access::components`. It constructs a `(chunk<KernelDims>, range<BufferDims>) -> subrange<BufferDims>` range mapper from `BufferDims` individual mappers of type `(chunk<KernelDim>, range<1>) -> subrange<1>`.

Together with the new, straight-forward mapper `access::kernel_dim` that creates a `subrange<1>` from a single kernel dimension, this allows us to re-write the custom range mapper above like so:
```cpp
const auto rows = access::components(access::kernel_dim(0), access::all());
```
Note that any other range mapper that produces `subrange<1>` can be used for each component, such as `fixed`.


test/runtime_tests.cc
```diff
@@ -185,6 +185,51 @@ namespace detail {
...
+		}
+		{
+			range_mapper rm{kernel_dim(1), cl::sycl::access::mode::read, range<1>{128}};
+			auto sr = rm.map_1(chunk<3>{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}});
+			CHECK(sr.offset == id<1>{2});
+			CHECK(sr.range == range<1>{5});
+		}
+	}
+
+	TEST_CASE("shuffle built-in range mapper behaves as expected", "[range-mapper]") {
```
PeterTh, line 204: `shuffle` -> `components`?

test/runtime_tests.cc
```diff
@@ -185,6 +185,51 @@ namespace detail {
 		}
 	}
 
+	TEST_CASE("dim built-in range mapper behaves as expected", "[range-mapper]") {
```
PeterTh, line 188: `dim` -> `kernel_dim`?

PeterTh (commented):
Interesting. I'm not sure about the `kernel_dim` name, but I do like `components`.

fknorr:
Live bikeshedding update about the naming of `access::kernel_dim`:

- All other built-in range mapperns describe a *pattern* while `kernel_dim` describes a *function*. That works within the `components` "DSL" but looks out-of-place as a stand-alone RM.
- @PeterTh suggested we could extend `one_to_one` to optionally take a `dim` parameter and then act as `chunk<Dims> -> subrange<1>`, because on a high level we think of the `rows` RM from the OP to act like a `one_to_one` in the first dimension and like an `all` in the second one. However this breaks again when regarding this specialization as a standalone RM, where a `chunk<2> -> subrange<1>` with that definition is an **n-to-one**, not a one-to-one.

PR #196, opened 2023-07-20
Fix behaviour of fence and horizons in dry runs

PeterTh:
Also added a unit test.

Note that I very minorly changed the existing dry run test: it was the only test to still use `test_utils::set_test_env`. Changing this to use the equivalent `libenvpp` functionality allowed me to get rid of the entire `set_test_env` test utility (which even had platform-specific code), making this a -20 net LoC change. Yay!

**Edit**:
Also fixed processing of horizons in dry runs and added a test for that.

### DIFF
--- a/src/scheduler.cc
+++ b/src/scheduler.cc
@@ -19,7 +19,14 @@ namespace detail {
 
 	void abstract_scheduler::schedule() {
 		graph_serializer serializer([this](command_pkg&& pkg) {
-			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch) { return; }
+			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch && pkg.get_command_type() != command_type::fence) {
+				// in dry runs, skip everything except epochs and fences
+				return;
+			}
+			if(m_is_dry_run && pkg.get_command_type() == command_type::fence) {
+				CELERITY_WARN("Encountered a \"fence\" command while \"CELERITY_DRY_RUN_NODES\" is set. "
+				              "The result of this operation will not match the expected output of an actual run.");
+			}
 			// Executor may not be set during tests / benchmarks
 			if(m_exec != nullptr) { m_exec->enqueue(std::move(pkg)); }
 		});
--- a/src/scheduler.cc
+++ b/src/scheduler.cc
@@ -19,8 +19,9 @@ namespace detail {
 
 	void abstract_scheduler::schedule() {
 		graph_serializer serializer([this](command_pkg&& pkg) {
-			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch && pkg.get_command_type() != command_type::fence) {
-				// in dry runs, skip everything except epochs and fences
+			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch && pkg.get_command_type() != command_type::horizon
+			    && pkg.get_command_type() != command_type::fence) {
+				// in dry runs, skip everything except epochs, horizons and fences
 				return;
 			}
 			if(m_is_dry_run && pkg.get_command_type() == command_type::fence) {
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -1188,9 +1188,10 @@ namespace detail {
 
 #endif
 
+	const std::string dryrun_envvar_name = "CELERITY_DRY_RUN_NODES";
+
 	void dry_run_with_nodes(const size_t num_nodes) {
-		const std::string dryrun_envvar_name = "CELERITY_DRY_RUN_NODES";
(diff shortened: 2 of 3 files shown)


PeterTh:
(Note: this is currently failing due to our CI rework; will only merge after all of that is resolved)

fknorr (approved):
LGTM! Also, libenvpp is :fire:.

test/runtime_tests.cc
```diff
@@ -1228,6 +1229,27 @@ namespace detail {
 		dry_run_with_nodes(num_nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run proceeds and warns on fences", "[dryrun]") {
```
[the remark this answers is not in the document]
PeterTh: Right, I was planning to do that when I wrote the test name but that would mean using the log-inspection infrastructure which is currently a bit wonky, so I thought it's not sufficiently important.

Will change the name to reflect that.

src/scheduler.cc
```diff
@@ -19,7 +19,15 @@ namespace detail {
...
 	void abstract_scheduler::schedule() {
 		graph_serializer serializer([this](command_pkg&& pkg) {
-			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch) { return; }
+			if(m_is_dry_run && pkg.get_command_type() != command_type::epoch && pkg.get_command_type() != command_type::horizon
+			    && pkg.get_command_type() != command_type::fence) {
+				// in dry runs, skip everything except epochs, horizons and fences
+				return;
+			}
+			if(m_is_dry_run && pkg.get_command_type() == command_type::fence) {
+				CELERITY_WARN("Encountered a \"fence\" command while \"CELERITY_DRY_RUN_NODES\" is set. "
```
psalz, line 28: Should we print this every time a fence is encountered or just once? In general I think we might want to have some "warn once" utility.

test/runtime_tests.cc
```diff
@@ -1228,6 +1229,55 @@ namespace detail {
 		dry_run_with_nodes(num_nodes);
 	}
 
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run proceeds on fences", "[dryrun]") {
```
psalz, line 1232:
```suggestion
	TEST_CASE_METHOD(test_utils::runtime_fixture, "dry run proceeds on fences", "[dryrun]") {
```

test/runtime_tests.cc
```diff
@@ -1228,6 +1229,55 @@ namespace detail {
...
+			accessor acc{buf, cgh, all{}, write_only_host_task};
+			cgh.host_task(on_master_node, [=] { acc = true; });
+		});
+
+		auto ret = experimental::fence(q, buf);
+		bool val = *ret.get(); // this will hang if fences are not processed in the dry run
+		CHECK_FALSE(val);      // extra check that the task was not actually executed
+	}
+
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "Dry run processes horizons", "[dryrun]") {
```
psalz, line 1253:
```suggestion
	TEST_CASE_METHOD(test_utils::runtime_fixture, "dry run processes horizons", "[dryrun]") {
```

test/runtime_tests.cc
```diff
@@ -1228,6 +1229,55 @@ namespace detail {
...
+
+		REQUIRE(rt.is_dry_run());
+
+		buffer<bool, 0> buf{false};
+		q.submit([&](handler& cgh) {
+			accessor acc{buf, cgh, all{}, write_only_host_task};
+			cgh.host_task(on_master_node, [=] { acc = true; });
+		});
+
+		auto ret = experimental::fence(q, buf);
```
psalz, line 1248: Maybe add a TODO here that we should check for the warning being printed once we've sorted out the issues with `log_capture`.

test/runtime_tests.cc
```diff
@@ -1228,6 +1229,55 @@ namespace detail {
...
+		bool val = *ret.get(); // this will hang if fences are not processed in the dry run
```
psalz, line 1249:
```suggestion
                REQUIRE(ret.wait_for(std::chrono::seconds(1)) == std::future_status::ready);
		const bool val = *ret.get();
```

psalz (changes requested):
Good change!

src/scheduler.cc
PeterTh: I agree that a warn once utility would be nice, but I think it should probably be a separate PR.

test/runtime_tests.cc
PeterTh: Good idea, done!

psalz (approved):

PR #197, opened 2023-07-27
Full recording of tasks and commands on request (e.g. for graph printing)

PeterTh:
This was supposed to be a small thing I'll quickly finish before doing what I actually want to do. Oh well.

Results of some in-person discussions during development:

* We want the recorded information to be standalone, simple data, without any dependencies at the point in time when it is printed.
* Specifically, everything required for the output needs to be resolved when a task or command is recorded.
* This is sufficiently important for us to accept that we will have some data structure code duplication in the storage data structures (which are completely independent from the data structures used during the actual execution).
* Recording can be turned on with an environment variable. We do not believe that there will be any measurable performance impact while it is turned off (basically, one additional correctly predicted branch per command/task).

I still want to extend test coverage a bit before making the full PR, I noticed that a few things aren't covered at all during development.

*Edit*:
Added the additional tests and squashed into meaningful commit chunks. Ready for review.


include/print_graph.h
```diff
@@ -13,8 +15,119 @@ namespace detail {
...
+	};
+
+	// Task recording
+
+	using task_dependency_list = std::vector<dependency_record<task_id>>;
+
+	struct task_printing_information {
+		task_printing_information(const task& from, const buffer_manager* buff_man);
+
+		const task_id m_tid;
```
fknorr, line 50: public (struct) members should not have the `m_` prefix.

include/print_graph.h
```diff
@@ -13,8 +15,119 @@ namespace detail {
...
+	};
+
+	const std::optional<task_recorder> no_task_recorder = {};
+
+	// Command recording
+
+	using command_dependency_list = std::vector<dependency_record<command_id>>;
+
+	struct command_printing_information {
+		const command_id m_cid;
```
fknorr, line 83: public (struct) members should not have the `m_` prefix.

src/print_graph.cc
```diff
@@ -11,6 +11,208 @@
 namespace celerity {
 namespace detail {
 
+	namespace {
```
fknorr, line 14: Why the anonymous namespace? The names are already hidden from the public in `celerity::detail`. If you feel like this should be scoped further, I would go for something explicit like `celerity::detail::print_graph_detail`.

Also let's try to avoid nesting namespace declarations because it will cause unnecessary indentation with our current clang-format config.

src/runtime.cc
```diff
@@ -150,10 +150,14 @@ namespace detail {
 
 		m_reduction_mngr = std::make_unique<reduction_manager>();
 		m_host_object_mngr = std::make_unique<host_object_manager>();
-		m_task_mngr = std::make_unique<task_manager>(m_num_nodes, m_h_queue.get());
+		std::optional<task_recorder> t_rec;
+		if(m_cfg->is_recording()) t_rec = task_recorder{m_buffer_mngr.get()};
```
fknorr, line 154: Should this also ask for `spdlog::should_log(log_level::trace)`?

src/runtime.cc
```diff
@@ -267,6 +241,33 @@ namespace detail {
 
 	host_object_manager& runtime::get_host_object_manager() const { return *m_host_object_mngr; }
 
+	std::string runtime::gather_command_graph() const {
+		const auto graph_str = m_schdlr->print_command_graph();
+
+		// Send local graph to rank 0
+		if(m_local_nid != 0) {
+			const uint64_t size = graph_str.size();
```
fknorr, line 249: The size field can be an `int` because we can only send `int`-sized payload anyway.

fknorr (commented):
I like it!

One design decision that strikes me as odd is having the task manager / graph generator _own_ the recorder. IMO we would have better separation of concerns if these classes received a pointer to an existing recorder, which would also allow us to get rid of both `print_graph` member functions.

src/runtime.cc
PeterTh: No, the idea is that you might want to record for other reasons than writing to the log file (e.g. in a runtime unit test).

src/runtime.cc
PeterTh: Done!

PeterTh:
Review comments addressed.

This includes a pretty large (in terms of LOC) change to all unit tests working directly with tasks, due to the introduction of a new helper context, which allows us to throw away a lot of boilerplate.

test/runtime_tests.cc
```diff
@@ -230,7 +230,7 @@ namespace detail {
 	}
 
 	TEST_CASE("task_manager invokes callback upon task creation", "[task_manager]") {
-		task_manager tm{1, nullptr};
+		task_manager tm{1, nullptr, {}};
```
psalz, line 233: Since this is no longer a `std::optional` I'd suggest to change this for consistency (here and below).
```suggestion
		task_manager tm{1, nullptr, nullptr};
```

test/runtime_tests.cc
```diff
@@ -1228,19 +1228,18 @@ namespace detail {
 	}
 
 	TEST_CASE_METHOD(test_utils::mpi_fixture, "Config reads environment variables correctly", "[env-vars][config]") {
-		std::unordered_map<std::string, std::string> valid_test_env_vars{
-		    {"CELERITY_LOG_LEVEL", "debug"},
-		    {"CELERITY_GRAPH_PRINT_MAX_VERTS", "1"},
+		const auto var = std::string("debug");
```
psalz, line 1231: ?

test/system/distr_tests.cc
```diff
@@ -371,5 +372,66 @@ namespace detail {
...
+
+		const auto graph = celerity::detail::runtime::get_instance().gather_command_graph();
+
+		// only check on master node, since that is where the graph is gathered
+		int global_rank = 0;
+		MPI_Comm_rank(MPI_COMM_WORLD, &global_rank);
+		if(global_rank == 0) {
+			// Smoke test: It is valid for the dot output to change with updates to graph generation. If this test fails, verify that the printed graphs are
+			// sane and complete, and if so, replace the `expected` values with the new dot graph.
+			const std::string expected =
```
psalz, line 407: Since these are annoying to update, couldn't we just check for something like `StringContains("<b>execution</b> [[0,0,0] - [8,16,1]]") && StringContains("<b>execution</b> [[8,0,0] - [16,16,1]]")`?

psalz (changes requested):
Good stuff!

include/distributed_graph_generator.h
```diff
@@ -5,6 +5,7 @@
 #include <variant>
 
 #include "command_graph.h"
+#include "print_graph.h"
```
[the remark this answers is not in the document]
PeterTh: Done.

include/print_graph.h
```diff
@@ -13,8 +15,115 @@ namespace detail {
...
+		IdType node;
+		dependency_kind kind;
+		dependency_origin origin;
+	};
+
+	// Task recording
+
+	using task_dependency_list = std::vector<dependency_record<task_id>>;
+
+	struct task_printing_information {
```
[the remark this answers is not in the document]
PeterTh: Good question - done!

test/buffer_manager_tests.cc
```diff
@@ -48,7 +48,6 @@ namespace detail {
 		REQUIRE_FALSE(bm.has_buffer(b_id));
 
 		// TODO: check whether error was printed or not
```
psalz, line 50:
```suggestion
```

test/runtime_tests.cc
PeterTh: Yep.

test/system/distr_tests.cc
PeterTh: Since this is the only test that actually checks the graph composition from multiple sources, I'd like to keep it complete.

include/print_graph.h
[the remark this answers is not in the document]
PeterTh: Yes we should. Done.

include/task_manager.h
```diff
@@ -9,6 +9,7 @@
 
 #include "handler.h"
 #include "host_queue.h"
+#include "print_graph.h"
```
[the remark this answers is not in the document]
PeterTh: Done.

README.md
```diff
@@ -122,7 +122,7 @@ Celerity's runtime behavior:
   automatically assign a unique device to each worker on a host.
 - `CELERITY_PROFILE_KERNEL` controls whether SYCL queue profiling information
   should be queried (currently not supported when using hipSYCL).
-- `CELERITY_GRAPH_PRINT_MAX_VERTS` sets the maximum number of vertices the
-  task/command graphs can have above which their GraphViz output will be omitted.
+- `CELERITY_RECORDING` enables recording of the generated tasks and commands,
```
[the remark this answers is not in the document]
PeterTh: I thought about this, but to me `CELERITY_RECORDING` is a bit more intuitive in this use case, since it makes it clearer (at least in my mind) that I'm just toggling something on and off. `CELERITY_RECORD` could also specify e.g. what to record, or where to record it.

src/recorders.cc
```diff
@@ -0,0 +1,206 @@
...
+	return std::regex_replace(simplified, std::regex(">"), "&gt;");
+}
+
+task_record::task_record(const task& from, const buffer_manager* buff_mngr)
+    : tid(from.get_id()), debug_name(simplify_and_escape_name(from.get_debug_name())), cgid(from.get_collective_group_id()), type(from.get_type()),
+      geometry(from.get_geometry()), reductions(build_reduction_list(from, buff_mngr)), accesses(build_access_list(from, buff_mngr)),
+      side_effect_map(from.get_side_effect_map()), dependencies(build_task_dependency_list(from)) {}
+
+void task_recorder::record_task(const task& tsk) {
+	CELERITY_TRACE("Recording task {}", tsk.get_id());
```
psalz, line 60: Should we really log every time a task / command is being recorded? This effectively doubles the number of messages we produce.

psalz (approved):
LGTM

src/recorders.cc
PeterTh: Hmm, good point. It's only `TRACE` so I didn't think much about it, but we also currently log more useful stuff in `TRACE`.

PeterTh: Resolved by removing the tracing. It's not really valuable outside of developing the recording infrastructure, which should hopefully be simple enough so that it's hard to break in a way which needs such a crude debugging tool.

fknorr (approved):

test/distributed_graph_generator_test_utils.h
```diff
@@ -493,14 +500,10 @@ class dist_cdag_test_context {
 
 	void maybe_print_graphs() {
 		if(test_utils::print_graphs) {
-			test_utils::maybe_print_graph(*m_tm);
-
+			print_task_graph();
```
fknorr, line 503: This is currently a no-op.

```suggestion
			CELERITY_INFO("Task graph:\n\n{}\n", print_task_graph());
```

IMO it is suboptimal that "printing" refers both to "pure" stringification and loggin.

fknorr (changes requested):
Something I missed in the initial review:

src/utils.cc
```diff
@@ -0,0 +1,45 @@
+#include <utils.h>
```
psalz, line 1:
```suggestion
#include "utils.h"
```

psalz (approved):
Very nice that the task naming thing worked out so well in the end!

test/utils_tests.cc
```diff
@@ -0,0 +1,26 @@
...
+	CHECK(simplify_task_name(simple) == "name");
+
+	const string namespaced = "ns::another::name2";
+	CHECK(simplify_task_name(namespaced) == "name2");
+
+	const string templated = "name3<class, int>";
+	CHECK(simplify_task_name(templated) == "name3<...>");
+
+	const string real = "set_identity<float>(celerity::distr_queue, celerity::buffer<float, 2>, "
+	                    "bool)::{lambda(celerity::handler&)#1}::operator()(celerity::handler&) const::set_identity_kernel<const celerity::buffer&>";
```
fknorr, line 19: This probably wont' work for MSVC :thinking:

PeterTh: Yeah, I'm aware, but I think we have to fix that once we have some form of (at least build-only) MSVC CI.

src/utils.cc
PeterTh: Done.

test/utils_tests.cc
```diff
@@ -0,0 +1,26 @@
+#include <utils.h>
```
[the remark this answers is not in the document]
PeterTh: Done.

PR #199, opened 2023-08-03
Implement breadth-triggered Horizons

PeterTh:
This implements Horizons which are triggered when the breadth of the execution front exceeds a given limit, which can be configured at runtime (currently defaults to `64`).

This also makes it impossible to run out of task slots in normal usage (i.e. unless the user manually sets the limit to a value that exceeds the task ring buffer capacity).

This PR includes #196 and #197 which should be merged first, and is actually rather small by itself.
I will rebase it once those are merged.


include/host_utils.h
```diff
@@ -0,0 +1,41 @@
...
+				for(size_t d2 = 0; d2 < part.get_subrange().range[2]; ++d2) {
+					func(detail::make_item<3>(id<3>(d0, d1, d2) + part.get_subrange().offset, part.get_subrange().offset, part.get_global_size()));
+				}
+			}
+		}
+	}
+}
+
+template <int Dims, typename Functor>
+void for_each_item(const range<Dims> range, Functor func) {
```
fknorr, line 36: This allows the user to summon the "magic" type `celerity::item` for the first time. Could we alternatively iterate over `id`s instead of items, or if items are necessary for this to work, stage the change through `namespace celerity::experimental`?

include/intrusive_graph.h
```diff
@@ -36,6 +36,7 @@ namespace detail {
 
 		auto& front() const { return *m_first; }
 		bool empty() const { return m_first == m_last; }
+		size_t size() const { return m_last - m_first; }
```
fknorr, line 39: Since this type has so far been well-defined for any iterator, `size()` should use `std::distance` instead of `operator-`.

include/ranges.h
```diff
@@ -223,11 +223,10 @@ id<DimsOut> id_cast(const coordinate<InterfaceIn, DimsIn>& in) {
...
 
-struct zero_range_t {
-} inline static constexpr zero_range;
-struct unit_range_t {
-} inline static constexpr unit_range;
-
+struct zero_t {
+} inline static constexpr zero;
+struct one_t {
+} inline static constexpr one;
```
fknorr, line 229: I'm not entirely happy with this renaming. Without knowing the definition one could assume that `id(one)` refers to `[0,0,1]` instead of `[1,1,1]`. `zero` is unambiguous in that regard but from the name alone it's unclear that it is only used to construct ranges or ids.

How about `zero_coord` and `unit_coord`, or at least `zeros` and `ones`?

examples/independent_tasks/independent_tasks.cc
```diff
@@ -0,0 +1,53 @@
+#include <celerity.h>
```
fknorr, line 1: Any background on why this is a new example? What user-facing feature of Celerity does it demonstrate? To me it looks like more of a `runtime_test`.

PeterTh: It's an example because I needed the ability to easily invoke it standalone with different arguments for the paper extension I'm writing ;)

Which is obviously not a good reason, I'll change that to something else.

include/ranges.h
PeterTh: I like `zeros` and `ones`, I'll do that.

include/host_utils.h
PeterTh: I'm not sure I understand the problem here: it iterates over `item`s because that's what `parallel_for` does and the idea is for this to have the equivalent signature.

include/intrusive_graph.h
PeterTh: Done!

include/host_utils.h
fknorr: Fom the user's perspective, so far the only way to create a `celerity::item` was to launch a kernel. `nd_item` is a "magic type" because of its synchronization functionality, and although `item` is effectively just a struct, I would categorize it to be "equally magic" and something only the runtime should be able to create.

In contrast, the `range` overload of `for_each_item` allows users to create `item`s wherever they want. This is not a problem for the `partition` overload, because a `partition` is not user-constructible.

include/ranges.h
PeterTh: Done.

include/host_utils.h
PeterTh: I see what you mean now, but I don't feel like having this new signature makes a practical difference: the user still isn't creating the item, it's just handed to the functor they provide by a Celerity function.  
And I don't think we are introducing any new significant constraints on what a `celerity::item` can be, since its already part of an external interface (arguable of *the* most important external interface in Celerity).

The reason I'd like to keep this interface the same is that this way, it is a 1:1 match with `parallel_for`, which I feel is more natural than iterating over something else here.

examples/independent_tasks/independent_tasks.cc
PeterTh: This is now a new benchmark (of a new type, since we didn't really have full system benchmarks so far, so in a new file).

PeterTh:
This should now be ready for re-review.

include/host_utils.h
psalz: > [or] stage the change through namespace celerity::experimental

I think this would be a good idea regardless, until we have a real-world use case for it.

include/host_utils.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "host_queue.h"
+#include "item.h"
+
+namespace celerity {
+
+template <int Dims, typename Functor>
+void for_each_item(const partition<Dims> part, Functor func) {
```
psalz, line 9: Since this is a bit of a drive-by API addition, could you add it to the (not yet existing) `[Unreleased]` section of the changelog, so we don't forget about it?

include/ranges.h
```diff
@@ -422,7 +429,7 @@ namespace detail {
 	  public:
 		subscript_proxy(Target& tgt, const id<TargetDims> id) : m_tgt(tgt), m_id(id) {}
 
-		inline decltype(auto) operator[](const size_t index) const {
+		inline decltype(auto) operator[](const size_t index) const { //
```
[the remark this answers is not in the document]
PeterTh: I updated my clangd, and it decided to format this differently. Since at some point we'll probably want to update, I went with this, since it causes both the CI version and my current (more up to date) version to format it the same.

test/system_benchmarks.cc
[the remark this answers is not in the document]
PeterTh: Included as part of the benchmarks executable, as discussed offline.

include/host_utils.h
PeterTh: Good idea!
I'll do that and also add the 2 new env variables. (Those aren't relevant for most users, but they are still part of the interface I guess and should be documented in the full changelog)

include/ranges.h
```diff
@@ -311,6 +310,14 @@ class id : public detail::coordinate<id<Dims>, Dims> {
...
 	constexpr id(const range<Dims>& range) : coordinate(detail::make_from, range) {}
 
+	constexpr id(const detail::zeros_t /* tag */) {}
+
+	constexpr id(const detail::ones_t /* tag */) {
+		for(int d = 0; d < Dims; ++d) {
+			(*this)[d] = 1;
+		}
+	}
+
```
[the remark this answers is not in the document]
PeterTh: I like being able to explicitly say that I'm initializing to `zeros`, rather than requiring the reader to know that the default constructor does that. (In principle, I'd rather have no default constructor at all, but that often makes the type more difficult to use in some contexts)

test/range_tests.cc
```diff
@@ -158,4 +158,66 @@ TEST_CASE("0-dimensional ranges are empty types", "[range]") {
...
 
+TEST_CASE("for_each_item behaves as expected", "[host_utils]") {
+	SECTION("for 0-dimensional partitions") {
+		const auto test_partition_0d = detail::make_0d_partition();
+		int call_count = 0;
+		for_each_item(test_partition_0d, [&](const item<0> item) { call_count++; });
+		CHECK(call_count == 1);
+	}
+
+	SECTION("for 1-dimensional parititons") {
```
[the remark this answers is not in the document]
PeterTh: Good catch! Done.

include/host_utils.h
PeterTh: I've now put the new interface into `celerity::experimental`.

test/system_benchmarks.cc
```diff
@@ -0,0 +1,69 @@
...
+class bench_runtime_fixture : public test_utils::runtime_fixture {};
+
+TEMPLATE_TEST_CASE_METHOD_SIG(
+    bench_runtime_fixture, "benchmark independent task pattern with N tasks", "[benchmark][system-benchmarks][indep-tasks]", ((int N), N), 100, 1000, 5000) {
+	constexpr size_t num_tasks = N;
+	constexpr size_t num_repeats = 2;
+	constexpr size_t items_per_task = 256;
+
+#ifndef NDEBUG
+	if(N > 100) { SKIP("Skipping larger-scale benchmark in debug build to save CI time"); }
```
fknorr, line 22: Good thinking, but do we want to run benchmarks at all in debug builds?

PeterTh: Valid question -- another reason I added this code specifically was that I wanted to run the benchmark in debug for development (and, well, *debugging*) reasons, and the larger sizes would take forever. Of course for that use case you could also manually select the individual versions, but with templated test/benchmark cases that isn't particularly convenient.

fknorr: We're mixing `BENCHMARK`s and `CHECK`s in a single test case here so Catch2's `--skip-benchmarks` will probably fail the test. We could think about running the benchmark binaries with `--benchmark-samples=1` to keep the checks working. But anyway, I'm fine with doing this in future PR together with the necessary adjustments to the DAG benchmarks.

fknorr (approved):

include/ranges.h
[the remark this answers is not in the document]
psalz: Ok, I'm still not sure whether we'll ever need the `ones` variant though.

psalz (approved):

include/ranges.h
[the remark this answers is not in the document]
PeterTh: We discussed this offline, and while @psalz is still not fully convinced of the utility of the `ones` variant, we decided to keep it.

PR #200, opened 2023-08-03
Remove old workarounds; Fix build with current hipSYCL

fknorr:
- Removes the ComputeCpp CMake integration
- Removes all `#preprocessor` workarounds for ComputeCpp
- Bumps the minimum hipSYCL version to d2bd9fc7 to follow the `get_info` API change from enum parameters to type parameters (see illuhad/hipSYCL#987) and removes all workarounds for older versions of hipSYCL
- Removes the `subscript_result_t` workaround for ComputeCpp's Clang-8 buggy implementation of `decltype(auto)`
- ~~Transitions all examples to unnamed kernels (kernel names were only required for CCPP)~~

This allows building Celerity with the latest hipSYCL (The `latest` image is not tagged yet due to a CI bug, so 23.04 appears broken even though it passes on a manual trigger).

Unfortunately the DPC++ CI is still broken due to what appears to be UB in the Intel IGC we've recently upgraded to. At some point we should bisect IGC versions to find a working revision and downgrade both our Docker environments and CI runner hosts accordingly.

Partially addresses celerity/meta#20, but some non-trivial workarounds remain that were introduced because of CCPP.


examples/convolution/convolution.cc
```diff
@@ -65,7 +65,7 @@ int main(int argc, char* argv[]) {
 		celerity::accessor gauss{gaussian_mat_buf, cgh, celerity::access::all{}, celerity::read_only};
 		celerity::accessor out{image_tmp_buf, cgh, celerity::access::one_to_one{}, celerity::write_only, celerity::no_init};
 
-		cgh.parallel_for<class gaussian_blur>(celerity::range<2>(image_height, image_width), [=, fs = filter_size](celerity::item<2> item) {
```
PeterTh, line 68: Should we remove *all* kernel names in the examples? They are optional now, but they do serve some useful documentation purpose in some cases IMHO (i.e. if a kernel is called `gaussian_blur` or `sharpen` you can more quickly recognize what it's supposed to do).

PeterTh (commented):
LGTM, other than one minor remark regarding the examples.  
Always nice to get rid of a lot of obsolete code.

CMakeLists.txt
```diff
@@ -345,14 +298,10 @@ add_custom_target(
...
 
-if(CELERITY_SYCL_IMPL STREQUAL "ComputeCpp")
-  target_link_libraries(celerity_runtime PUBLIC ComputeCpp::ComputeCpp)
-endif()
-
 set(DEVICE_SOURCES "")
 if(CELERITY_SYCL_IMPL STREQUAL "hipSYCL")
   # For hipSYCL we have to pass all source files into add_sycl_to_target
-  # Don't just do it in general (yields better ComputeCpp build performance)
+  # Don't just do it in general (could yield better DPC++ build performance?)
```
psalz, line 304: Our hand-rolled `add_sycl_to_target` for DPC++ ignores `SOURCES`, so we could also just get rid of this entire block.

psalz (approved):
Feels good!

examples/convolution/convolution.cc
psalz: Good point, also useful in graph printing.

CMakeLists.txt
fknorr: Even better! Removed.

examples/convolution/convolution.cc
fknorr: Many of these example kernels are the only `submits` from a function that already aptly names them, but I agree it makes graph printing less descriptive. I dropped this change until we either come up with an infrastructure that does not rely on types for kernel names (think `celerity::debug::set_cgf_name(cgh)`) or we use C++20 `std::source_location` to name kernels based on their surrounding function.

PR #201, opened 2023-08-06
Re-implement box and region types

fknorr:
What started as academic curiosity turned into a classic month-long summertime deep-dive into a topic that is not even close to the top of our priority list. Oh well.

Currently, Celerity uses AllScale's `GridBox` and `GridRegion` types to describe point sets that originate from boxes or tilings of boxes. Out of these, `GridRegion` is particularly interesting: Internally a `vector<GridBox>`, it ensures that the tiling does not contain overlaps, and also opportunistically merges adjacent boxes when computing unions.

On the path to this PR, I set out on the following mission:

1. Concisely re-implement box and region types in a way that fits current Celerity, namely through an `int` dimensions parameter, unsigned coordinate types and support for 0-dimensional instances.
2. Investigate whether there's a better, or even optimal way of computing minimal, non-overlapping tilings for regions.
3. Optimize the implementation of union, intersection and difference operations without resorting to a full R-tree.

The user-facing API is mostly self-explanatory and similar to the old one, but the inner workings require some explanation. I have made an effort to document all algorithmic decisions as well as possible.

Finally, this PR is able to remove the entirety of AllScale from our dependencies (some 20k LOC) :partying_face: 

## Region Normalization

The key innovation in the new region implementation is _normalization_. While regions are still represented as vectors of boxes, the new region type normalizes the tiling such that any two regions that cover the same points have identical box sequences. Normalization of an arbitrary user-provided box set works as follows:

1. Remove all empty boxes
2. In all dimensions except the last ("fastest"), collect the set of box boundaries (minima and maxima) in that dimension as the set of _dissection lines_
3. Dissect all boxes along the obtained lines
4. Remove all pairwise fully-covered boxes
5. Merge connected / overlapping boxes, first along the the last ("fastest") dimension, up until the first ("slowest") dimension
6. Sort boxes according to their coordinates

The resulting representation is unique for the covered set of points and maximizes the extent along the last ("fastest") dimension while minimizing the number of total boxes. A small example in 2D:

region-normalization

All region set operations (union, intersection, difference) work on these normalized regions and take the appropriate step to ensure their result remains normalized as well.

## Effective Dimensionality

Like we saw in the implementation of the new region map, treating every region as three-dimensional for the purpose of "dimensionality erasure" can be costly. This is why this PR introduces the concept of _effective dimensionality_:

> If the last _k_ coordinates of an _n_-dimensional range are `1`, of an `id` are `0` or of a box are `(0, 1)`, its effective dimensionality is _n-k_.

Practically this means that the result of a `box_cast<3>(box<1>)` has _storage dimensionality_ 3 but retains _effective dimensionality_ 1, and it can be `box_cast` back to a 1-dimensional box without data loss.

The efficiency of region algorithms profits from knowledge about the data's effective dimensionality. Instead of true dimensionality erasure through a _pimpl_ or similar, the region algorithms will detect the effective dimensionality on the data itself and dispatch to the correct, optimal (templated) implementation.

To improve robustness, this PR turns `{range,id,chunk,subrange,box,region}_cast` into _checked_ casts, meaning that a debug build will assert that no type can be cast to a storage dimensionality below its effective dimensionality. Multiple tests were using the `cast` functions to create lower-dimensionality ranges in generic tests, for these, I have introduced `test_utils::truncate_{range,id,...}` as a replacement.

## Grid Microbenchmarks

For development, I have implemented a bunch of benchmarks on both old and new implementations (for the old impl, benchmark code is added in  `Re-implement grid data structures with normalized regions` and removed again in `Remove inclusion of old grid implementation`).

Grid Microbenchmark Results

The new impl appears to be about 4x faster on average across a wide variety of scenarios, but there are also individual instances where it is >5x _slower_. Bear in mind that this is not an apples-to-apples comparison because the resulting tilings differ between old and new impl.

## TODO

- [x] Upload grid microbenchmark results comparing new and old implementations
- [x] Move benchmarking code out of `all_tests`
- [ ] Run the runtime microbenchmark suite (once CI is no longer broken)

## Future Work

- Use the 0-dimensional region type in our new `region_map` implementation
- Return regions instead of boxes in `region_map::get_region_values`
- Investigate if `region_map` can be made to operate in the same normalized representation to guarantee determinism


fknorr:
Closing this in favor of #204 since GH Actions jobs cannot review / comment on PRs that originate from forks, which is needed for our clang-tidy / benchmark report integration.

PR #202, opened 2023-08-10
Remove/work around noisy clang-tidy checks

psalz:
1. Disable `readability-qualified-auto` (which suggests to use `const auto* const` instead of `const auto`).
2. Add a new `as<>` wrapper around `static_cast` for use with `isa<>`. Move both to `utils` namespace since they aren't exclusive to commands.

### DIFF
--- a/src/distributed_graph_generator.cc
+++ b/src/distributed_graph_generator.cc
@@ -220,7 +220,7 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
 				// we have to include it in exactly one of the per-node intermediate reductions.
 				for(const auto& reduction : tsk.get_reductions()) {
 					if(nid == reduction_initializer_nid && reduction.init_from_buffer) {
-						static_cast<execution_command*>(cmd)->set_is_reduction_initializer(true);
+						utils::as<execution_command>(cmd)->set_is_reduction_initializer(true);
 						break;
 					}
 				}
@@ -313,7 +313,7 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
 								// possibly even multiple for partially already-replicated data.
 								// TODO: Can and/or should we consolidate?
 								auto* const push_cmd = create_command<push_command>(bid, 0, nid, trid, grid_box_to_subrange(replicated_box));
-								assert(!isa<await_push_command>(m_cdag.get(wcs)) && "Attempting to push non-owned data?!");
+								assert(!utils::isa<await_push_command>(m_cdag.get(wcs)) && "Attempting to push non-owned data?!");
 								m_cdag.add_dependency(push_cmd, m_cdag.get(wcs), dependency_kind::true_dep, dependency_origin::dataflow);
 								generated_pushes.push_back(push_cmd);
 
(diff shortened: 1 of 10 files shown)


src/distributed_graph_generator.cc
```diff
@@ -450,13 +450,13 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
 			assert(writer_cmd != nullptr);
 
 			// We're only interested in writes that happen within the same task as the push
-			if(isa<task_command>(writer_cmd) && static_cast<task_command*>(writer_cmd)->get_tid() == tsk.get_id()) {
+			if(utils::isa<task_command>(writer_cmd) && utils::as<task_command>(writer_cmd)->get_tid() == tsk.get_id()) {
```
fknorr, line 453: This would be cleaner with an if-with-initializer that doesn't need to repeat the target type, i.e.
```c++
if(const auto writer_tcmd = dynamic_cast<const task_command*>(writer_cmd); writer_tcmd && writer_tcmd->get_tid() == tsk.get_id())
```
and even prettier if we had another utility `try_as`:
```c++
template <typename T, typename P>
auto try_as(P* p) {
	return dynamic_cast<std::conditional_t<std::is_const_v<P>, const T*, T*>>(p);
}
```
```c++
if(const auto writer_tcmd = utils::try_as<task_command>(writer_cmd); writer_tcmd && writer_tcmd->get_tid() == tsk.get_id())
```

src/distributed_graph_generator.cc
```diff
@@ -506,7 +506,7 @@ void distributed_graph_generator::generate_anti_dependencies(
 			auto* const cmd = d.node;
 
 			// We might have already generated new commands within the same task that also depend on this; in that case, skip it
-			if(isa<task_command>(cmd) && static_cast<task_command*>(cmd)->get_tid() == tid) continue;
+			if(utils::isa<task_command>(cmd) && utils::as<task_command>(cmd)->get_tid() == tid) continue;
```
fknorr, line 509: Same here

src/graph_serializer.cc
```diff
@@ -40,10 +40,10 @@ namespace detail {
 		const auto flush_recursive = [this, &check_tid, &flush_count](abstract_command* cmd, auto recurse) -> void {
 			(void)check_tid;
 #if defined(CELERITY_DETAIL_ENABLE_DEBUG)
-			if(isa<task_command>(cmd)) {
+			if(utils::isa<task_command>(cmd)) {
 				// Verify that all commands belong to the same task
-				assert(check_tid == task_id(-1) || check_tid == static_cast<task_command*>(cmd)->get_tid());
-				check_tid = static_cast<task_command*>(cmd)->get_tid();
+				assert(check_tid == task_id(-1) || check_tid == utils::as<task_command>(cmd)->get_tid());
+				check_tid = utils::as<task_command>(cmd)->get_tid();
```
fknorr, line 46: And here

fknorr (commented):
Generally in favor, with another opportunity to improve our internal APIs:

src/distributed_graph_generator.cc
psalz: I guess this comes down to preference, to me this doesn't look cleaner at all. Yes, it doesn't repeat the type, but instead introduces a name that is not used within the following block, while also requiring a `nullptr` check (with an implicit conversion too bool, which I'm also not a huge fan of) so the whole thing ends up being longer.

But I agree that a `try_as` function could be useful in cases where we *do* use the name in the following block. I think we have a couple of those as well (currently using `dynamic_cast`).

fknorr: Sure, this does not necessarily need to be addressed in this PR.

PR #203, opened 2023-08-14
Fix build with current DPC++

fknorr:
Currently the continuous DPC++ build is broken on three issues:

- oneAPI is incompatible with mimalloc (see oneapi-src/unified-runtime#803)
- `malloc_shared` produces host-inaccessible alloctions on Intel Arc GPUs
- Intel Arc GPUs do not support device-to-device copies.

This PR works around all of these issues in a minimally invasive way.


psalz (approved):

PeterTh (approved):
Nice!
Well, the amount of stuff which simply doesn't work on Intel right now isn't so nice, but it's very nice to be back in green CI land.

PR #206, opened 2023-08-24
WIP: Evaluate performance impact of gch::small_vector with mimalloc enabled

fknorr:
This PR (temporarily) replaces `gch::small_vector` with `std::vector` for graph dependency lists to see how much performance there is when working with mimalloc.


fknorr:
As we can see there is still a small advantage to `small_vector`, which also holds true for #201 . Since this dependency does not cause us any headaches at the moment, we'll keep it in place for now.

PR #211, opened 2023-09-07
Add out-of-bounds check for host side accessors

GagaLP:
This implements the out-of-bounds check for host side accessors.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -406,6 +406,22 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 
 	template <access_mode M = Mode>
 	inline std::enable_if_t<detail::access::mode_traits::is_producer(M), DataT&> operator[](const id<Dims>& index) const {
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		if(m_oob_indices != nullptr) {
+			const bool is_within_bounds_lo = all_true(index >= m_accessed_virtual_subrange.offset);
+			const bool is_within_bounds_hi = all_true(index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range));
+
+			if((!is_within_bounds_lo || !is_within_bounds_hi)) {
+				std::lock_guard<std::mutex> guard(m_oob_mutex);
+				for(int d = 0; d < Dims; ++d) {
+					m_oob_indices[0][d] = std::min(m_oob_indices[0][d], index[d]);
+					m_oob_indices[1][d] = std::max(m_oob_indices[1][d], index[d] + 1);
+				}
+				return m_oob_fallback_value;
+			}
+		}
+#endif
+
 		return m_host_ptr[get_linear_offset(index)];
 	}
 
@@ -521,6 +537,16 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 	// m_host_ptr must be defined *last* for it to overlap with the sequence of range and id members in the 0-dimensional case
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS DataT* m_host_ptr = nullptr;
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+	id<3>* m_oob_indices = nullptr;
+	// This mutex has to be inline static, since accessors are copyable making the mutex otherwise useless.
+	// It is a workaround until atomic_ref() can be used on m_oob_indices in c++20.
+	inline static std::mutex m_oob_mutex;
+
+	// This value (or a reference to it) is returned for all out-of-bounds accesses.
+	mutable DataT m_oob_fallback_value = DataT{};
+#endif
+
 	template <target Target = target::host_task, typename Functor>
 	accessor(ctor_internal_tag /* tag */, const buffer<DataT, Dims>& buff, handler& cgh, const Functor& rmfn) : m_virtual_buffer_range(buff.get_range()) {
 		using range_mapper = detail::range_mapper<Dims, std::decay_t<Functor>>; // decay function type to function pointer
@@ -547,6 +573,10 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 		m_backing_buffer_range = other.m_backing_buffer_range;
 		m_virtual_buffer_range = other.m_virtual_buffer_range;
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		m_oob_indices = other.m_oob_indices;
+#endif
+
 		if(detail::is_embedded_hydration_id(m_host_ptr)) {
 			if(detail::cgf_diagnostics::is_available() && detail::cgf_diagnostics::get_instance().is_checking()) {
 				detail::cgf_diagnostics::get_instance().register_accessor(detail::extract_hydration_id(m_host_ptr), target::host_task);
@@ -558,6 +588,10 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 				m_backing_buffer_offset = detail::id_cast<Dims>(info.backing_buffer_offset);
 				m_backing_buffer_range = detail::range_cast<Dims>(info.backing_buffer_range);
 				m_accessed_virtual_subrange = detail::subrange_cast<Dims>(info.accessed_virtual_subrange);
+
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+				m_oob_indices = info.out_of_bounds_indices;
+#endif
 			}
 		}
 	}
--- a/src/worker_job.cc
+++ b/src/worker_job.cc
@@ -160,7 +160,21 @@ namespace detail {
 				const auto [bid, mode] = access_map.get_nth_access(i);
 				const auto sr = access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()).get_subrange();
 				const auto info = m_buffer_mngr.access_host_buffer(bid, mode, sr);
+
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+				// oob_indices[0] contains the lower bound oob indices
+				// oob_indices[1] contains the upper bound oob indices
+				std::vector<id<3>> oob_indices{2};
+				constexpr size_t size_t_max = std::numeric_limits<size_t>::max();
+				const auto buffer_dims = m_buffer_mngr.get_buffer_info(bid).dimensions;
+				oob_indices[0] = id<3>{size_t_max, buffer_dims > 1 ? size_t_max : 0, buffer_dims == 3 ? size_t_max : 0};
+				oob_indices[1] = id<3>{0, 0, 0};
+				access_infos.push_back(
+				    closure_hydrator::accessor_info{info.ptr, info.backing_buffer_range, info.backing_buffer_offset, sr, oob_indices.data()});
+				m_oob_indices_per_accessor.push_back(std::move(oob_indices));
+#else
 				access_infos.push_back(closure_hydrator::accessor_info{info.ptr, info.backing_buffer_range, info.backing_buffer_offset, sr});
+#endif
 			}
 
 			closure_hydrator::get_instance().arm(target::host_task, std::move(access_infos));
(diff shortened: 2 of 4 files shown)


src/worker_job.cc
```diff
@@ -175,6 +187,25 @@ namespace detail {
...
+				const id<3>& oob_min = m_oob_indices_per_accessor[i][0];
+				const id<3>& oob_max = m_oob_indices_per_accessor[i][1];
+
+				if(oob_max != id<3>{1, 1, 1}) {
+					const auto& access_map = tsk->get_buffer_access_map();
+					const auto acc_sr =
+					    grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
+					const auto oob_sr = subrange<3>(oob_min, range_cast<3>(oob_max - oob_min));
+					CELERITY_ERROR("Out-of-bounds access in host kernel detected: Accessor {} for buffer {} attempted to access indices between {} which are "
+					               "outside of mapped subrange {}", i, access_map.get_nth_access(i).first, oob_sr, acc_sr);
```
PeterTh, line 204: We might want to use the buffer name here if one is available?
(this would also apply to device accesses)

test/accessor_tests.cc
```diff
@@ -684,5 +684,38 @@ namespace detail {
...
+	TEMPLATE_TEST_CASE_METHOD_SIG(oob_fixture, "host accessor reports out-of-bounds accesses", "[accessor][oob]", ((int Dims), Dims), 1, 2, 3) {
+#if !CELERITY_ACCESSOR_BOUNDARY_CHECK
+		SKIP("CELERITY_ACCESSOR_BOUNDARY_CHECK=0");
+#endif
+		buffer<int, Dims> buff(range_cast<Dims>(range<3>{10, 20, 30}));
+		const auto accessible_sr = subrange_cast<Dims>(subrange<3>{{5, 10, 15}, {1, 2, 3}});
+		const auto oob_idx_lo = id_cast<Dims>(id<3>{1, 2, 3});
+		const auto oob_idx_hi = id_cast<Dims>(id<3>{7, 13, 25});
+
+		// we need to be careful about the orderign of the construction and destruction
```
PeterTh, line 696: "ordering" typo here (probably originally from me if this is copied, and also in the other test)

PeterTh (commented):
Looks good, just 2 minor remarks.

PeterTh (approved):
LGTM

include/accessor.h
```diff
@@ -407,6 +407,25 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 
 	template <access_mode M = Mode>
 	inline std::enable_if_t<detail::access::mode_traits::is_producer(M), DataT&> operator[](const id<Dims>& index) const {
+		
+	// analogue to the device buffer. 
+	// without sycl::memory ...
```
psalz, line 412:
```suggestion
```
I think this is clear from context.

However come to think of it, we might have to use atomics on the host as well. While there is no issue between multiple chunks running concurrently (they receive different `accessor` instances), there's nothing stopping the user from doing manual multi-threading inside a host task, therefore IMO `operator[]` should be thread safe. Thoughts @PeterTh, @fknorr ?

src/worker_job.cc
```diff
@@ -160,7 +160,19 @@ namespace detail {
 				const auto [bid, mode] = access_map.get_nth_access(i);
 				const auto sr = grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
 				const auto info = m_buffer_mngr.access_host_buffer(bid, mode, sr);
-				access_infos.push_back(closure_hydrator::accessor_info{info.ptr, info.backing_buffer_range, info.backing_buffer_offset, sr});
+
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+					// for host tasks this can be a vector (RAII)
+					std::vector<id<3>> oob_idx {2};
```
psalz, line 166:
```suggestion
					std::vector<id<3>> oob_indices {2};
```

I think it's clear from context why you can use a vector here instead of doing a `sycl::malloc_shared`. Also, this is not really RAII because you're moving it into `m_oob_indices_per_accessor` a few lines further down.

I missed this in my original review of the device accessor boundary check, but this should really be called `oob_indices` (we use plural form everywhere else). Maybe you could also fix this for the device accessor?

src/worker_job.cc
```diff
@@ -175,6 +187,27 @@ namespace detail {
...
+				const id<3>& oob_max = m_oob_indices_per_accessor[i][1];
+
+				if(oob_max != id<3>{1, 1, 1}) {
+					const auto& access_map = tsk->get_buffer_access_map();
+					const auto acc_sr =
+					    grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
+					const auto oob_sr = subrange<3>(oob_min, range_cast<3>(oob_max - oob_min));
+					const auto buffer_id = access_map.get_nth_access(i).first;
+					const auto buffer_name = m_buffer_mngr.get_debug_name(buffer_id);
+					CELERITY_ERROR("Out-of-bounds access in host kernel detected: Accessor {} for buffer {} attempted to access indices between {} which are "
```
psalz, line 205:
```suggestion
					CELERITY_ERROR("Out-of-bounds access in host task detected: Accessor {} for buffer {} attempted to access indices between {} which are "
```

test/accessor_tests.cc
```diff
@@ -650,16 +650,20 @@ namespace detail {
...
 	class acc_out_of_bounds_kernel {};
 
-	TEMPLATE_TEST_CASE_METHOD_SIG(oob_fixture, "accessor reports out-of-bounds accesses", "[accessor][oob]", ((int Dims), Dims), 1, 2, 3) {
+	TEMPLATE_TEST_CASE_METHOD_SIG(oob_fixture, "device accessor reports out-of-bounds accesses", "[accessor][oob]", ((int Dims), Dims), 1, 2, 3) {
 #if !CELERITY_ACCESSOR_BOUNDARY_CHECK
 		SKIP("CELERITY_ACCESSOR_BOUNDARY_CHECK=0");
 #endif
-		buffer<int, Dims> buff(range_cast<Dims>(range<3>{10, 20, 30}));
+		buffer<int, Dims> buff_nl(range_cast<Dims>(range<3>{10, 20, 30}));
+		buffer<int, Dims> buff_nf(range_cast<Dims>(range<3>{10, 20, 30}));
```
psalz, line 658: From what I gather these variants are for unnamed/named, but I cannot figure out what the "nl" and "nf" abbreviations stand for..? 🤔

psalz (changes requested):
Good stuff!

include/accessor.h
fknorr: I agree on atomics. As a user I would deem read-accessors thread safe. Unfortunately we don't have `std::atomic_ref` in C++17, but something along the lines of `std::vector<std::unique_ptr<std::array<std::atomic<size_t>, 3>>>` should do.

include/accessor.h
```diff
@@ -522,6 +540,13 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 	// m_host_ptr must be defined *last* for it to overlap with the sequence of range and id members in the 0-dimensional case
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS DataT* m_host_ptr = nullptr;
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+	id<3>* m_oob_indices = nullptr;
+	inline static std::mutex m_mutex;
```
PeterTh, line 545: I'd call this `m_oob_mutex`, and maybe add a comment briefly documenting why it is `inline static`.

PeterTh (changes requested):

include/accessor.h
```diff
@@ -522,6 +540,16 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 	// m_host_ptr must be defined *last* for it to overlap with the sequence of range and id members in the 0-dimensional case
 	CELERITY_DETAIL_NO_UNIQUE_ADDRESS DataT* m_host_ptr = nullptr;
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+	id<3>* m_oob_indices = nullptr;
+	// This mutex has to be inline static, since accessors are copyable making the mutex otherwise useless.
+	// It is a workaround until atomic_fetch() can be used in c++20.
```
psalz, line 546:
```suggestion
	// It is a workaround until atomic_ref() can be used on m_oob_indices in c++20.
```

include/accessor.h
```diff
@@ -407,6 +407,24 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
+		
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		if(m_oob_indices != nullptr) {
+			const id<Dims> all_true = detail::id_cast<Dims>(id<3>(true, true, true));
+			const bool is_within_bounds_lo = (index >= m_accessed_virtual_subrange.offset) == all_true;
+			const bool is_within_bounds_hi = (index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range)) == all_true;
+
+			if((!is_within_bounds_lo || !is_within_bounds_hi)) {
+				for(int d = 0; d < Dims; ++d) {
+					std::lock_guard<std::mutex>guard (m_oob_mutex);
```
psalz, line 419:
```suggestion
					std::lock_guard<std::mutex> guard (m_oob_mutex);
```
Should've been caught by CI, unfortunately the check was broken (fixed in #215).

psalz (approved):
LGTM!

include/accessor.h
```diff
@@ -407,6 +407,24 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
 
 	template <access_mode M = Mode>
 	inline std::enable_if_t<detail::access::mode_traits::is_producer(M), DataT&> operator[](const id<Dims>& index) const {
+		
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+		if(m_oob_indices != nullptr) {
+			const id<Dims> all_true = detail::id_cast<Dims>(id<3>(true, true, true));
+			const bool is_within_bounds_lo = (index >= m_accessed_virtual_subrange.offset) == all_true;
+			const bool is_within_bounds_hi = (index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range)) == all_true;
```
fknorr, line 415: Since #201 is merged, we now have `celerity::detail::all_true()` for this use case!

include/accessor.h
```diff
@@ -407,6 +407,24 @@ class accessor<DataT, Dims, Mode, target::host_task> : public detail::accessor_b
...
+
+			if((!is_within_bounds_lo || !is_within_bounds_hi)) {
+				for(int d = 0; d < Dims; ++d) {
+					std::lock_guard<std::mutex> guard (m_oob_mutex);
```
fknorr, line 419: Not that this is overly performance sensitive, but the lock can be moved out of this loop.

src/worker_job.cc
```diff
@@ -160,7 +160,18 @@ namespace detail {
 				const auto [bid, mode] = access_map.get_nth_access(i);
 				const auto sr = grid_box_to_subrange(access_map.get_requirements_for_nth_access(i, tsk->get_dimensions(), data.sr, tsk->get_global_size()));
 				const auto info = m_buffer_mngr.access_host_buffer(bid, mode, sr);
-				access_infos.push_back(closure_hydrator::accessor_info{info.ptr, info.backing_buffer_range, info.backing_buffer_offset, sr});
+
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+					std::vector<id<3>> oob_indices {2};
```
fknorr, line 165: Maybe add a comment here that `oob_indices[0]` carries minima and `oob_indices[1]` carries maxima.

src/worker_job.cc
```diff
@@ -175,6 +186,27 @@ namespace detail {
...
 
+#if CELERITY_ACCESSOR_BOUNDARY_CHECK
+			const auto data = std::get<execution_data>(pkg.data);
+			auto tsk = m_task_mngr.get_task(data.tid);
+
+			for(size_t i = 0; i < m_oob_indices_per_accessor.size(); ++i) {
+				const id<3>& oob_min = m_oob_indices_per_accessor[i][0];
+				const id<3>& oob_max = m_oob_indices_per_accessor[i][1];
+
+				if(oob_max != id<3>{1, 1, 1}) {
```
fknorr, line 197: Now that I'm reading this code again (it's the same as the device accessor OOB check), I wonder why the sentinel value for the maximum is `[1,1,1]` instead of `[0,0,0]`. If the current (task) chunk receives an empty subrange on this buffer, an access to `[0,0,0]` would be illegal is not detected here without also checking the minima (which are initialized to `SIZE_MAX`).

Does this still work when we replace the `[1,1,1]` by `[0,0,0]` (also at the location where these are initialized)?

fknorr (changes requested):
A couple of small remarks:

PeterTh (approved):

fknorr (approved):
LGTM!

PR #212, opened 2023-09-12
Add experimental split constraints API

psalz:
This adds a new `experimental::constrain_split()` API that can be used to limit the ways in which a kernel can be split, similar to how ND-range kernels are constrained to be split along work group boundaries.

The implementation is rather trivial, as it mostly reuses the already existing task geometry mechanism we have. The only addition of note is that if an ND-range kernel is additionally constrained, the effective constraint will be the last common multiple `LCM(wg-size, user-constraint)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Introduce new experimental `for_each_item` utility to iterate over a celerity range (#199)
 - Add new environment variables `CELERITY_HORIZON_STEP` and `CELERITY_HORIZON_MAX_PARALLELISM` to control Horizon generation (#199)
+- Add new `experimental::constrain_split` API to limit how a kernel can be split (#?)
 
 ## Changed
 
--- a/include/handler.h
+++ b/include/handler.h
@@ -28,8 +28,34 @@
 #endif
 
 namespace celerity {
-
 class handler;
+}
+
+namespace celerity::experimental {
+
+/**
+ * Constrains the granularity at which a task's global range can be split into chunks.
+ *
+ * In some situations an output buffer access is only guaranteed to write to non-overlapping subranges
+ * if the task is split in a certain way. For example when computing the row-wise sum of a 2D matrix into
+ * a 1D vector, a split constraint is required to ensure that each element of the vector is written by
+ * exactly one chunk.
+ *
+ * Another use case is for performance optimization, for example when the creation of lots of small chunks
+ * would result in hardware under-utilization and excessive data transfers.
+ *
+ * Since ND-range parallel_for kernels are already constrained to be split with group size granularity,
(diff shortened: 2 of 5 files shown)


include/handler.h
```diff
@@ -434,6 +444,23 @@ class handler {
...
 	void extend_lifetime(std::shared_ptr<detail::lifetime_extending_state> state) { m_attached_state.emplace_back(std::move(state)); }
 
+	template <int Dims>
+	void experimental_constrain_split(const range<Dims>& constraint) {
+		assert(m_task == nullptr);
+		m_split_constraint = detail::range_cast<3>(constraint);
+	}
+
+	template <int Dims>
+	range<3> get_constrained_granularity(const range<Dims>& granularity) const {
```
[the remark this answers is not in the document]
psalz: Rebased and implemented

include/handler.h
```diff
@@ -434,6 +465,27 @@ class handler {
...
+	range<3> get_constrained_granularity(const range<Dims>& global_size, const range<Dims>& granularity) const {
+		range<3> result = detail::range_cast<3>(granularity);
+		for(int i = 0; i < Dims; ++i) {
+			const auto lcm = std::lcm(granularity[i], m_split_constraint[i]);
+			if(lcm == 0) { throw std::runtime_error("Split constraint cannot be 0"); }
+			result[i] = lcm;
+		}
+		if(global_size % detail::range_cast<Dims>(result) != range<Dims>(detail::zeros)) {
+			throw std::runtime_error(
+			    fmt::format("The split constraint {} does not evenly divide the global size {}", detail::range_cast<Dims>(result), global_size));
```
fknorr, line 484: I would make this a bit more descriptive to the user (who might not know where the exception comes from). e.g. `The effective split constraint {} does not evenly divide the kernel global size {}`.

fknorr (approved):
LGTM!

include/handler.h
psalz: Yes, good point. I've adjusted the message to only include "effective" if it's an nd-range kernel though!

fknorr: I like it!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Introduce new experimental `for_each_item` utility to iterate over a celerity range (#199)
 - Add new environment variables `CELERITY_HORIZON_STEP` and `CELERITY_HORIZON_MAX_PARALLELISM` to control Horizon generation (#199)
+- Add new `experimental::constrain_split` API to limit how a kernel can be split (#?)
```
PeterTh, line 15: You could fill the PR# in now.

PeterTh (approved):
LGTM

CHANGELOG.md
psalz: As to not trigger yet another unnecessary CI run I'll defer this!

PR #213, opened 2023-09-12
Add naming capability to tasks

GagaLP:
This adds naming to tasks similar to buffer naming.

### DIFF
--- a/include/handler.h
+++ b/include/handler.h
@@ -42,6 +42,8 @@ namespace detail {
 	void add_reduction(handler& cgh, const reduction_info& rinfo);
 	void extend_lifetime(handler& cgh, std::shared_ptr<detail::lifetime_extending_state> state);
 
+	void set_task_name(handler& cgh, const std::string& debug_name);
+
 	template <typename Name>
 	std::string kernel_debug_name() {
 		// we need to typeid a pointer, since the name is often undefined
@@ -374,6 +376,8 @@ class handler {
 	friend void detail::add_reduction(handler& cgh, const detail::reduction_info& rinfo);
 	friend void detail::extend_lifetime(handler& cgh, std::shared_ptr<detail::lifetime_extending_state> state);
 
+	friend void detail::set_task_name(handler &cgh, const std::string& debug_name);
+
 	detail::task_id m_tid;
 	detail::buffer_access_map m_access_map;
 	detail::side_effect_map m_side_effects;
@@ -383,6 +387,7 @@ class handler {
 	size_t m_num_collective_nodes;
 	detail::hydration_id m_next_accessor_hydration_id = 1;
 	std::vector<std::shared_ptr<detail::lifetime_extending_state>> m_attached_state;
+	std::optional<std::string> m_usr_def_task_name;
 
 	handler(detail::task_id tid, size_t num_collective_nodes) : m_tid(tid), m_num_collective_nodes(num_collective_nodes) {}
 
@@ -442,6 +447,8 @@ class handler {
 		}
 		m_task =
 		    detail::task::make_host_compute(m_tid, geometry, std::move(launcher), std::move(m_access_map), std::move(m_side_effects), std::move(m_reductions));
+
+		m_task->set_debug_name(m_usr_def_task_name.value_or(""));
 	}
 
 	void create_device_compute_task(detail::task_geometry geometry, std::string debug_name, std::unique_ptr<detail::command_launcher_storage_base> launcher) {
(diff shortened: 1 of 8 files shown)


test/debug_naming_tests.cc
```diff
@@ -0,0 +1,65 @@
...
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_string.hpp>
+
+#include <celerity.h>
+
+#include "test_utils.h"
+
+namespace celerity {
+namespace detail {
```
fknorr, line 14: We used to put tests into `celerity::detail` in the beginning, but new tests should be in the global namespace. I would suggest adding
```c++
using namespace celerity;
using namespace celerity::detail;
```
here instead.

test/debug_naming_tests.cc
```diff
@@ -0,0 +1,65 @@
...
+				celerity::debug::set_task_name(cgh, task_name);
+			});
+
+			const auto tid_b = test_utils::add_host_task(tt.tm, on_master_node, [&](handler& cgh) {});
+
+			const auto its_a = tt.tm.get_task(tid_a)->get_debug_name();
+			const auto its_b = tt.tm.get_task(tid_b)->get_debug_name();
+
+			CHECK_THAT(its_a, Catch::Matchers::Equals(task_name));
+			CHECK_THAT(its_b, Catch::Matchers::Equals(""));
```
fknorr, line 31:
```suggestion
			CHECK(its_a == task_name);
			CHECK(its_b.empty());
```

same for all other uses of `Catch::Matchers::Equals` in these tests.

test/debug_naming_tests.cc
```diff
@@ -0,0 +1,65 @@
...
+
+		SECTION("Host Task") {
+			const auto tid_a = test_utils::add_host_task(tt.tm, on_master_node, [&](handler& cgh) {
+				celerity::debug::set_task_name(cgh, task_name);
+			});
+
+			const auto tid_b = test_utils::add_host_task(tt.tm, on_master_node, [&](handler& cgh) {});
+
+			const auto its_a = tt.tm.get_task(tid_a)->get_debug_name();
+			const auto its_b = tt.tm.get_task(tid_b)->get_debug_name();
```
fknorr, line 28: I'm not sure what `its` stands for. Maybe we can just get rid of these variables entirely? The expressions are pretty short.

test/debug_naming_tests.cc
```diff
@@ -0,0 +1,65 @@
...
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_string.hpp>
+
+#include <celerity.h>
+
+#include "test_utils.h"
+
+namespace celerity {
+namespace detail {
+	TEST_CASE_METHOD(test_utils::runtime_fixture, "tasks debug names", "[debug]") {
```
fknorr, line 15: A more descriptive test case name like "debug names can be set and retrieved from tasks" could improve this!

include/handler.h
```diff
@@ -456,17 +463,24 @@ class handler {
 		// Note that cgf_diagnostics has a similar check, but we don't catch void side effects there.
 		if(!m_side_effects.empty()) { throw std::runtime_error{"Side effects cannot be used in device kernels"}; }
 		m_task =
-		    detail::task::make_device_compute(m_tid, geometry, std::move(launcher), std::move(m_access_map), std::move(m_reductions), std::move(debug_name));
+		    detail::task::make_device_compute(m_tid, geometry, std::move(launcher), std::move(m_access_map), std::move(m_reductions));
+
+		m_task->set_debug_name(m_debug_task_name.value_or(debug_name));
```
fknorr, line 468: Can we improve naming of the `debug_name` parameter or `m_debug_task_name` to emphasize that one is from the kernel name template parameter and one is from `set_task_name`? This line is a bit confusing otherwise

fknorr (changes requested):
Thanks! This looks good overall but could use a bit of polishing.

test/debug_naming_tests.cc
GagaLP: Good to know :+1:

test/debug_naming_tests.cc
GagaLP: fixed

test/debug_naming_tests.cc
GagaLP: you're right

include/handler.h
GagaLP: should be better now

fknorr (approved):
LGTM!

include/task.h
```diff
@@ -215,9 +216,9 @@ namespace detail {
 		}
 
 		static std::unique_ptr<task> make_device_compute(task_id tid, task_geometry geometry, std::unique_ptr<command_launcher_storage_base> launcher,
-		    buffer_access_map access_map, reduction_set reductions, std::string debug_name) {
+		    buffer_access_map access_map, reduction_set reductions) {
 			return std::unique_ptr<task>(new task(tid, task_type::device_compute, collective_group_id{}, geometry, std::move(launcher), std::move(access_map),
-			    {}, std::move(reductions), std::move(debug_name), {}, nullptr));
+			    {}, std::move(reductions), {}, {}, nullptr));
```
psalz, line 221: Couldn't we just get rid of that parameter altogether then?

test/debug_naming_tests.cc
psalz, line 1: Do we really need a separate TU just for these tests, and if so, shouldn't we also move the tests for buffers and the fallback (inferred) task names here as well?

psalz (approved):
Thanks!

psalz (approved):

PR #215, opened 2023-09-14
CI: Fix formatting check and HEAD -> latest tagging

psalz:
Both of them were broken due to unfortunate quotation mark placements. Apparently GitHub changed the evaluation behavior at some point (these used to work).

### DIFF
--- a/.github/workflows/celerity_ci.yml
+++ b/.github/workflows/celerity_ci.yml
@@ -258,7 +258,7 @@ jobs:
             echo "$unformatted"
             echo EOF
           } >> "$GITHUB_OUTPUT"
-      - uses: "celerity/ci-report-action@v6"
+      - uses: "celerity/ci-report-action@v7"
         with:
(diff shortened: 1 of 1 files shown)


PeterTh (approved):

psalz:
While the original commit did fix the outputs not being passed to subsequent steps, the annotations for unformatted files were broken. After some back and forth (damn word splitting!) everything should be back in order now.

I've also gone ahead and updated our [ci-report-action](https://github.com/celerity/ci-report-action/releases/tag/v7) to get rid of some warnings and fix a long standing bug where having unformatted files would not cause the CI run to fail.

You can see an example of a run that fails because of unformatted files [here](https://github.com/celerity/tmp-runtime-ci-test/actions/runs/6197212706/job/16825432557).
And [here](https://github.com/celerity/tmp-runtime-ci-test/actions/runs/6197311401) is an example of a run that had build warnings generated (which does not cause the CI run to fail).

fknorr (approved):

PR #216, opened 2023-09-21
Remove task_manager / buffer_manager dependency from recorders

fknorr:
This PR is a clean-up refactoring for parts of #197, and A back-port from the instruction graph branch which will remove `buffer_manager` and extend the recording / printing infrastructure further.

- Adds `buffer_recorder` for buffer-debug-name tracking (so we don't have to ask `buffer_manager`)
- Passes `buffer_recorder` into `print_task_graph` / `print_command_graph`, and `task_recorder` into `print_command_graph`, thereby removing duplicate information from records

### DIFF
--- a/include/buffer_manager.h
+++ b/include/buffer_manager.h
@@ -84,7 +84,7 @@ namespace detail {
 		friend struct buffer_manager_testspy;
 
 	  public:
-		enum class buffer_lifecycle_event { registered, unregistered };
+		enum class buffer_lifecycle_event { registered, unregistered, set_debug_name };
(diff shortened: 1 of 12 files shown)


psalz (commented):
> Passes buffer_recorder into print_task_graph / print_command_graph, and task_recorder into print_command_graph, thereby removing duplicate information from records

This is a functional change, though (for which we evidently didn't have a test case): Before it was possible the change the buffer name over time, and it would be reflected in the record, whereas now it will only use the last name!

PR #217, opened 2023-10-02
Added celerity blockchain for task divergence checking

GagaLP:
This pull request adds a divergence checking mechanism for tasks. 

It does so by periodically gathering hashes of all tasks from task_recording and comparing them. When a divergence is detected an error containing the diverged tasks and their full task record is printed like: 

```
[2023-10-02 17:31:07.784] [error] Divergence detected in task graph at index 1:

0x471b0f1db5e4b8e6 on nodes 1 
0xe9fbff654e3748e1 on nodes 0 

[2023-10-02 17:31:07.784] [error] Task record for hash 0x471b0f1db5e4b8e6:

id: 1, debug_name: task_b_4, type: device-compute, cgid: 0
geometry:
         dimensions: 2, global_size: [1,1,1], global_offset: [0,0,0], granularity: [1,1,1]
accesses: 
         bid: 0, buffer_name: , mode: R, req: {[64,0,0] - [128,1,1]}
dependencies: 
         node: 0, kind: true-dep, origin: last-epoch
```

Additionally it also includes a rudimentary deadlock detection for nodes which are stuck by printing a warning after a given amount of time (eg 10 seconds):

```
[warning] After 10 seconds of waiting nodes 1, did not move to the next task. The runtime might be stuck.
```

All of this is automatically turned on by running the program with task recording enabled.


PeterTh (changes requested):
Regarding the overall design, I mostly like it, but one thing I noticed is that there is a lot of code that's purely for testing which is in the actual "main logic" part. So far, I believe we mostly try to keep testing-only code in testing helpers, which makes it easier to see at a glance what is actually happening in a production run vs. what is done only in testing. We might want to move the testing-related functionality out of the main code.

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,251 @@
...
+		if(nodes[0] == m_local_nid) { print_task_record(check_map, m_task_recorder_window[task_num], hash); }
+	}
+}
+
+bool abstract_block_chain::check_for_divergence() {
+	add_new_hashes();
+
+	const auto [min_size, max_size] = collect_sizes();
+
+	// TODO: if one node doesn't do anything for a while print warining
```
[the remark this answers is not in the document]
GagaLP: Thanks. This TODO shouldn't even be here because it's already implemented.

include/print_utils.h
```diff
@@ -3,6 +3,8 @@
 #include "grid.h"
 #include "ranges.h"
 
+#include "intrusive_graph.h"
```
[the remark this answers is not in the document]
GagaLP: This is a leftover form an earlier version.

include/grid.h
```diff
@@ -257,6 +258,27 @@ class region {
...
+		celerity::detail::utils::hash_combine(seed, std::hash<celerity::id<Dims>>{}(r.get_min()), std::hash<celerity::id<Dims>>{}(r.get_max()));
+		return seed;
+	};
+};
+
+template <int Dims>
+struct std::hash<celerity::detail::region<Dims>> {
+	std::size_t operator()(const celerity::detail::region<Dims> r) {
+		std::size_t seed = 0;
+		for(size_t i = 0; i < r.get_boxes().size(); ++i) {
```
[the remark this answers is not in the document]
GagaLP: Agree. Makes it much cleaner without the manual indexing.

include/divergence_block_chain.h
```diff
@@ -0,0 +1,179 @@
...
+
+#include "recorders.h"
+#include <mutex>
+#include <thread>
+#include <vector>
+
+namespace celerity::detail {
+// in c++23 replace this with mdspan
+template <typename T>
+struct mpi_multidim_send_wrapper {
```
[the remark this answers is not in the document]
GagaLP: Yes. It was originally intended to be multidimensional but I decided not to over-engineer it...

PeterTh (approved):
LGTM, other than the minor documentation and formatting things discussed offline.

include/divergence_block_chain.h
```diff
@@ -0,0 +1,161 @@
...
+using divergence_map = std::unordered_map<task_hash, std::vector<node_id>>;
+
+/**
+ *  @brief This class is the base implementation for the divergence check.
+ *
+ *	It is responsible for collecting the task hashes from all nodes and checking for differences -> divergence.
+ *	When a divergence is found, the task record for the diverging task is printed and the program is terminated.
+ *  Additionally it also checks for deadlocks and prints a warning if one is detected.
+ *
+ *  The class is abstract to allow a different divergence check implementation in tests
```
fknorr, line 74: What part of the `divergence_block_chain` does not work for tests? Is it just the reporting?

fknorr (changes requested):
Thanks! This is becoming a valuable on-boarding debug feature.

I have not looked closely at the algorithm so far, but some more documentaiton / clarification on what the workflow around the state and the member functions of `abstract_block_chain` is would help a lot.

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+	void check_for_deadlock() const;
+
+	static void log_node_divergences(const divergence_map& check_map, const int task_num);
+
+	static void log_task_record(const divergence_map& check_map, const task_record& task, const task_hash hash);
+
+	void log_task_record_once(const divergence_map& check_map, const int task_num) const;
+};
+
+class divergence_sceduler {
```
psalz, line 124:
```suggestion
class divergence_scheduler {
```

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+		assert(ij[0] * m_width + ij[1] < m_data.size());
+		return m_data[ij[0] * m_width + ij[1]];
+	}
+
+	T* data() { return m_data.data(); }
+
+  private:
+	std::vector<T> m_data;
+	const size_t m_width;
+};
```
psalz, line 30:
```suggestion
/**
 * Stores the hashes of tasks for each node.
 * The data is stored densely so it can easily be exchanged through MPI collective operations.
 */
struct per_node_task_hashes {
  public:
	per_node_task_hashes(const size_t max_hash_count, const size_t num_nodes) : m_data(max_hash_count * num_nodes), m_max_hash_count(max_hash_count){};
	const task_hash& operator()(const node_id nid, const size_t i) const { return m_data.at(nid * m_max_hash_count + i); }
	task_hash* data() { return m_data.data(); }

  private:
	std::vector<task_hash> m_data;
	size_t m_max_hash_count;
};
```

No need to go overly generic on this if it's only being used in one place. For clarity I'd suggest something like this instead.

psalz: Also move into `divergence_block_chain_detail` namespace!

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+ *	It is responsible for collecting the task hashes from all nodes and checking for differences -> divergence.
+ *	When a divergence is found, the task record for the diverging task is printed and the program is terminated.
+ *  Additionally it also checks for deadlocks and prints a warning if one is detected.
+ */
+
+class divergence_block_chain {
+	friend struct divergence_block_chain_testspy;
+
+  public:
+	divergence_block_chain(const std::vector<task_record>& task_recorder, std::unique_ptr<communicator> comm)
```
psalz, line 80:
```suggestion
	divergence_block_chain(const std::vector<task_record>* task_records, std::unique_ptr<communicator> comm)
```

Unfortunately using a const reference here means that it could also accept a lifetime-extended temporary; consider using a pointer instead (suggested by @fknorr!).

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+
+	T* data() { return m_data.data(); }
+
+  private:
+	std::vector<T> m_data;
+	const size_t m_width;
+};
+
+/**
+ *  @brief This class gives a view into a const vector.
```
psalz, line 33:
```suggestion
 *  @brief This class gives a view into a const vector reference.
```

I was really confused what this was about at first!

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+ *
+ *	It is used to give us the currently unhashed task records while keeping track of the offset and width.
+ */
+template <typename T>
+struct sliding_window {
+  public:
+	sliding_window(const std::vector<T>& value) : m_value(value) {}
+
+	const T& operator[](size_t i) const {
+		assert(i >= 0 && i < m_width);
```
psalz, line 43:
```suggestion
		assert(i < m_width);
```
Can be simplified.

Also this assertion fails if `size()` is not called before, which seems weird?!

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+		return m_value[m_offset + i];
+	}
+
+	size_t size() {
+		m_width = m_value.size() - m_offset;
+		return m_width;
+	}
+
+	void slide(size_t i) {
+		assert(i == 0 || (i >= 0 && i <= m_width));
```
psalz, line 53:
```suggestion
		assert(i <= m_width);
```
Can be simplified.

include/divergence_block_chain.h
```diff
@@ -0,0 +1,166 @@
...
+	divergence_sceduler(divergence_sceduler&&) = default;
+
+	divergence_sceduler& operator=(const divergence_sceduler&) = delete;
+	divergence_sceduler& operator=(divergence_sceduler&&) = delete;
+
+	~divergence_sceduler() { stop(); }
+
+  private:
+	void start() {
+		stop();
```
psalz, line 143: ??

psalz (changes requested):
Thanks! I did a first pass and added some comments, mostly about comprehensibility / naming.

include/communicator.h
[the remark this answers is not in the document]
GagaLP: In my use case, this makes sense. But this was an addition, which I included after talking to @fknorr about the changes he had suggested. The idea here is to keep it as similar as possible to @fknorr communicator, which he is going to add for the IDAG, such that we can combine it.

include/divergence_block_chain.h
GagaLP: Again an oversight of my first implementation where it was possible to start the divergence checker two times, which had some different problems.

include/recorders.h
```diff
@@ -53,16 +53,21 @@ struct task_record {
...
+	using task_records = std::vector<task_record>;
+	using task_callback = std::function<void(const task_record&)>;
 
 	task_recorder(const buffer_manager* buff_mngr = nullptr) : m_buff_mngr(buff_mngr) {}
 
 	void record_task(const task& tsk);
 
-	const task_record& get_tasks() const { return m_recorded_tasks; }
+	void add_callback(task_callback callback);
+	void invoke_callbacks(const task_record& tsk) const;
```
psalz, line 64: Should not be public!

include/recorders.h
```diff
@@ -53,16 +53,21 @@ struct task_record {
 
 class task_recorder {
   public:
-	using task_record = std::vector<task_record>;
+	using task_records = std::vector<task_record>;
```
psalz, line 56: Why the renaming? A "record" could be considered a single instance or a collection of instances (which I assume was the intent, cc @PeterTh). In any case, it's inconsistent with ` command_recorder::command_record` now!

include/runtime.h
```diff
@@ -101,6 +102,8 @@ namespace detail {
 		size_t m_num_nodes;
 		node_id m_local_nid;
 
+		std::unique_ptr<divergence_checker_detail::divergence_checker> m_divergence_check;
```
psalz, line 105:
```suggestion
		std::unique_ptr<divergence_checker> m_divergence_check;
```
The `divergence_checker` is the public interface, so it should not be in `divergence_checker_detail`!

include/runtime.h
```diff
@@ -7,6 +7,7 @@
 #include "command.h"
 #include "config.h"
 #include "device_queue.h"
+#include "divergence_block_chain.h"
```
psalz, line 10:
```suggestion
#include "divergence_checker.h"
```

include/mpi_communicator.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+#include <memory>
```
psalz, line 2:
```suggestion

#include<memory>
```

include/divergence_block_chain.h
```diff
@@ -0,0 +1,137 @@
...
+		while(!is_finished || m_is_running) {
+			is_finished = m_block_chain.check_for_divergence();
+
+			std::this_thread::sleep_for(std::chrono::milliseconds(100));
+		}
+	}
+
+	std::thread m_thread;
+	bool m_is_running = false;
+	divergence_block_chain m_block_chain;
```
psalz, line 135: We usually have data members above functions!

include/mpi_communicator.h
```diff
@@ -0,0 +1,38 @@
...
+		return static_cast<size_t>(size);
+	}
+
+	node_id local_nid_impl() override {
+		int rank = -1;
+		MPI_Comm_rank(m_comm, &rank);
+		return static_cast<node_id>(rank);
+	}
+
+	MPI_Comm m_comm;
```
psalz, line 36: Move above member functions!

CMakeLists.txt
```diff
@@ -186,6 +186,7 @@ set(SOURCES
   src/command_graph.cc
   src/config.cc
   src/device_queue.cc
+  src/divergence_block_chain.cc
```
psalz, line 189:
```suggestion
  src/divergence_checker.cc
```

test/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ set(TEST_TARGETS
   test_utils_tests
   utils_tests
   device_selection_tests
+  divergence_check_tests
```
psalz, line 51:
```suggestion
  divergence_checker_tests
```

(Same for test utils)

src/runtime.cc
```diff
@@ -224,17 +232,27 @@ namespace detail {
...
+					CELERITY_TRACE("Command graph:\n\n{}\n", cmd_graph);
+				}
 			}
-			// must be called on all nodes
-			auto cmd_graph = gather_command_graph();
-			if(m_local_nid == 0) {
-				std::this_thread::sleep_for(std::chrono::milliseconds(500)); // Avoid racing on stdout with other nodes (funneled through mpirun)
-				CELERITY_TRACE("Command graph:\n\n{}\n", cmd_graph);
+
+			if(m_divergence_check != nullptr) {
```
psalz, line 250: Since `m_cfg->is_recording()` this has to be true always, right? I'd suggest moving this check to the top level instead.

src/runtime.cc
```diff
@@ -224,17 +232,27 @@ namespace detail {
...
+				// Sychronize all nodes before reseting shuch that we don't get into a deadlock
```
psalz, line 251:
```suggestion
				// Sychronize all nodes before resetting such that we don't get into a deadlock
```
Could you briefly explain why we could deadlock here?

src/runtime.cc
```diff
@@ -224,17 +232,27 @@ namespace detail {
...
+				MPI_Barrier(MPI_COMM_WORLD);
+				m_divergence_check.reset();
+			} else {
+				CELERITY_WARN("Divergence block chain not initialized");
```
psalz, line 255: Since this is outside the users control (basically only if a cosmic ray hits), I don't think we need to warn them about it.

include/ranges.h
```diff
@@ -229,6 +230,17 @@ struct ones_t {
 
 }; // namespace celerity::detail
 
+template <typename Interface, int Dims>
+struct std::hash<celerity::detail::coordinate<Interface, Dims>> {
+	std::size_t operator()(const celerity::detail::coordinate<Interface, Dims>& r) const noexcept {
+		std::size_t seed = 0;
+		for(int i = 0; i < Dims; ++i) {
+			celerity::detail::utils::hash_combine(seed, std::hash<int>{}(r[i]));
```
psalz, line 238: Shouldn't this be a `std::hash<size_t>`?

include/grid.h
```diff
@@ -257,6 +258,27 @@ class region {
...
+		celerity::detail::utils::hash_combine(seed, std::hash<celerity::id<Dims>>{}(r.get_min()), std::hash<celerity::id<Dims>>{}(r.get_max()));
+		return seed;
+	};
+};
+
+template <int Dims>
+struct std::hash<celerity::detail::region<Dims>> {
+	std::size_t operator()(const celerity::detail::region<Dims> r) {
+		std::size_t seed = 0;
+		for(auto box : r.get_boxes()) {
```
psalz, line 274:
```suggestion
		for(auto& box : r.get_boxes()) {
```

include/communicator.h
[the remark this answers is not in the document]
psalz: Fine with me, if the implementations actually end up having enough overlap to be merged (cc @fknorr)!

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+	divergence_map check_map;
+	for(size_t i = 0; i < m_num_nodes; ++i) {
+		check_map[task_hashes(i, task_num)].push_back(i);
+	}
+	return check_map;
+}
+
+void divergence_block_chain::check_for_deadlock() const {
+	auto diff = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - m_last_cleared);
+	static auto last = std::chrono::seconds(0);
```
psalz, line 51: Does this really have to be a static..?

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+}
+
+
+divergence_map divergence_block_chain::create_check_map(const per_node_task_hashes& task_hashes, const int task_num) const {
+	divergence_map check_map;
+	for(size_t i = 0; i < m_num_nodes; ++i) {
+		check_map[task_hashes(i, task_num)].push_back(i);
+	}
+	return check_map;
+}
```
psalz, line 47:
```suggestion
divergence_map divergence_block_chain::create_divergence_map(const per_node_task_hashes& task_hashes, const int task_num) const {
	divergence_map check_map;
	for(node_id nid = 0; nid < m_num_nodes; ++nid) {
		check_map[task_hashes(nid, task_num)].push_back(nid);
	}
	return check_map;
}
```

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+		}
+		return false;
+	}
+
+	const per_node_task_hashes task_graphs = collect_hashes(min_hash_count);
+
+	for(int j = 0; j < min_hash_count; ++j) {
+		const divergence_map check_map = create_check_map(task_graphs, j);
+
+		if(check_map.size() > 1) { divergence_out(check_map, j); }
```
psalz, line 151:
```suggestion
		// If there is more than one hash for this task, we have a divergence!
		if(check_map.size() > 1) { divergence_out(check_map, j); }
```

test/divergence_check_test_utils.h
```diff
@@ -0,0 +1,140 @@
+#pragma once
+
+#include "divergence_block_chain.h"
+
+using namespace celerity;
+using namespace celerity::detail;
+using namespace celerity::detail::divergence_checker_detail;
+
+namespace celerity::detail::divergence_checker_detail {
```
psalz, line 9:
```suggestion
```
I'd put this on the top level namespace!

test/divergence_check_tests.cc
```diff
@@ -0,0 +1,205 @@
...
+#include "test_utils.h"
+
+
+using namespace celerity;
+using namespace celerity::detail;
+using namespace celerity::test_utils;
+using celerity::access::fixed;
+
+TEST_CASE("test diverged task execution on device tasks", "[divergence]") {
+	using namespace cl::sycl::access;
```
psalz, line 18:
```suggestion
```
I realize you probably got this from some other test case, but this is very outdated. We have our own `celerity::access_mode` now!

test/divergence_check_test_utils.h
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	static void post_check(divergence_block_chain& div_test, const int min_size) { div_test.clear(min_size); }
+
+	static void call_check_for_divergence_with_pre_post(std::vector<std::unique_ptr<divergence_block_chain>>& div_test) {
+		std::vector<size_t> sizes;
+		std::transform(div_test.begin(), div_test.end(), std::back_inserter(sizes), [](auto& div) { return div->m_task_records.size(); });
+		auto [min, max] = std::minmax_element(sizes.begin(), sizes.end());
+
+		std::vector<per_node_task_hashes> extended_lifetime_hashes;
```
psalz, line 24: I don't understand what this is for..?

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return check_map;
+}
+
+void divergence_block_chain::check_for_deadlock() const {
+	auto diff = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - m_last_cleared);
+	static auto last = std::chrono::seconds(0);
+
+	if(diff >= std::chrono::seconds(10) && diff - last >= std::chrono::seconds(5)) {
+		std::string warning = fmt::format("After {} seconds of waiting nodes", diff.count());
```
psalz, line 54:
```suggestion
		std::string warning = fmt::format("After {} seconds of waiting node(s)", diff.count());
```

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+
+		for(size_t i = 0; i < m_num_nodes; ++i) {
+			if(m_per_node_hash_counts[i] == 0) { warning += fmt::format(" {},", i); }
+		}
```
psalz, line 58:
```suggestion
		std::vector<node_id> stuck_nodes;
		for(node_id nid = 0; nid < m_num_nodes; ++nid) {
			if(m_per_node_hash_counts[nid] == 0) stuck_nodes.push_back(nid);
		}
		warning += fmt::format(" {} ", fmt::join(stuck_nodes, ","));
```

This avoids the awkward final comma!

test/divergence_check_tests.cc
```diff
@@ -0,0 +1,205 @@
...
+
+		test_utils::log_capture log_capture;
+
+		divergence_block_chain_testspy::call_check_for_divergence_with_pre_post(div_tests);
+
+		CHECK_THAT(log_capture.get_log(), !Catch::Matchers::ContainsSubstring("Divergence detected"));
+	}
+}
+
+TEST_CASE("test divergence warning for tasks that are stale longer than 10 seconds", "[divergence]") {
```
psalz, line 131:
```suggestion
TEST_CASE("test deadlock warning for tasks that are stale longer than 10 seconds", "[divergence]") {
```

psalz (changes requested):
Things are coming together! GitHub decided to post some of my comments early for some reason... Well, here's the rest :P

By the way, we have a section on diverging execution in ours docs / [website](https://celerity.github.io/docs/pitfalls#diverging-host-execution-on-different-nodes), it would be great if you could add a sentence there mentioning this check and how it can be enabled!

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+
+bool divergence_block_chain::check_for_divergence() {
+	add_new_hashes();
+
+	const auto [min_hash_count, max_hash_count] = collect_hash_counts();
+
+	if(min_hash_count == 0) {
+		if(max_hash_count != 0 && m_local_nid == 0) {
+			check_for_deadlock();
+		} else if(max_hash_count == 0) {
```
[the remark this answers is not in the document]
GagaLP: Yes, this early return only implies that all current tasks are checked. The thread isn't terminated solely on this flag. For the divergence check to be completely done it needs both this to be finished and the running flag to be false. Until one of those conditions isn't met the divergence checker won't be done.

src/divergence_block_chain.cc
```diff
@@ -0,0 +1,177 @@
...
+
+		warning += " did not move to the next task. The runtime might be stuck.";
+
+		CELERITY_WARN("{}", warning);
+		last = diff;
+	}
+}
+
+void divergence_block_chain::log_node_divergences(const divergence_map& check_map, const int task_num) {
+	std::string error = fmt::format("Divergence detected in task graph at index {}:\n\n", task_num);
```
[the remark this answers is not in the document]
GagaLP: As also discussed in-person I reformulated the whole error message. Now we get an error message like this one:

```
[2023-12-18 11:50:13.109] [0] [error] Divergence detected. Task Nr 4 diverges on nodes:

Following task-hash 0xaa7e7279ec138abe resulted on node  2 
Following task-hash 0x816fa348a1f52942 resulted on node  1 
Following task-hash 0x1fecf12dff1bff6c resulted on nodes 0 3 

[2023-12-18 11:50:13.109] [0] [error] Task record for hash 0x1fecf12dff1bff6c:

id: 4, debug_name: one_third_kernel, type: device-compute, cgid: 0
geometry:
         dimensions: 1, global_size: [10000,1,1], global_offset: [0,0,0], granularity: [1,1,1]
accesses: 
         bid: 0, buffer_name: B0 "mat_a", mode: 3, req: {[0,0,0] - [10000,1,1]}
dependencies: 
         node: 3, kind: anti-dep, origin: dataflow
         node: 0, kind: true-dep, origin: last-epoch
```

This includes now the `task_id` instead of `task_num`, which is much more informative.

test/divergence_check_test_utils.h
GagaLP: The `extended_lifetime_hashes` are needed, because of how the single node `divergence_test_communicator` works. The idea is following:
* each _simulated node_ calls allgather sequentially. The send/receive buffer is saved internally as simple pointers.
* When all _simulated nodes_ called the allgather the data is copied. 

To prevent those send/receive buffers for `per_node_task_hashes` from going out of scope we need to "extend" their lifetime. For this, I included the `extended_lifetime_hashes` vector.

include/divergence_checker.h
```diff
@@ -67,34 +63,38 @@ class divergence_block_chain {
...
 	std::vector<int> m_per_node_hash_counts;
 	std::mutex m_task_records_mutex;
 
 	std::chrono::time_point<std::chrono::steady_clock> m_last_cleared = std::chrono::steady_clock::now();
+	std::chrono::seconds m_time_of_last_warning = std::chrono::seconds(0);
 
 	std::unique_ptr<communicator> m_communicator;
 
-	void divergence_out(const divergence_map& check_map, const int task_num);
+	void reprot_divergence(const divergence_map& check_map, const int task_num);
```
psalz, line 76:
```suggestion
	void report_divergence(const divergence_map& check_map, const int task_num);
```

src/runtime.cc
```diff
@@ -177,12 +177,18 @@ namespace detail {
...
-			m_divergence_check =
-			    std::make_unique<divergence_checker_detail::divergence_checker>(*m_task_recorder, std::make_unique<mpi_communicator>(comm), m_test_mode);
-		}
+
+#if CELERITY_DIVERGENCE_CHECK
+		MPI_Comm comm = nullptr;
+		MPI_Comm_dup(MPI_COMM_WORLD, &comm);
+		m_divergence_check = std::make_unique<divergence_checker>(*m_task_recorder, std::make_unique<mpi_communicator>(comm), m_test_mode);
+#endif
+		// if (m_cfg->is_recording()) {
```
psalz, line 186: Comment

src/runtime.cc
```diff
@@ -247,15 +253,19 @@ namespace detail {
...
-				// Sychronize all nodes before reseting shuch that we don't get into a deadlock
-				MPI_Barrier(MPI_COMM_WORLD);
-				m_divergence_check.reset();
-			} else {
-				CELERITY_WARN("Divergence block chain not initialized");
-			}
+			// // Sychronize all nodes before reseting shuch that we don't get into a deadlock
+			// // With this barrier we can be shure that every node is fully finished and all mpi operations are done. (Sending ...)
+			// MPI_Barrier(MPI_COMM_WORLD);
+			// m_divergence_check.reset();
```
psalz, line 259: Comment

src/runtime.cc
```diff
@@ -224,20 +238,34 @@ namespace detail {
...
 			}
+
+			// // Sychronize all nodes before reseting shuch that we don't get into a deadlock
+			// // With this barrier we can be shure that every node is fully finished and all mpi operations are done. (Sending ...)
+			// MPI_Barrier(MPI_COMM_WORLD);
+			// m_divergence_check.reset();
 		}
 
+#if CELERITY_DIVERGENCE_CHECK
+		// Sychronize all nodes before reseting shuch that we don't get into a deadlock
```
PeterTh, line 263: A few typos "Sychronize" "reseting" "shuch"

PeterTh (approved):
LGTM, other than a few typos and commented out code.

CMakeLists.txt
```diff
@@ -289,6 +294,7 @@ target_compile_definitions(celerity_runtime PUBLIC
   CELERITY_FEATURE_UNNAMED_KERNELS=$<BOOL:${CELERITY_FEATURE_UNNAMED_KERNELS}>
   CELERITY_DETAIL_HAS_NAMED_THREADS=$<BOOL:${CELERITY_DETAIL_HAS_NAMED_THREADS}>
   CELERITY_ACCESSOR_BOUNDARY_CHECK=$<BOOL:${CELERITY_ACCESSOR_BOUNDARY_CHECK}>
+  CELERITY_DIVERGENCE_CHECK=$<BOOL:${CELERITY_DIVERGENCE_CHECK}>
```
psalz, line 297: Also needs to be added to `cmake/celerity-config.cmake.in`!

src/config.cc
```diff
@@ -201,7 +201,12 @@ namespace detail {
 			const auto has_dry_run_nodes = parsed_and_validated_envs.get(env_dry_run_nodes);
 			if(has_dry_run_nodes) { m_dry_run_nodes = *has_dry_run_nodes; }
 
+#if CELERITY_DIVERGENCE_CHECK
+			// divergence checker needs recording
+			m_recording = true;
+#else
 			m_recording = parsed_and_validated_envs.get_or(env_recording, false);
+#endif
```
psalz, line 209: Should we print a warning that recording is being force-enabled here? What about the user explicitly setting `CELERITY_RECORDING=0`? cc @PeterTh @fknorr

src/divergence_checker.cc
```diff
@@ -0,0 +1,178 @@
...
+	for(size_t i = m_hashes_added; i < m_task_records.size(); ++i) {
+		std::size_t seed = m_local_hashes.empty() ? m_last_hash : m_local_hashes.back();
+		celerity::detail::utils::hash_combine(seed, std::hash<task_record>{}(m_task_records[i]));
+		m_local_hashes.push_back(seed);
+	}
+	m_last_hash = m_local_hashes.empty() ? m_last_hash : m_local_hashes.back();
+	m_hashes_added = m_task_records.size();
+}
+
+void divergence_block_chain::clear(const int min_progress) {
```
fknorr, line 56: Naming: This does not clear the chain (similar to what `vector::clear` would do. Maybe something along the lines of `erase_front`, `prune_leading`, or similar?

fknorr (changes requested):
A few more comments!

Also since the clang-tidy check for this seems broken in CI: Please go over all new function definitions and make sure parameters that can be const are const.

src/config.cc
psalz: Agreed, maybe we should just get rid of it in a small follow-up PR.

include/divergence_checker.h
```diff
@@ -0,0 +1,137 @@
...
+	divergence_checker(const divergence_checker&&) = delete;
+
+	divergence_checker& operator=(const divergence_checker&) = delete;
+	divergence_checker& operator=(divergence_checker&&) = delete;
+
+	~divergence_checker() { stop(); }
+
+  private:
+	std::thread m_thread;
+	bool m_is_running = false;
```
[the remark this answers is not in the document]
psalz: Or just use an atomic.

psalz:
Okay so as discussed offline, we won't include this in 0.5.0 as it needs another revision. The main points:

- Deadlock detection as-is would produce too many false positive warnings; not sure yet how to proceed on this.
- Testing infrastructure invokes UB; needs multi-threading to properly mock blocking collective operations
- We should have a test case (distr_test / integration test?) that exercises the case that one node submits a task while the other does not (the divergence then occurs between that task and the shutdown epoch).

PR #221, opened 2023-10-23
Improve CSV reporting of benchmarks, add stencil system benchmark

PeterTh:
This PR is in preparation of a follow-up which will use tags in the evaluation for categorization.
This needs to be done in 2 steps due to the involvement of the repository state in the CI benchmark worklow.

**Preparatory changes**:
* Outputs the tag list in to the CSV reporter.
* Changes the script to be resilient to csv layout/order changes.
* Unifies tagging in benchmarks.
* Adds another stencil-like system benchmark.

The last point might seem a bit unrelated. I added it because otherwise, our "full system" benchmark category would include *only* the many-task benchmark, which is quite far from Celerity's common use case currently.  
We always wanted to extend this set anyway, so now I made a first step with a very simple iterative stencil benchmark that should better represent some common usage scenarios in terms of runtime system performance impact.

The goal is to be able to finish https://github.com/celerity/meta/issues/47 in the follow-up.


PeterTh:
This goes to show that the threshold of "significance" also needs to be tweaked in the follow-up commit which reworks this.

test/system_benchmarks.cc
```diff
@@ -67,3 +67,73 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+		}
+		queue.slow_full_sync();
+	};
+
+	// check result
+	celerity::buffer<bool, 0> success_buffer = true;
+	queue.submit([&](celerity::handler& cgh) {
+		celerity::accessor r{buffer_a, cgh, celerity::access::all{}, celerity::read_only_host_task};
+		celerity::accessor succ{success_buffer, cgh, celerity::access::all{}, celerity::write_only_host_task};
+		cgh.host_task(celerity::on_master_node_tag{}, [=]() {
```
psalz, line 127:
```suggestion
		cgh.host_task(celerity::on_master_node, [=]() {
```

test/system_benchmarks.cc
```diff
@@ -67,3 +67,73 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+	BENCHMARK("iterations") {
+		for(size_t r = 0; r < num_iterations; ++r) {
+			queue.submit([&](celerity::handler& cgh) {
+				celerity::accessor read{buffer_a, cgh, celerity::access::neighborhood(1, 1), celerity::read_only};
+				celerity::accessor write{buffer_b, cgh, celerity::access::one_to_one(), celerity::write_only, celerity::no_init};
+				cgh.parallel_for(size, [=](celerity::item<2> item) {
+					float sum = 0.f;
+					float included_items = 0.f;
+					for(int i = -1; i <= 1; ++i) {
+						int x = static_cast<int>(item.get_id(0)) + i;
```
psalz, line 105:
```suggestion
						const int x = static_cast<int>(item.get_id(0)) + i;
```

test/system_benchmarks.cc
```diff
@@ -67,3 +67,73 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+						if(x < 0 || x >= side_length) continue;
+						for(int j = -1; j <= 1; ++j) {
+							int y = static_cast<int>(item.get_id(1)) + j;
```
psalz, line 108:
```suggestion
							const int y = static_cast<int>(item.get_id(1)) + j;
```

psalz (approved):
LGTM!

PR #222, opened 2023-10-24
CI: use benchmark groups when reporting results

PeterTh:
* Also only add benchmarks to the charts if there is a notable difference
* Tweak thresholds to be a bit less sensitive
* Improve script documentation

Closes https://github.com/celerity/meta/issues/47.


PeterTh:
The data above is fake (or rather, from a run on a completely different system from our canonical benchmark system), I just included it for testing. will now remove that commit.

psalz (approved):

fknorr (approved):
LGTM!

PR #223, opened 2023-10-25
Fix command graph generation bugs around reductions

fknorr:
Implementing IDAG reductions uncovered two bugs around reductions in distributed command graph generation:

- For an all-to-all reduction, we emit push commands for the partial results from our node followed by a reduction command. The reduction command logically overwrites the buffer contents, so it must anti-depend on these pushes. This bug does not appear to break reductions in the current runtime, most likely because the final reduction result is only committed to device memory once it's being read in the next consumer task.
- We elide reduction commands if there only is a single producer chunk. If the result is subsequently read by multiple nodes, we generate push commands on the producer, but failed to generate the corresponding await-pushes on the consumer node.

I've added unit tests for both cases.


src/distributed_graph_generator.cc
```diff
@@ -185,7 +185,7 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
 	const node_id reduction_initializer_nid = 0;
 
 	const box<3> empty_box({0, 0, 0}, {0, 0, 0});
```
psalz, line 187:
```suggestion
	const box<3> empty_reduction_box({0, 0, 0}, {0, 0, 0});
```
This is used in one location, when generating notification-only reduction pushes. Should it also be renamed?

src/distributed_graph_generator.cc
```diff
@@ -473,11 +473,22 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
 				}
 				m_cdag.add_dependency(writer_cmd, push_cmd, dependency_kind::anti_dep, dependency_origin::dataflow);
 			}
+
+			// reduction commands will overwrite their buffer, so they must anti-depend on their partial-result push-commands
```
psalz, line 477: Oh wow, so we found another "intra-task race condition"!

psalz (approved):
Good stuff!

test/graph_gen_reduction_tests.cc
```diff
@@ -212,3 +212,34 @@ TEST_CASE("reductions that do not include the current value generate anti-depend
...
 	CHECK(dctx.query(tid_write).have_successors(dctx.query(tid_reduce), dependency_kind::anti_dep));
 }
+
+TEST_CASE("reduction commands anti-depend on their partial-result push commands", "[distributed_graph_generator][command-graph][reductions]") {
+	const size_t num_nodes = 2;
+	dist_cdag_test_context dctx(2);
+	auto buf = dctx.create_buffer(range<1>(1));
+
+	const auto tid_producer = dctx.device_compute(range<1>(num_nodes)).reduce(buf, false /* include_current_buffer_value */).submit();
+	const auto tid_consumer = dctx.device_compute(range<1>(num_nodes)).read(buf, acc::all{}).submit();
```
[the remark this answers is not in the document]
PeterTh: We have a lot of these in the tests already, but we probably still shouldn't add more. I'd either remove the name or add `[[maybe_unused]]`.

PeterTh (approved):
LGTM

fknorr:
I re-ran the benchmarks because there seemed to be significant jitter in the system benchmarks, but it appears that "benchmark independent task pattern with 100 tasks" is indeed slowing down, even though the change should not affect code without reductions.

fknorr:
@PeterTh discovered that results of our multi-threaded benchmarks, especially system benchmarks, are not as stable and reliable as we thought, and our benchmarking setup needs some work.

Aside from extremely obscure reason in instruction cache, OS scheduling or similar, I'm going to trust the command-graph benchmarks which measure this change in isolation and do not show a change in performance.

PR #225, opened 2023-11-02
Stabilize experimental::fence

fknorr:
`celerity::experimental::fence` and `celerity::experimental::buffer_snapshot` are now `celerity::distr_queue::fence` (a member function) and `celerity::buffer_snapshot`. The experimental declarations are aliases of the new names, and a _deprecation test_ ensures they still work.

I have decided to keep the `fence` header since the definitions of `buffer_snapshot` and `*_fence_promise` don't fit into what I believe should be `distr_queue.h`.

Closes celerity/meta#69.


GagaLP (approved):
Nice work. Besides the seemingly odd experimental::host_object function this looks like a good change.

psalz (approved):

PR #226, opened 2023-11-08
CI: Core pinning for benchmark runs to reduce variance

PeterTh:
We noticed that there are extremely large (~100%) differences in the runtime of specific benchmarks on the current CI target system, which are related to thread scheduling. In order to mitigate this issue, the CI benchmark script now pins the benchmarks to a small subset of cores.

As the following data of 2 benchmark runs on master and 2 on this branch shows, this both reduces overall runtimes and, more importantly, reduces the std deviation overall and in particular for the system benchmarks.

std_devs
times

The initial idea for this was to use `mpirun` core pinning, but this fails due to HWLOC not working within the CI docker image. This now uses `taskset`, since that has equivalent results.


fknorr (commented):
The stats look pretty good!

I noticed though that you're only assigning 4 "cores", which is below the recommended minimum of 5, which Celerity warns about [in the benchmarks CI job](https://github.com/celerity/celerity-runtime/actions/runs/6796237093/job/18475892925):

```
[warning] Celerity has detected that only 4 logical cores are available to this process. It is recommended to assign at least 5 logical cores. Performance may be negatively impacted.
```

PeterTh:
Yes, this is a bit regrettable, but on purpose. Using more increases the variance by a noticeable amount.

fknorr (approved):
Sure, this is purely an internal thing anyway.

Any reason why there's no benchmark report / CI in general on this PR?

PeterTh:
I didn't want it to run general CI since that would occupy `gpuc2` (and there are 0 relevant general CI changes).

The benchmark report is also disabled as a consequence of that, but it would be meaningless anyway. I guess it would look good though, Celerity getting that much *faster* :P

fknorr:
Ah, I didnt see the `[skip ci]` and was wondering if there was another CI hiccup :+1:

ci/run-benchmarks.sh
```diff
@@ -11,6 +11,7 @@ if [[ $# -ne 0 ]]; then
     exit 1
 fi
 
-bash /root/capture-backtrace.sh test/benchmarks \
+# we taskset to pin the process to a specific set of cores and their HTs to reduce test result variance
+bash /root/capture-backtrace.sh taskset -c 4-5,16-17 test/benchmarks \
```
psalz, line 15: Why `4-5,16-17` and not `0-1,12-13`? Is it arbitrary? In either case, for future reference consider adding a note that this is specific to our gpuc2 server, which uses a Threadripper such and such.

psalz (approved):

ci/run-benchmarks.sh
PeterTh: It's not completely arbitrary, I wanted to use a node range that's not in the first CCX, but contained in one CCX. I'll add a note.

GagaLP (approved):

PR #227, opened 2023-11-16
Add experimental task hints API, support for 2D splits

psalz:
This adds a new `experimental::hint` API that can be used to provide the runtime with additional information on how to process a task. The first two hints we have are `experimental::hints::split_1d` and `experimental::hints::split_2d`. As a 1D split is currently the default, the former is a no-op. For the latter this introduces a new `split_2d` function that attempts to produce a number of evenly sized chunks while satisfying the provided granularity constraints.

The algorithm works by trying to find two factors for `num_chunks`, the number of requested chunks, that are as close to `sqrt(num_chunks)` as possible while also producing the requested number of chunks, degrading to a 1D split in the worst case. Due to granularity constraints it is not always possible to produce the requested number of chunks however, in that case the algorithm will attempt to get as close as possible. Finding the factors requires at most `O(sqrt(num_chunks))` time.

Given the two factors, the algorithm uses two heuristics to decide to which side of the domain they should be assigned:
- If the domain is square-ish, the algorithm prefers to have more splits along the slower dimension.
- If the domain is non-square, the algorithm prefers the split that produces overall shorder edges (smaller sum of circumferences).

Here's a visual representation of some of the 2D splits from the unit tests. Each split is labelled in the form `<size0>x<size1> ~ <granularity0>x<granularity1> | <num chunks>`. Red lines represent split constraints (i.e., individual red boxes can not be split further):

image

### DIFF
--- a/include/handler.h
+++ b/include/handler.h
@@ -13,6 +13,7 @@
 #include "cgf_diagnostics.h"
 #include "closure_hydrator.h"
 #include "device_queue.h"
+#include "hint.h"
 #include "host_queue.h"
 #include "item.h"
 #include "range_mapper.h"
@@ -391,6 +392,8 @@ class handler {
 	friend void detail::add_reduction(handler& cgh, const detail::reduction_info& rinfo);
 	template <int Dims>
 	friend void experimental::constrain_split(handler& cgh, const range<Dims>& constraint);
+	template <typename Hint>
+	friend void experimental::hint(handler& cgh, Hint&& h);
 	friend void detail::extend_lifetime(handler& cgh, std::shared_ptr<detail::lifetime_extending_state> state);
 
 	friend void detail::set_task_name(handler& cgh, const std::string& debug_name);
@@ -406,6 +409,7 @@ class handler {
 	std::vector<std::shared_ptr<detail::lifetime_extending_state>> m_attached_state;
 	std::optional<std::string> m_usr_def_task_name;
 	range<3> m_split_constraint = detail::ones;
+	std::vector<std::unique_ptr<detail::hint_base>> m_hints;
 
 	handler(detail::task_id tid, size_t num_collective_nodes) : m_tid(tid), m_num_collective_nodes(num_collective_nodes) {}
 
@@ -462,6 +466,19 @@ class handler {
 	void experimental_constrain_split(const range<Dims>& constraint) {
 		assert(m_task == nullptr);
 		m_split_constraint = detail::range_cast<3>(constraint);
+	}
+
+	template <typename Hint>
+	void experimental_hint(Hint&& hint) {
+		static_assert(std::is_base_of_v<detail::hint_base, std::decay_t<Hint>>, "Hint must extend hint_base");
+		static_assert(std::is_move_constructible_v<Hint>, "Hint must be move-constructible");
+		for(auto& h : m_hints) {
+			// We currently don't allow more than one hint of the same type for simplicity; this could be loosened in the future.
+			auto& hr = *h; // Need to do this here to avoid -Wpotentially-evaluated-expression
+			if(typeid(hr) == typeid(hint)) { throw std::runtime_error("Providing more than one hint of the same type is not allowed"); }
+			h->validate(hint);
+		}
+		m_hints.emplace_back(std::make_unique<std::decay_t<Hint>>(std::forward<Hint>(hint)));
 	}
 
 	template <int Dims>
@@ -604,6 +621,9 @@ class handler {
 		for(auto state : m_attached_state) {
 			m_task->extend_lifetime(std::move(state));
 		}
+		for(auto& h : m_hints) {
+			m_task->add_hint(std::move(h));
+		}
 		return std::move(m_task);
 	}
 };
@@ -682,5 +702,10 @@ namespace celerity::experimental {
 template <int Dims>
 void constrain_split(handler& cgh, const range<Dims>& constraint) {
 	cgh.experimental_constrain_split(constraint);
+}
+
+template <typename Hint>
+void hint(handler& cgh, Hint&& hint) {
+	cgh.experimental_hint(std::forward<Hint>(hint));
 }
 } // namespace celerity::experimental
--- a/include/hint.h
+++ b/include/hint.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include <memory>
+#include <stdexcept>
+#include <vector>
+
+namespace celerity {
+class handler;
+}
+
+// Definition is in handler.h to avoid circular dependency
+namespace celerity::experimental {
+template <typename Hint>
+void hint(handler& cgh, Hint&& hint);
+}
+
+namespace celerity::detail {
+
+class hint_base {
+  public:
+	hint_base() = default;
+	hint_base(const hint_base&) = default;
+	hint_base& operator=(const hint_base&) = default;
+	hint_base(hint_base&&) = default;
+	hint_base& operator=(hint_base&&) = default;
+	virtual ~hint_base() = default;
+
+  private:
+	friend class celerity::handler;
+	virtual void validate(const hint_base& other) const {}
+};
+
+} // namespace celerity::detail
+
+namespace celerity::experimental::hints {}; // namespace celerity::experimental::hints
--- a/include/hint.h
+++ b/include/hint.h
@@ -32,4 +32,31 @@ class hint_base {
 
 } // namespace celerity::detail
 
-namespace celerity::experimental::hints {}; // namespace celerity::experimental::hints
+namespace celerity::experimental::hints {
+
+/**
+ * Suggests that the task should be split into 1D chunks.
+ * This is currently the default behavior.
+ */
+class split_1d : public detail::hint_base {
+  private:
+	void validate(const hint_base& other) const override;
+};
+
+/**
+ * Suggests that the task should be split into 2D chunks.
+ */
+class split_2d : public detail::hint_base {
+  private:
+	void validate(const hint_base& other) const override;
+};
+
+inline void split_1d::validate(const hint_base& other) const {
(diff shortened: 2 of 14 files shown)


src/split.cc
```diff
@@ -0,0 +1,155 @@
...
+		result[i].offset[0] += i * large_chunk_size[0];
+	}
+	for(auto i = num_large_chunks[0]; i < actual_num_chunks[0]; ++i) {
+		result[i].range[0] = small_chunk_size[0];
+		result[i].offset[0] += num_large_chunks[0] * large_chunk_size[0] + (i - num_large_chunks[0]) * small_chunk_size[0];
+	}
+
+#ifndef NDEBUG
+	sanity_check_split(full_chunk, result);
+#endif
```
PeterTh, line 67: I appreciate this. (And it's another good reason that we should emphasize the utility of debug -- or optimized-with-debug-features -- builds for development in the documentation in the future)

test/split_tests.cc
```diff
@@ -0,0 +1,306 @@
...
+TEST_CASE("split_2d respects granularity constraints") {
+	SECTION("simple constrained split") {
+		const auto full_chunk = make_full_chunk<2>({128, 128});
+		const auto chunks = split_2d(full_chunk, {8, 8, 1}, 8);
+		check_2d_split(full_chunk, chunks,
+		    {
+		        {32, {64, 64}}, //
+		        {32, {64, 64}}, //
+		        {32, {64, 64}}, //
+		        {32, {64, 64}}  //
```
PeterTh, line 135: In the other tests, I think the `,` at the last line forces clang-format to format it as intended, but here you use `//`. If this is accurate, then for consistency I'd change this and the previous instance to also use the `,` style.

include/handler.h
```diff
@@ -476,6 +480,20 @@ class handler {
 		m_split_constraint = detail::range_cast<3>(constraint);
 	}
 
+	template <typename Hint>
+	void experimental_hint(Hint&& h) {
```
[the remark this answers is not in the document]
psalz: Yes good point, I like that better!

include/hint.h
```diff
@@ -0,0 +1,62 @@
...
+class split_2d : public detail::hint_base {
+  private:
+	void validate(const hint_base& other) const override;
+};
+
+inline void split_1d::validate(const hint_base& other) const {
+	if(dynamic_cast<const split_2d*>(&other) != nullptr) { throw std::runtime_error("Cannot combine split_1d and split_2d hints"); }
+}
+
+inline void split_2d::validate(const hint_base& other) const {
```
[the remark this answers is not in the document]
psalz: Hmm true. The whole validation thing was a last minute addition after we decided to also add `split_1d`, before that there was no way of producing a conflict. The reason why I'm doing it here and not later is so I can throw directly within the call to `experimental::hint`. But since we wanted to pivot to aborting on these kinds of errors anyway, I guess it could also be deferred.

That being said, we may want to wait for another use case for more complex validation logic, `split_2d` also works for 1D and 0D kernels (the latter is a no-op).

src/split.cc
```diff
@@ -0,0 +1,155 @@
...
+		const auto count0 = split0[0] * split0[1];
+		const auto count1 = split1[0] * split1[1];
+
+		if(count0 == count1) {
+			// If we're tied for the number of chunks we can create, try some heuristics to decide.
+
+			// If domain is square(-ish), prefer splitting along slower dimension.
+			// (These bounds have been chosen arbitrarily!)
+			const double squareishness = std::sqrt(full_chunk.range.size()) / static_cast<double>(full_chunk.range[0]);
+			if(squareishness > 0.95 && squareishness < 1.05) { return (f0 >= f1) ? split0 : split1; }
```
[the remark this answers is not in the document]
psalz: `f0` and `f1` are the two split factors, i.e., how many cuts there are on each side of the domain. E.g. for 16 nodes `f0 = f1 = 4`, and the choice doesn't matter. But for 8 nodes `f0 = 2` and `f1 = 4`, so there's a choice to make. In this case, if the domain is about square, it will assign `f1` (more cuts) to dimension 0 (the slower dimension), and `f0` (fewer cuts) to dimension 1.

test/split_tests.cc
psalz: Ah, that's why, I wrote those tests a while ago and was confused why seemingly only some needed the comments. Will fix.

include/handler.h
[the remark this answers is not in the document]
fknorr: But that way you would always have to provide the type template argument, right?

include/hint.h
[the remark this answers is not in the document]
fknorr: I would allow any hint in weird situations as long as it has reasonable fallback behavior. A `range<2>` can also be 1-dimensional at runtime.

include/hint.h
```diff
@@ -0,0 +1,62 @@
...
+
+// Definition is in handler.h to avoid circular dependency
+namespace celerity::experimental {
+template <typename Hint>
+void hint(handler& cgh, Hint&& h);
+}
+
+namespace celerity::detail {
+
+class hint_base {
```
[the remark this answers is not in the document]
fknorr: I think move-constructible is enough because it's being forwarded?

src/split.cc
```diff
@@ -0,0 +1,155 @@
...
+			// For non-square domains, prefer split that produces shorter edges (compare sum of circumferences)
+			const auto circ0 = full_chunk.range[0] / split0[0] + full_chunk.range[1] / split0[1];
+			const auto circ1 = full_chunk.range[0] / split1[0] + full_chunk.range[1] / split1[1];
+			return circ0 < circ1 ? split0 : split1;
+
+			// TODO: Yet another heuristic we may want to consider is how even chunk sizes are,
+			// i.e., how balanced the workload is.
+		}
+		if(count0 > count1) { return split0; }
+		return split1;
```
fknorr, line 112: Consider making these two the first branches in an if-else statement (they're the easiest to decide).

src/split.cc
```diff
@@ -0,0 +1,155 @@
...
+std::vector<chunk<3>> split_2d(const chunk<3>& full_chunk, const range<3>& granularity, const size_t num_chunks) {
+#ifndef NDEBUG
+	assert(num_chunks > 0);
+	for(int d = 0; d < 3; ++d) {
+		assert(granularity[d] > 0);
+		assert(full_chunk.range[d] % granularity[d] == 0);
+	}
+#endif
+
+	const auto assign_factors = [&full_chunk, &granularity, &num_chunks](const size_t factor) {
```
fknorr, line 82: This lambda could use some documentation: what's the return value, whats the `factor` in a 2d setting?. Also making it a free-standing function would make it easier to recognize that it's not sharing (mutable) state with any local variables of the outer function

src/split.cc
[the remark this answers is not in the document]
fknorr: I understand what's happening, but some of that explanation could be code comments. Also consider making some of the variable names more _wortreich_ to convey that `0` and `1` refer to the split direction each time.

src/split.cc
```diff
@@ -0,0 +1,155 @@
...
+		}
+		if(count0 > count1) { return split0; }
+		return split1;
+	};
+
+	// Factorize num_chunks
+	// Try to find factors as close to the square root as possible, that also produce
+	// (or come close to) the requested number of chunks (under the given constraints).
+	size_t f = std::floor(std::sqrt(num_chunks));
+	std::array<size_t, 2> best_f_counts = {0, 0};
```
fknorr, line 119: What does `best_f_counts` stand for? Number of cuts per direction if `f` were the split factor?

fknorr (changes requested):
Nice! That is an impressive amount of testing. Some code, especially `split_2d` internals, is rather hard to understand though.

fknorr:
I noticed that `task_manager::submit_command_group` has a `hints` parameter, I guess that's obsolete now.

include/handler.h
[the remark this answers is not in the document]
psalz: As discussed in person, will leave as is!

include/hint.h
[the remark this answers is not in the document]
psalz: I've added a static assertion for `is_move_constructible`.

src/split.cc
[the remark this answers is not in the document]
psalz: It should be more understandable now hopefully!

src/split.cc
psalz: I've moved it to a free function and added some documentation!

src/split.cc
psalz: It means "the closest we've gotten to the requested number of chunks with any of the previous factors `f`". Renamed to `best_chunk_counts` and documented.

src/split.cc
```diff
@@ -0,0 +1,166 @@
+#include "split.h"
+
+#include <array>
```
fknorr, line 3: I feel like most (all?) of the uses of `std::array<size_t, Dims>` could instead be `range<Dims>`.

include/handler.h
```diff
@@ -476,6 +480,19 @@ class handler {
 		m_split_constraint = detail::range_cast<3>(constraint);
 	}
 
+	template <typename Hint>
+	void experimental_hint(Hint&& hint) {
+		static_assert(std::is_base_of_v<detail::hint_base, std::decay_t<Hint>>, "Incompatible hint");
```
fknorr, line 485:
```suggestion
		static_assert(std::is_base_of_v<detail::hint_base, Hint>, "Hint must extend hint_base");
```
`task::get_hint()` has a better error message for this case.

fknorr (approved):
Looks great!

src/split.cc
psalz: Agreed for small/large chunk sizes; for chunk counts I feel like the `array` is a better semantic fit!

PeterTh (approved):

PR #229, opened 2023-12-04
Don't let partial replication unnecessarily divide push commands

psalz:
This was surfaced by 2D splits, but can also be constructed (somewhat awkwardly) for 1D: The extra region map lookup into `buffer_state.replicated_regions` could sometimes cause a box with the same last writer command to be unnecessarily split into several push commands. The fix is to collect all non-replicated boxes and do a final merge before generating the push commands.


PeterTh (approved):
LGTM!
I'm taking your word for the fact that the new unit test replicates the previously problematic behaviour.

src/distributed_graph_generator.cc
```diff
@@ -315,27 +315,33 @@ void distributed_graph_generator::generate_distributed_commands(const task& tsk)
...
+								non_replicated_boxes.push_back(replicated_box);
+							}
 
-								// Generate separate push command for each last writer command for now,
-								// possibly even multiple for partially already-replicated data.
-								// TODO: Can and/or should we consolidate?
-								auto* const push_cmd = create_command<push_command>(bid, 0, nid, trid, replicated_box.get_subrange());
+							// Merge all connected boxes to determine final set of pushes
+							const auto push_boxes = region<3>(std::move(non_replicated_boxes));
+							for(auto& push_box : push_boxes.get_boxes()) {
```
fknorr, line 333:
```suggestion
							const auto push_boxes = region<3>(std::move(non_replicated_boxes)).into_boxes();
							for(auto& push_box : push_boxes.into_boxes()) {
```
I just remembered that `region::into_boxes() &&` exists, that way the variable name also makes more sense.

fknorr (approved):
LGTM! This now also generates the correct number of sends in the IDAG.

src/distributed_graph_generator.cc
psalz: I need the region again below the loop for updating `buffer_state.replicated_regions`. Renamed to `push_region`!

PR #230, opened 2023-12-05
Re-define transfer_id as struct{task_id, buffer_id, reduction_id}

fknorr:
This is another back-port from IDAG development.

The `transfer_id` is used to match incoming MPI transfers to await-push commands. Previously this was an integer encoding of "task + chunk id", which together with a `buffer_id`, and (optionally) a `reduction_id` uniquely identifies an await-push on the target node.

It turns out that instead of the "magic" chunk id, the task id is sufficient. This PR bundles it with the buffer id and reduction id such that `struct transfer_id` now uniquely identifies the receiving await-push, improving readability of logs / dot graphs and avoiding repetitions of `(transfer_id, buffer_id, reduction_id)` in the code.


fknorr:
Bikeshedding, from an in-person discussion with @psalz: `transfer_id` might not be a good name because it does not uniquely identify a transfer: There can be multiple transfers per `transfer_id` on each node (for different subregions) and the same `transfer_id + region` can exist multiple times in the cluster if the same data is sent to more than one node.

How about `buffer_version` or `buffer_version_id` (see source comments)?

include/types.h
```diff
@@ -76,6 +76,34 @@ struct reduction_info {
 
 constexpr node_id master_node_id = 0;
 
+inline constexpr reduction_id no_reduction_id = 0;
```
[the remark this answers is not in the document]
psalz: Hmm, what's going on here...? 🤔

include/types.h
```diff
@@ -4,6 +4,7 @@
 #include <functional>
 #include <utility>
 
+
```
PeterTh, line 7: Do we usually have 2 newlines here? (nitpick)

include/types.h
```diff
@@ -76,6 +76,34 @@ struct reduction_info {
...
+/// Uniquely identifies one version of a buffer's (distributed) data at task granularity. The structure is used to tie together the sending and receiving ends
+/// of peer-to-peer data transfers.
+struct transfer_id {
+	/// The first task (by order of task id) to require this version of the buffer.
+	task_id consumer_tid = -1;
+
+	/// The buffer's id.
+	buffer_id bid = -1;
+
+	/// The reduction the data belongs to. If so, the transferred data consists of partial results that will be consumed by a subsequent reduction command to
```
PeterTh, line 90: I'd say "If set".

test/print_graph_tests.cc
```diff
@@ -90,11 +90,11 @@ TEST_CASE("command graph printing is unchanged", "[print_graph][command-graph]")
 	    "shape=box];}id_0_0->id_0_1[color=orchid];id_0_1->id_0_2[];id_0_3->id_0_2[];id_0_3[label=<C3 on N0<br/><b>reduction</b> R1<br/> B0 {[0,0,0] - "
 	    "[1,1,1]}> fontcolor=black "
 	    "shape=ellipse];id_0_1->id_0_3[];id_0_4->id_0_3[];id_0_5->id_0_3[color=limegreen];id_0_6->id_0_3[color=limegreen];id_0_7->id_0_3[color=limegreen];id_0_"
-	    "4[label=<C4 on N0<br/>(R1) <b>await push</b> transfer 8589934592 <br/>BB0 {[0,0,0] - [1,1,1]}> fontcolor=black "
-	    "shape=ellipse];id_0_0->id_0_4[color=orchid];id_0_5[label=<C5 on N0<br/>(R1) <b>push</b> transfer 8589934593 to N1<br/>BB0 [0,0,0] + [1,1,1]> "
-	    "fontcolor=black shape=ellipse];id_0_1->id_0_5[];id_0_6[label=<C6 on N0<br/>(R1) <b>push</b> transfer 8589934594 to N2<br/>BB0 [0,0,0] + [1,1,1]> "
-	    "fontcolor=black shape=ellipse];id_0_1->id_0_6[];id_0_7[label=<C7 on N0<br/>(R1) <b>push</b> transfer 8589934595 to N3<br/>BB0 [0,0,0] + [1,1,1]> "
-	    "fontcolor=black shape=ellipse];id_0_1->id_0_7[];}";
+	    "4[label=<C4 on N0<br/>(R1) <b>await push</b> T2.B0.R1 <br/>B0 {[0,0,0] - [1,1,1]}> fontcolor=black "
```
PeterTh, line 93: Nice to get rid of the `BB`.

PeterTh (approved):
LGTM! Very minor notes.

include/types.h
fknorr: Now we don't.

include/types.h
fknorr: Improved.

fknorr:
From in-person discussion: We're leaving the name as-is for the moment to avoid rebase pain in the IDAG branch, and in the hope that documentation will make up for the imperfect name.

PR #233, opened 2023-12-06
CI: Let clang-tidy auto-detect config files

psalz:
clang-tidy-review v0.14.0 added support for this; it allows us to use different files for different folders.

Based on #232, so that should be merged first.


psalz:
Closing as I had to revert back to clang-tidy-review@0.9.0 for #232 (0.14.0 fails when installing pip dependencies on Ubuntu 22.04). Revisit this at some point.

PR #234, opened 2023-12-08
Capture and inspect all logs in Catch2 tests, fix race

fknorr:
This PR completely redefines how logging is handled inside unit tests.

- fixes a long-standing data race when capturing and inspecting logs inside a test
- silences logs by default, printing them post-mortem when a test fails
- detects if an unexpected warning or error is logged and fails the test

It works by unconditionally capturing logs in the global thread-safe `test_utils::log_test_capture`. This allows the test code to query captured messages, and will echo the captured logs through a global Catch2 hook if the test fails. By default, this mechanism will fail any test that logs a warning or error, but it provides functions to relax that behavior either explicitly through `test_utils::allow_max_log_level` or implicitly through `runtime_fixture` / `device_queue_fixture` which permit system-dependent warnings to appear.

The new mechanism is fully independent of the `CELERITY_LOG_LEVEL` environment, both in which messages it captures and also which ones it dumps in case of a test failure.

Some refactorings and fixes happened alogn the way:

- Simplified log.h: `log_context` and its printing was unnecessarily complex
- Moved std::tuple utils from log.h to utils.h
- Make logging in runtime and from `utils::panic` independent of LOG_LEVEL when called from tests
- Fixed some uninit-read bugs in tests that were surfaced by the new log-level restrictions


test/test_utils_tests.cc
```diff
@@ -208,3 +208,16 @@ TEST_CASE("most operations fail when called on an empty query", "[command_query]
 	CHECK_THROWS_WITH(query.have_successors(command_query_testspy::create_for({})), "Operation 'have_successors' not allowed on empty query");
 	CHECK_THROWS_WITH(query.have_type(command_type::epoch), "Operation 'have_type' not allowed on empty query");
 }
+
+TEST_CASE("tests that log any message in excess of level::info fails by default", "[test_utils][log][!shouldfail]") { CELERITY_WARN("spooky message!"); }
```
PeterTh, line 212: Should be either "test that logs" or "fail".

PeterTh (commented):
Nice. It's a bit silly that this took so much infrastructure to do, but it's great to be rid of that race condition.

test/runtime_tests.cc
```diff
@@ -1462,37 +1477,34 @@ namespace detail {
...
+		distr_queue q;
+		buffer<int, 1> buf(1);
 
-			SECTION("in device kernels") {
-				q.submit([&](handler& cgh) {
-					accessor acc(buf, cgh, celerity::access::all(), celerity::read_only);
-					cgh.parallel_for(range(1), [=](item<1>) { (void)acc; });
-				});
-			}
+		std::string expected_error_message;
```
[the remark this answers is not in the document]
fknorr: Done.

test/test_utils_tests.cc
fknorr: You're right! Fixed.

include/log.h
```diff
@@ -25,68 +21,41 @@
...
-		static_assert(std::tuple_size_v<Tuple> % 2 == 0, "an even number of entries is required");
-		tuple_for_each_pair_impl(tuple, cb, std::make_index_sequence<std::tuple_size_v<Tuple>>{});
+	template <typename... Es>
+	explicit log_context(const std::tuple<Es...>& entries) {
+		static_assert(sizeof...(Es) % 2 == 0, "log_context requires key/value pairs");
+		int i = 0;
+		celerity::detail::utils::tuple_for_each_pair(entries, [&](auto& a, auto& b) {
+			if(i++ > 0) { fmt::format_to(std::back_inserter(repr), ", "); }
+			fmt::format_to(std::back_inserter(repr), "{}={}", a, b);
+		});
```
psalz, line 41:
```suggestion
		std::string keys_and_values;
		celerity::detail::utils::tuple_for_each_pair(entries, [&](auto& a, auto& b) {
			if(i++ > 0) { fmt::format_to(std::back_inserter(keys_and_values), ", "); }
			fmt::format_to(std::back_inserter(keys_and_values), "{}={}", a, b);
		});
		if(i > 0) { repr = fmt::format("[{}] ", keys_and_values); }
```

Square brackets and final space went missing!

psalz (changes requested):
Amazing work! I've added one small note. Also you could maybe consider squashing some of the commits before merging.

The one concern I have with this design (which is already infinitely better than what we had before!) is that by setting e.g. `allow_max_log_level(detail::log_level::warn)`, a test might inadvertently swallow some additional, unexpected warnings. There's already a mechanism that could kind of deal with that (`allow_higher_level_log_messages`), but then we'd have to repeat the message twice (once to allow it, and once to check that it actually was emitted). I'm just spitballing here, and this doesn't have to be in v1, but maybe something like this could work?

```c++
TEST_CASE("my test") {
  test_utils::log_inspector logs;
  logs.allow_max_level(detail::log_level::info); // no-op as info would be the default
  // do stuff
  CHECK(logs.contains(detail::log_level::warn, "some warning"));
  CHECK(logs.contains(detail::log_level::error, "some error"));
  // if there's any unconsumed messages > info once log_inspector goes out of scope, fail test in destructor.
}
```

include/log.h
fknorr: Done!

fknorr:
I agree, finer control over which warnings to allow would be great in a follow-up PR!

psalz (approved):

include/log.h
```diff
@@ -25,68 +21,45 @@
 #define CELERITY_ERROR(...) CELERITY_LOG(::celerity::detail::log_level::err, __VA_ARGS__)
 #define CELERITY_CRITICAL(...) CELERITY_LOG(::celerity::detail::log_level::critical, __VA_ARGS__)
 
-namespace celerity {
-namespace detail {
+namespace celerity::detail {
 
-	using log_level = spdlog::level::level_enum;
+using log_level = spdlog::level::level_enum;
+using spdlog::should_log;
```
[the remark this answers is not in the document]
fknorr: It is, in `runtime.cc`. :shrug:

PeterTh (approved):

PR #236, opened 2023-12-20
Introduce new CELERITY_PRINT_GRAPHS env var

psalz:
...to control whether task and command graphs are printed to stdout at the end of execution.

### DIFF
--- a/src/runtime.cc
+++ b/src/runtime.cc
@@ -149,7 +149,7 @@ namespace detail {
 		m_reduction_mngr = std::make_unique<reduction_manager>();
 		m_host_object_mngr = std::make_unique<host_object_manager>();
 
-		if(m_cfg->is_recording()) m_task_recorder = std::make_unique<task_recorder>(m_buffer_mngr.get());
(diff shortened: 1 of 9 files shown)


fknorr (approved):
LGTM!

src/runtime.cc
```diff
@@ -149,7 +149,7 @@ namespace detail {
 		m_reduction_mngr = std::make_unique<reduction_manager>();
 		m_host_object_mngr = std::make_unique<host_object_manager>();
 
-		if(m_cfg->is_recording()) m_task_recorder = std::make_unique<task_recorder>(m_buffer_mngr.get());
+		if(m_cfg->should_print_graphs()) m_task_recorder = std::make_unique<task_recorder>(m_buffer_mngr.get());
```
PeterTh, line 152: I'd prefer if this was still controlled by a configuration specifically referring to recording in terms of naming (especially in light of soon-upcoming future uses of recording).

psalz: I've added `config::should_record()` as (for now) an alias to `should_print_graphs()`.

PR #238, opened 2024-01-23
Add support for SimSYCL (and GCC / C++20 / Sanitizers)

fknorr:
This adds [SimSYCL](https://github.com/celerity/SimSYCL) as a third supported SYCL implementation. SimSYCL executes kernels synchronously on the host and therefore has support for AddressSanitizer and LeakSanitizer, which we can now use to find UB in Celerity from within our CI.

SimSYCL requires C++20, so our code had to be adapted in a few places and 3rd-party dependencies had to be upgraded to be compatible with the new standard.

Until now all our CI builds have been made with Clang derivatives, so I opted for performing the SimSYCL build with GCC to broaden our compiler support at the same time. This required some further patches, either to work around GCC bugs or to be compliant where Clang was overly permissive.

Finally, SimSYCL + AddressSanitizer actually managed to find two use-after-scope errors in the Celerity tests, which were fixed.

### Possible follow-up changes

- [ ] Add Windows / macOS CI with SimSYCL
- [ ] Use SimSYCL instead of hipSYCL images for clang-tidy
- [ ] Simulate multiple different systems, heterogeneous GPU configurations, ...
- [ ] Test device selection with SimSYCL instead of the current mocking setup


include/ranges.h
```diff
@@ -278,8 +277,6 @@ class range : public detail::coordinate<range<Dims>, Dims> {
 	}
 
   private:
-	friend class detail::coordinate<range<Dims>, Dims>;
```
PeterTh, line 281: Why do we no longer need this friend decl?

include/user_bench.h
```diff
@@ -74,8 +74,8 @@ namespace experimental {
 				void end_section(const std::string& name);
 
 				template <typename... Args>
-				void log(const std::string& format_string, Args&&... args) const {
-					spdlog::default_logger_raw()->log(spdlog::level::info, "[user] " + format_string, std::forward<Args>(args)...);
```
PeterTh, line 78: What's the reasoning for this change?

include/workaround.h
```diff
@@ -11,6 +11,15 @@
...
 
+#if defined(CELERITY_SIMSYCL)
+#define CELERITY_WORKAROUND_SIMSYCL 1
+#define CELERITY_WORKAROUND_VERSION_MAJOR SIMSYCL_VERSION_MAJOR
+#define CELERITY_WORKAROUND_VERSION_MINOR SIMSYCL_VERSION_MINOR
+#define CELERITY_WORKAROUND_VERSION_PATCH SIMSYCL_VERSION_PATCH
+#else
+#define CELERITY_WORKAROUND_SIMSYCL 0
+#endif
+
```
PeterTh, line 22: I might have missed it, but are any of these actually used?

If not, I wouldn't include them.

PeterTh (commented):
Very nice! Brings us much closer to providing a Celerity development experience that works almost "everywhere".

.github/workflows/celerity_ci.yml
```diff
@@ -167,12 +173,20 @@ jobs:
...
+      # by /root/capture-backtrace.sh (otherwise we would also preload for other processes like bash, which we don't want to do).
+      # Also, hwloc (used by OpenMPI) will attempt to dlopen(RTLD_DEEPBIND) for GPU backend discovery which is incompatible
+      # with sanitizers, but which is also irrelevant for a SimSYCL run, so we set HWLOC_COMPONENTS accordingly.
+      - name: Prepare Sanitizer run
+        if: matrix.sycl == 'simsycl' && matrix.build-type == 'Debug'
         run: |
-          echo "ONEAPI_DEVICE_SELECTOR=level_zero:*" >> $GITHUB_ENV
+          echo "ASAN_OPTIONS=detect_leaks=1" >> $GITHUB_ENV
+          echo "LSAN_OPTIONS=exitcode=23:suppressions=$GITHUB_WORKSPACE/ci/lsan.suppressions" >> $GITHUB_ENV
+          echo "int dlclose(void *p) { return 0; }" | cc -x c -shared -o libfake_dlclose.so -
```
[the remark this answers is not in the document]
fknorr: I was considering CMake target and everything for that, but since it has nothing to do with Celerity itself I went with this one-liner :shrug:

include/ranges.h
fknorr: It was used to allow the base class to default-construct `range` for every dimensionality, but the constructor is now public. The thing you actually noticed was probably that I forgot to remove the private default constructor (now fixed).

include/user_bench.h
fknorr: When building with C++20, fmtlib does compile-time format argument verification, so you can't just pass a `std::string` as the format argument, it either has to be an `fmt::format_string` which can be constexpr-constructed from a literal, or be wrapped in `fmt::runtime`. I went for the former.

include/workaround.h
PeterTh: It just feels a bit depressing to me to already implement machinery for manual workarounds when none are needed right now.
(Particularly in the case of SimSYCL, where one would hope that it would never need workarounds)

README.md
```diff
@@ -78,6 +78,7 @@ installed first.
 - A supported SYCL implementation, either
     - [hipSYCL](https://github.com/illuhad/hipsycl),
     - [DPC++](https://github.com/intel/llvm)
```
PeterTh, line 80:
```suggestion
    - [DPC++](https://github.com/intel/llvm), or
```

This is a good opportunity to fix something that bugged me about the readme ;)

include/workaround.h
fknorr: Alright, I got rid of them. Also none of the version-specific workaround macros are currently used either because CCPP was the only SYCL implementation we actually upgraded in full-version steps.

README.md
fknorr: Done.

fknorr:
Github Actions seems to mark this PR as failing :x: when the SimSYCL runs emit any warning, so I resolved all of them:
- GCC was complaining about `memcpy`'ing over `hydration_id` which is not a POD type. I've used the opportunity to re-define all numerical `_id` types as individual POD structs and get rid of the `PhantomType` template (which I always found annoying because compiler messages would read `celerity::detail::PhantomType<size_t, celerity::blah_blah_node_id>` instead of just `celerity::detail::node_id`)
- I bumped the `small_vector` dependency to the latest `HEAD` to fix a GCC `-Wmaybe-uninitialized` in tests that were copy-constructing `small_vector`s
- ... and also some minor valid complaints by GCC

Curiously, the hipSYCL builds still emit `-Wunknown-cuda-version` but this does not seem to affect the build status.

test/grid_tests.cc
```diff
@@ -37,7 +37,7 @@ TEST_CASE("split_box dissects boxes as expected - 3d", "[grid]") {
 
 template <int MergeDim, int Dims>
 void test_directional_merge(box_vector<Dims> unmerged, box_vector<Dims> merged) {
-	CAPTURE(MergeDim);
+		CAPTURE(MergeDim);
```
PeterTh, line 40: ?

PeterTh (approved):
LGTM, I'd maybe squash some of the commits before merging (e.g. all the dependency version bumps).

test/grid_tests.cc
fknorr: Our clang-format check is broken again. Fantastic!

.github/workflows/celerity_ci.yml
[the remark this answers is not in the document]
psalz: Haha this is great!

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.6.0] - ???
```
psalz, line 9:
```suggestion
## [Unreleased]
```

CMakeLists.txt
```diff
@@ -129,13 +149,20 @@ option(CELERITY_USE_MIMALLOC "Use the mimalloc memory allocator" ${CELERITY_USE_
 include(FetchContent)
 
 macro(fetch_content_from_submodule DEPNAME RELPATH)
-  if(NOT EXISTS "${PROJECT_SOURCE_DIR}/${RELPATH}/CMakeLists.txt")
-    message(FATAL_ERROR "The git submodule ${RELPATH} is missing.\nTry running `git submodule update --init`.")
+  if(${DEPNAME}_FOUND)
```
psalz, line 152: What is this about?

src/backend/CMakeLists.txt
```diff
@@ -1,19 +1,25 @@
...
+    find_package(CUDAToolkit)
+    set(CUDA_OR_CUDA_TOOLKIT_FOUND ${CUDAToolkit_FOUND})
+  else()
+    find_package(CUDA)
+    set(CUDA_OR_CUDA_TOOLKIT_FOUND ${CUDA_FOUND})
+  endif()
 endif()
 
-find_package(CUDAToolkit QUIET)
+# find_package(CUDAToolkit QUIET) # NOCOMMIT: Why does this line exist if we already found_package above?
```
psalz, line 12:
```suggestion
```

That's an oversight from when I added the backwards compatibility workaround!

psalz (approved):
Very cool!

CMakeLists.txt
fknorr: SimSYCL transitively pulls in libenvpp to parse its own environment, so this skips Celerity's `fetch_content` if the dependency is already satisfied (by SimSYCL or by a CMake project using Celerity as its dependency for example).

Hoping that the previously found dependency is a match *could* lead to version conflicts which I hope will be debuggable for the user from the `Using system $dep from $path` messages.

psalz: Ah, I had a similar issue with `fmt` recently - I like it!

PR #240, opened 2024-01-29
Report code coverage via coveralls.io

fknorr:
This PR
1. Instruments the simsycl-debug CI build with GCC's `--coverage` flag to export coverage data (one `.gcno` file for every `.o` file)
2. Runs unit tests and system tests (generating one `.gcda` file with counters per `.gcno` file)
3. Aggregates the `.gcda` files using [fastcov](https://github.com/RPGillespie6/fastcov) into an [LCOV](https://github.com/linux-test-project/lcov) info file
4. Uploads the info file to [coveralls.io](https://coveralls.io/github/celerity/celerity-runtime) to generate statistics and a coverage badge (which is now in the README).

Note that the coverage badge still says "unknown" because it reports the master branch, so the badge will be correct after the merge.


psalz (approved):
At long last we have the coverage badge, nice! LGTM!

PR #243, opened 2024-02-07
CI: Complain if clang-format is not installed

psalz:
Turns out our formatting check has been failing for a while due to a yet to be resolved issue with the `clang-format` symlink to `clang-format-14` missing in the linting container. This adds a check to see whether `clang-format` is available.

### DIFF
--- a/ci/find-unformatted-files.sh
+++ b/ci/find-unformatted-files.sh
@@ -1,7 +1,14 @@
 #!/bin/bash
 
+set -o pipefail -o nounset
+
 if [[ ! -d src ]]; then
     echo "Warning: This script should be run from within the repository root directory" >&2
+fi
+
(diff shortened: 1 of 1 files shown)


fknorr:
I believe we can catch this error more naturally and without searching for `clang-format` first by setting `-o errexit -o pipefail`. We just have to make sure not to die on a non-zero exit code from grep when it doesn't match anything (errexit is suppressed in a command that is used as the condition of an if statement):

```bash
set -o errexit -o nounset -o pipefail
# ...
if clang-format -output-replacements-xml "$s" | grep -qvE "<(/?replacements|\?xml)"); then
    echo "$s"
fi
```

fknorr (approved):
LGTM - although I would wait with the merge until we have the CI images fixed.

PR #246, opened 2024-02-09
Explicitly manage buffer / host object lifetimes in graph generation

fknorr:
This is a back-port from IDAG development and a second attempt at #216.

`task_manager` and `distributed_graph_generator` already maintain state for each buffer (`add_buffer()`) and host object (implicit on creation) , but keep that state around indefinitely even after the buffer or host object in question is destroyed. Also, neither graph generator has access to buffer debug names and thus can't include that information in error reports (such as uninitialized-read detection, there is a TODO in the code about this).
    
This PR adds explicit member functions for tracking the creation and destruction of buffers (`create_buffer`, `destroy_buffer`) and host objects (`create_host_object`, `destroy_host_object`) to `task_manager`, `distributed_graph_generator`, `scheduler`; and now by necessity, `runtime`, which receives these requests directly instead of indirectly through `buffer_manager::buffer_lifetime_callback`.
    
It also removes the `recorder` -> `buffer_manager` dependency by replicating the buffer name (like all other metadata) in both graph generators. This foreshadows the eventual removal of `buffer_manager` post-IDAG.

I purposefully changed the recording API such that its user is responsible for constructing records and the recorder simply aggregates them. This is now possible because the recorders do not need to know about `task_manager` and `buffer_manager` anymore, and will allow us, in the future, to add debug info to records that is not present in the "real" DAG (see [the upcoming instruction graph generator](https://github.com/fknorr/celerity-runtime/blob/bad6eaf71c3f64a6889cddd8d90d18d0ca06655a/src/instruction_graph_generator.cc#L873) for how this might look).


include/distributed_graph_generator.h
```diff
@@ -153,9 +171,6 @@ class distributed_graph_generator {
 	// they are executed in the same order on every node.
 	std::unordered_map<collective_group_id, command_id> m_last_collective_commands;
 
-	// Side effects on the same host object create true dependencies between task commands, so we track the last effect per host object.
-	side_effect_map m_host_object_last_effects;
```
psalz, line 157: The `side_effect_map` alias on line 128 can be dropped then (which is good, because we already have a different type with the same name in task.h!).

include/runtime.h
```diff
@@ -118,8 +129,7 @@ namespace detail {
 		runtime(const runtime&) = delete;
 		runtime(runtime&&) = delete;
 
-		void handle_buffer_registered(buffer_id bid);
-		void handle_buffer_unregistered(buffer_id bid);
+		void register_buffer(buffer_id bid, const range<3>& range, bool host_initialized);
```
psalz, line 132: Presumably this function is required because of forward declarations? Maybe you could add a comment.

include/runtime.h
```diff
@@ -73,7 +71,20 @@ namespace detail {
 
 		reduction_manager& get_reduction_manager() const;
 
-		host_object_manager& get_host_object_manager() const;
+		template <typename DataT, int Dims>
+		buffer_id create_buffer(const range<3>& range, const DataT* host_init_ptr) {
+			const auto bid = m_buffer_mngr->register_buffer<DataT, Dims>(range, host_init_ptr);
```
psalz, line 76: I'm a bit confused regarding the terminology, why is it called "create_buffer" everywhere except for the buffer manager? (FWIW, I would find it more intuitive to have a single "create" function on the runtime, and "register" everywhere else - but not a strong opinion).

psalz (approved):
Nice, so this also solves the issue of having different names for a single buffer throughout the course of an application! I like the parity between distributed_graph_generator and task_manager. Also the `operator<<` on recorders looks neat!

include/distributed_graph_generator.h
fknorr: Done.

include/runtime.h
fknorr: It's so I don't need to include `task_manager.h` and `distributed_graph_generator.h` here. Comment added.

include/runtime.h
fknorr: Two reasons:
- `create_buffer` is actually creating the buffer from perspectives of `runtime`, `distributed_graph_generator` and `task_manager`, it does not hand these classes a reference to a buffer that exists elsewhere.
- There is a clear dichotomy between `create_*` and `destroy_*` (and with #249 , `destroy` will actually emit instructions to free allocations and destroy host objects).

I could have refactored the buffer manager as well to have `create` and `destroy` functions but I did not bother because that code is due for removal anyway.

fknorr:
I've gone ahead and renamed every `buffer_state` variable to `buffer` in the DGGen for consistency.

include/runtime.h
PeterTh: In an offline discussion just now we tried to get to the bottom of this, 2 important outcomes:

* In general, the true **creation** of the concept of a buffer/host object is associated in the runtime with it **receiving an ID**.
* In the long term, we should get rid of the side-channel for transporting any (*functional*) information, and instead include it in the graphs (i.e. the task graph would include all functional information required for the command graph, and the command graph for the instruction graph).

In the short term, one more idea regarding this naming issue: how about `*_created` and `*_destroyed`? That would imply to me that it is about the specific component being notified of that event occurring, rather than making the event happen.

src/distributed_graph_generator.cc
```diff
@@ -20,19 +20,37 @@ distributed_graph_generator::distributed_graph_generator(
 	// Build initial epoch command (this is required to properly handle anti-dependencies on host-initialized buffers).
 	// We manually generate the first command, this will be replaced by applied horizons or explicit epochs down the line (see
 	// set_epoch_for_new_commands).
-	auto* const epoch_cmd = cdag.create<epoch_command>(task_manager::initial_epoch_task, epoch_action::none);
+	const auto epoch_tsk = tm.get_task(task_manager::initial_epoch_task);
+	auto* const epoch_cmd = cdag.create<epoch_command>(epoch_tsk->get_id(), epoch_action::none);
 	epoch_cmd->mark_as_flushed(); // there is no point in flushing the initial epoch command
-	if(m_recorder != nullptr) m_recorder->record_command(*epoch_cmd);
+	if(m_recorder != nullptr) { *m_recorder << command_record(*epoch_cmd, *epoch_tsk, {}); }
```
PeterTh, line 26: The new design retrieves the task from the task manager here even if recording is disabled (not that it matters in this context, but it's unnecessary).

src/distributed_graph_generator.cc
```diff
@@ -92,7 +110,11 @@ std::unordered_set<abstract_command*> distributed_graph_generator::build_task(co
 	// If we have a command_recorder, record the current batch of commands
 	if(m_recorder != nullptr) {
 		for(const auto& cmd : m_current_cmd_batch) {
-			m_recorder->record_command(*cmd);
+			buffer_name_map accessed_buffers;
+			for(const auto bid : tsk.get_buffer_access_map().get_accessed_buffers()) {
+				accessed_buffers.emplace(bid, m_buffers.at(bid).debug_name);
+			}
+			*m_recorder << command_record(*cmd, tsk, accessed_buffers);
```
PeterTh, line 117: This creates some busywork related to recording in the "actual" code, which I would rather avoid.
I briefly thought about it, and one idea I see to avoid that would be to pass a functor from `buffer_id` to string to the command recording. I.e.
```C++
command_record(*cmd, tsk, [&](buffer_id bid) { return print_buffer_debug_label(bid); });
```

psalz: Or maybe more generically

```c++
*m_recorder << [&]() {
    buffer_name_map accessed_buffers;
    for(const auto bid : tsk.get_buffer_access_map().get_accessed_buffers())
        accessed_buffers.emplace(bid, m_buffers.at(bid).debug_name);
    }
    return command_record(*cmd, tsk, accessed_buffers);
};
```

We could then have a `null_recorder` in case recording is disabled, which just discards everything.

src/distributed_graph_generator.cc
fknorr: Fixed.

src/distributed_graph_generator.cc
fknorr: I like @PeterTh's approach, building and passing the `unordered_map` was unfortunate anyway since it was not even stored in the record afterwards. Replaced the type with a `std::function` to carry a lambda. It's still called `buffer_name_map` because it's still a mapping, just now in the mathematical sense.

include/recorders.h
```diff
@@ -38,79 +39,90 @@ struct dependency_record {
...
+	task_dependency_list dependencies;
 };
 
 class task_recorder {
   public:
-	using task_record = std::vector<detail::task_record>;
+	using task_records = std::vector<detail::task_record>;
 
-	task_recorder(const buffer_manager* buff_mngr = nullptr) : m_buff_mngr(buff_mngr) {}
+	friend task_recorder& operator<<(task_recorder& recorder, task_record&& record) {
```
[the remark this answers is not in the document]
fknorr: I disagree and believe it would actually help readability especially in IDAG generation. Changed nevertheless because I want this to move forward.

PeterTh (approved):
LGTM.

PR #248, opened 2024-02-12
CI: Use Docker image from ghcr.io for benchmarks

fknorr:
This update was missed when transitioning `celerity_ci.yml` to pull from `ghcr.io`.

This also switches us to the pinned version of hipSYCL / ACpp to ensure we do not mistake performance improvements / regressions in the SYCL implementation for ones in Celerity. ~~I've updated the benchmark results to set an appropriate baseline.~~


psalz (approved):

fknorr:
The benchmark results above are bogus due to a performance bug in our benchmarking suite - see #250. I have removed the new results from this PR after all.

PR #250, opened 2024-02-12
Fix: Disable logging in benchmarks

fknorr:
**Note:** This PR is based on #248.

In #234, we set the global log level to `trace` in all tests in order to print a full log in every failing test. Since we did not perform a benchmark run, we did not notice its impact on our benchmarking suite (see the benchmark report in #248): Task-graph generation emits trace-logs whenever testing for horizon generation, which is triggered for every command group submission. These logs are being written to the test log capture buffer which is very expensive compared to the actual task generation.

This PR explicitly shuts off all logging for the duration of any benchmark, re-activating the behavior that was implicitly true before #248 (because benchmarks were run with the default `LOG_LEVEL`).

I have re-run the benchmarks to commit a new baseline.


test/test_utils.cc
```diff
@@ -194,6 +194,16 @@ struct global_setup_and_teardown : Catch::EventListenerBase {
 		g_test_log_capture->print_log_if_nonempty(); // if so, this is likely due to a fatal signal
 		g_test_log_capture->clear();
 	}
+
+	void benchmarkPreparing(Catch::StringRef name) override {
+		log_level_before_benchmark = spdlog::get_level();
+		spdlog::set_level(spdlog::level::off);
```
psalz, line 200: Maybe we should add a note here why this is important!

psalz (approved):
Ahh, mystery solved - thanks! We should really do nightly benchmark runs...

test/test_utils.cc
fknorr: Done.

PR #253, opened 2024-05-23
[IDAG] Out-of-Order Engine for Instruction Dispatch

fknorr:
This is the third PR in the Instruction Graph series and introduces the new executor state machine, dubbed the out-of-order-engine, as a stand-alone testable component.

Much like the current executor, the out-of-order engine keeps track of all instructions that have not yet completed and decides which instructions to schedule onto which backend resources at what time, and receives back information on which instructions have already completed. This will allow us to keep the instruction executor free of most instruction state tracking.

Unlike the current approach, this new form of scheduling is based on a definition of [backends](https://github.com/fknorr/celerity-runtime/blob/f937aa52f4eb568095b886dbcb8ea63a7b7568e4/include/backend/backend.h) which maintain an array of [in-order thread queues](https://github.com/fknorr/celerity-runtime/blob/f937aa52f4eb568095b886dbcb8ea63a7b7568e4/src/thread_queue.h) for host work and in-order SYCL queues for device submissions. This allows the engine to omit host / executor loop round-trips between consecutive GPU / CPU loads by scheduling successors onto the same in-order queues to implicitly fulfil dependencies, and thus hide SYCL and CUDA kernel launch latency

In the future I would like to improve this further with support to submit instructions with dependencies on multiple queues / devices earlier by waiting on in-flight SYCL events.

This PR still does not touch the actual executor yet, allowing us to see the unit [test coverage](https://coveralls.io/builds/67663683/source?filename=src%2Fout_of_order_engine.cc) of the engine in isolation.

### DIFF
--- a/include/out_of_order_engine.h
+++ b/include/out_of_order_engine.h
@@ -1,0 +1,86 @@
+#include "types.h"
+
+#include <memory>
+#include <optional>
+
+namespace celerity::detail::out_of_order_engine_detail {
+struct engine_impl;
+}
+
+namespace celerity::detail {
+
+class instruction;
+struct system_info;
+
+/// State machine controlling when and in what manner instructions are assigned to execution resources in adherence to the dependency graph.
+///
+/// Based on their type, instructions can either be assigned to begin executing immediately as soon as all their predecessors are complete, or they can be
+/// enqueued eagerly in one of the backend's thread queues or in-order device queues to hide launch latencies where possible.
+class out_of_order_engine {
+  public:
+	/// Identifies a category of execution resources that may be distinguished further by a device- or lane id.
+	enum class target {
+		/// Execution can begin immediately, no queueing takes place in the backend and no lane is assigned. Used for low-overhead instructions that do not
+		/// profit from additional concurrency such as horizons, as well as for instructions where asynchronicity is managed outside the backend (p2p transfers
+		/// through communicator and receive_arbiter).
+		immediate,
+
+		/// The instruction shall be inserted to the backend's (singular) host allocation queue. No lane is assigned. Since at least CUDA serializes the slow
+		/// alloc / free operations anyway to update page tables globally, the added concurrency from multiple thread queues would not increase throughput. The
+		/// separation between alloc_queue and host_queues further enforces a host round-trip between every alloc_instruction and its first successor, which we
+		/// use to inform the executor of the newly allocated pointer for the purpose of accessor hydration.
+		alloc_queue,
+
+		/// The instruction shall be submitted to a backend thread queue identified by the lane id. Used for host tasks and host-to-host copies.
+		host_queue,
+
+		/// The instruction shall be submitted to a backend in-order device queue identified by the lane id. Used for device kernels and host-to-device /
+		/// device-to-device / device-to-host copies.
+		device_queue,
+	};
+
+	/// A lane identifies a thread queue for target::host_task and an in-order device queue with target::device_queue.
+	using lane_id = size_t;
+
+	/// Directions on how a single (ready) instruction is to be dispatched by the executor.
+	struct assignment {
+		const detail::instruction* instruction = nullptr;
+		out_of_order_engine::target target = out_of_order_engine::target::immediate;
+		std::optional<device_id> device; ///< Identifies the device to submit to (if target == device_queue) or to allocate on (if target == alloc_queue).
+		std::optional<lane_id> lane; ///< Identifies the thread queue (target == host_queue) or the in-order queue for the given device (target == alloc_queue).
+
+		assignment(const detail::instruction* instruction, const out_of_order_engine::target target, const std::optional<device_id> device = std::nullopt,
+		    const std::optional<lane_id> lane = std::nullopt)
+		    : instruction(instruction), target(target), device(device), lane(lane) {}
+	};
+
+	/// Constructor requires a `system_info` to enumerate devices and perform `memory_id -> device_id` mapping.
+	explicit out_of_order_engine(const system_info& system);
+
+	out_of_order_engine(const out_of_order_engine&) = delete;
+	out_of_order_engine(out_of_order_engine&&) = default;
+	out_of_order_engine& operator=(const out_of_order_engine&) = delete;
+	out_of_order_engine& operator=(out_of_order_engine&&) = default;
+
+	~out_of_order_engine();
+
+	/// True when all submitted instructions have completed.
+	bool is_idle() const;
+
+	/// Begin tracking an instruction so that it is eventually returned through `assign_one`. Must be called in topological order of dependencies, i.e. no
+	/// instruction must be submitted before one of its predecessors in the graph.
+	void submit(const instruction* instr);
+
+	/// Produce an assignment for one instruction for which either all predecessors have completed, or for which all incomplete predecessors can be implicitly
+	/// fulfilled by submitting to the same backend queue. If multiple instructions are eligible, assign the one with the highest priority.
+	[[nodiscard]] std::optional<assignment> assign_one();
+
+	/// Call once an instruction that was previously returned from `assign_one` has completed synchronously or asynchronously. For simplicity it is permitted
+	/// to mark assigned instructions as completed in any order, even if that would violate internal dependency order.
+	void complete_assigned(const instruction* instr);
+
+  private:
+	std::unique_ptr<out_of_order_engine_detail::engine_impl> m_impl;
+};
+
+} // namespace celerity::detail
--- a/include/system_info.h
+++ b/include/system_info.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "dense_map.h"
+#include "types.h"
+
+#include <bitset>
+
+namespace celerity::detail {
+
+/// Memory id for (unpinned) host memory allocated for or by the user. This memory id is assumed for pointers passed for buffer host-initialization and for the
+/// explicit user-side allocation of a buffer_snapshot that is performed before a buffer fence.
+inline constexpr memory_id user_memory_id = 0;
+
+/// Memory id for (pinned) host memory that the executor will obtain from the backend for buffer allocations and staging buffers.
+inline constexpr memory_id host_memory_id = 1;
+
+/// Memory id for the first device-native memory, if any.
+inline constexpr memory_id first_device_memory_id = 2;
+
+static constexpr size_t max_num_memories = 64;
+using memory_mask = std::bitset<max_num_memories>;
+
+/// Information about a single device in the local system.
+struct device_info {
+	/// Before accessing any memory on a device, instruction_graph_generator will prepare a corresponding allocation on its `native_memory`. Multiple
+	/// devices can share the same native memory. No attempts at reading from peer or shared memory to elide copies are currently made, but could be in the
+	/// future.
+	memory_id native_memory;
+};
+
+/// Information about a single memory in the local system.
+struct memory_info {
+	/// This mask contains a 1-bit for every memory_id that the associated backend queue can copy data from or to directly. instruction_graph_generator
+	/// expects this mapping to be reflexive, i.e. `system_info::memories[a].copy_peers[b] == system_info::memories[b].copy_peers[a]`.
+	/// Further, copies must always be possible between `host_memory_id` and `user_memory_id` as well as between `host_memory_id` and every other memory.
+	/// instruction_graph_generator will create a staging copy in host memory if data must be transferred between two memories that are not copy peers.
+	memory_mask copy_peers;
+};
+
+/// All information about the local system that influences the generated instruction graph.
+struct system_info {
+	dense_map<device_id, device_info> devices;
+	dense_map<memory_id, memory_info> memories;
+};
+
+} // namespace celerity::detail
--- a/src/out_of_order_engine.cc
+++ b/src/out_of_order_engine.cc
@@ -1,0 +1,444 @@
+#include "out_of_order_engine.h"
+#include "dense_map.h"
+#include "instruction_graph.h"
+#include "system_info.h"
+#include "utils.h"
+
+#include <queue>
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+#include <matchbox.hh>
+
+namespace celerity::detail::out_of_order_engine_detail {
+
+using target = out_of_order_engine::target;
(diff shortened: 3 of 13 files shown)


include/system_info.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "dense_map.h"
+#include "types.h"
+
+#include <bitset>
+
```
PeterTh, line 7: I noticed that `host_memory_id` and `user_memory_id` are currently defined in `types.h`. Semantically I feel like it might make more sense to define them in this file, but this is not a big deal either way and if there are any dependency constraints preventing that then ignore this point.

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+
+#include <matchbox.hh>
+
+namespace celerity::detail::out_of_order_engine_detail {
+
+using target = out_of_order_engine::target;
+using lane_id = out_of_order_engine::lane_id;
+using assignment = out_of_order_engine::assignment;
+
+/// Comparision operator to make priority_queue<instruction*> return concurrent instructions in decreasing priority.
```
PeterTh, line 20: Typo `Comparision`.

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+struct instruction_priority_less {
+	bool operator()(const instruction* lhs, const instruction* rhs) const { return lhs->get_priority() < rhs->get_priority(); }
+};
+
+/// Instruction is not eligible for assignment yet. Implies `num_incomplete_predecessors > 0`.
+struct unassigned_state {
```
PeterTh, line 26: This type name uses `assigned`, while the others use `assignable`. The comment also talks in terms of being *eligible* for assignment. Should this be `unassignable_state`, or is there an actual semantic implication?

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+	/// Flag allowing us to shortcut the eager-assignment check if we know that it cannot succeed (anymore). Set to false statically for the immediate and
+	/// alloc_queue target, and also set to false if the instruction was in `eagerly_assignable_state` but eager assignment had to be aborted because the lane
+	/// was not in the expected state anymore.
+	bool probe_for_eager_assignment = true;
+};
+
+/// Instruction is inserted into `assignment_queue` after being popped might be assignable by eager enqueueing to the lane where all its remaining predecessors
```
PeterTh, line 33: I think there's a word missing here.

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+/// State maintained for every "multi-lane" target, i.e., `target::host_queue` and `target::device_queue[num_devices]`.
+struct target_state {
+	std::vector<lane_state> lanes;
+};
+
+// Implementation is behind a pimpl because we expect it to grow in complexity when adding support for waiting on inter-queue events to increase scheduling
+// "eagerness" when an instruction needs to wait on multiple incomplete predecessors. This can already be implemented with AdaptiveCpp using the
+// enqueue_custom_operation extension, but requires a similar interop story in DPC++, see https://github.com/intel/llvm/issues/13706 .
+struct engine_impl {
+	system_info system; // immutable
```
PeterTh, line 86: Why not `const`?

src/utils.cc
```diff
@@ -61,6 +61,11 @@ std::string escape_for_dot_label(std::string str) {
 	return str;
 }
 
+[[noreturn]] void unreachable() {
+	assert(!"executed unreachable code");
```
PeterTh, line 65: Should this do something like `__builtin_unreachable();`?

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+	}
+}
+
+lane_state& engine_impl::get_lane_state(const target tgt, const std::optional<device_id>& device, const lane_id lane) {
+	return get_target_state(tgt, device).lanes.at(lane);
+}
+
+lane_id engine_impl::get_free_lane_id(const target tgt, const std::optional<device_id>& device) {
+	auto& target_state = get_target_state(tgt, device);
+	for(lane_id lid = 0; lid < target_state.lanes.size(); ++lid) {
```
PeterTh, line 141: I assume we expect the number of lanes per target to remain rather small, so doing a linear search should be fine. But how small exactly, and how do we ensure that?

src/out_of_order_engine.cc
PeterTh: After reading more of the implementation, I think the type name is correct but maybe the comment should be updated. As far as I can tell, an instruction in this state isn't necessarily "not eligible for assignment yet", it just hasn't been assigned yet?

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,414 @@
...
+	    },
+	    [&](const device_kernel_instruction& dkinstr) {
+		    node.target = target::device_queue;
+		    node.eligible_devices.push_back(dkinstr.get_device_id());
+	    },
+	    [&](const host_task_instruction& htinstr) { //
+		    node.target = target::host_queue;
+	    },
+	    [&](const auto& /* other */) { //
+		    node.target = target::immediate;
```
PeterTh, line 260: Is it a good idea to have basically a fallthrough case here?
I'm not sure how many individual cases would have to be covered, but this way someone adding a new instruction type might overlook what happens to it in the OoO Engine.

test/test_utils.cc
```diff
@@ -243,29 +243,29 @@ const char* const expected_device_enumeration_warnings_regex =
...
 		if(supports_d2d_copies) {
-			for(memory_id peer = first_device_memory_id; peer < info.memories.size(); ++peer) {
-				info.memories[mid].copy_peers.set(peer);
+			for(memory_id peer = first_device_memory_id; peer < system.memories.size(); ++peer) {
+				system.memories[mid].copy_peers.set(peer);
 			}
 		}
 	}
-	return info;
+	return system;
```
PeterTh, line 268: This local rename touches a lot of lines for no real reason.

PeterTh (commented):
Conceptually neat, with one point that I'm concerned about regarding scaling -- but mostly my comments just relate to my understanding of what's going on.

include/system_info.h
fknorr: I agree. `types.h` is kind of a kitchen sink anyway. It defines a lot of types that are not conceptually related simply because they are all integer aliases.

src/out_of_order_engine.cc
fknorr: My spell checker was okay with that for some reason. Fixed.

src/out_of_order_engine.cc
fknorr: I try to follow [this Guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref) in all of my code

src/out_of_order_engine.cc
fknorr: Re-worded.

src/utils.cc
fknorr: It could be an inline function that does
```c++
assert(!"executed unreachable code");
__builtin_trap();
```
Question is, does it matter how this is implemented?

fknorr: But your suggestion kinda does makes me want to use more intrinsics in the codebase, especially `__builtin_ia32_pause()` in the executor loop :P

test/test_utils.cc
fknorr: But it's not very hard to review, is it? :)

src/out_of_order_engine.cc
fknorr: It's kinda ugly enumerating all these, but I agree that there should be a compile-time error when somebody forgets to update this. Fixed.

src/out_of_order_engine.cc
fknorr: There is a difference:

- `unassigned` means "not ready for immediate or eager assignment"
- `eagerly assignable` means "can be submitted to an in-order queue / thread queue to implicitly fulfill dependencies, but it is conditional (now renamed for that reason)
- `assigned` means "has been returned from `assign_one`" and is waiting for a call to `complete_assigned`.

I improved the documentation.

test/out_of_order_engine_tests.cc
```diff
@@ -0,0 +1,562 @@
...
+	const auto copy_indep3 = octx.copy({d3_k0 /* already complete */}, first_device_memory_id, first_device_memory_id + 3);
+
+	{
+		const auto iq = octx.assign_all();
+		CHECK(iq.assigned_instructions_are({copy_indep0, copy_indep1, copy_indep2, copy_indep3}));
+		CHECK((iq.queued_on_device(copy_indep0, device_id(1)) || iq.queued_on_device(copy_indep0, device_id(0))));
+		CHECK((iq.queued_on_device(copy_indep1, device_id(0)) || iq.queued_on_device(copy_indep1, device_id(1))));
+		CHECK((iq.queued_on_device(copy_indep2, device_id(1)) || iq.queued_on_device(copy_indep2, device_id(2))));
+		CHECK((iq.queued_on_device(copy_indep3, device_id(0)) || iq.queued_on_device(copy_indep3, device_id(3))));
+	}
```
psalz, line 280: What is this testing? It's not eager submission since the dependencies are already completed?

psalz (approved):
LGTM! Nice work!

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,430 @@
...
+
+	assigned_state assigned;
+	if(const auto& eagerly_assignable = std::get_if<conditional_eagerly_assignable_state>(&node.assignment)) {
+		// After an instruction is popped from the assignment queue, "eager-assignability" is unconditional. Force-assign to the same lane to implicitly fulfill
+		// all remaining dependencies through queue ordering.
+		assigned.device = eagerly_assignable->device;
+		assigned.lane = eagerly_assignable->lane;
+		assert(
+		    assigned.lane.has_value()
+		    && get_lane_state(node.target, assigned.device, *assigned.lane).last_incomplete_submission == eagerly_assignable->expected_last_submission_on_lane);
```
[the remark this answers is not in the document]
fknorr: Good catch - it can be which causes the eager submission to be retracted, so it was inconsequential, but the invariant should still be that this value is never null in the eagerly-assignable state. I've added an appropriate early-return in `try_mark_for_assignment` to catch this condition early and added a test.

src/out_of_order_engine.cc
```diff
@@ -0,0 +1,430 @@
...
+			assert(node.num_incomplete_predecessors > 0); // otherwise this would be an immediately_assignable_state
+			const auto& lane = get_lane_state(node.target, eagerly_assignable_when_pushed->device, eagerly_assignable_when_pushed->lane);
+			if(lane.last_incomplete_submission == nullptr
+			    || lane.last_incomplete_submission == eagerly_assignable_when_pushed->expected_last_submission_on_lane) {
+				// Our preferred lane is still in the required state to go through with eager assignment
+				return &node;
+			} else {
+				// A third instruction has been submitted to our preferred lane since, so we drop eager assignment and don't need to attempt it again, since
+				// none of our dependencies can now end up last in the queue anymore.
+				node.assignment.emplace<unassigned_state>().probe_for_eager_assignment = false;
```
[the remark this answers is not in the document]
fknorr: That's what we hope for, exactly. Going through with eager assignment in this case means we would serialize instructions that can otherwise be concurrent.

include/out_of_order_engine.h
```diff
@@ -0,0 +1,86 @@
...
+		/// Execution can begin immediately, no queueing takes place in the backend and no lane is assigned. Used for low-overhead instructions that do not
+		/// profit from additional concurrency such as horizons, as well as for instructions where asynchronicity is managed outside the backend (p2p transfers
+		/// through communicator and receive_arbiter).
+		immediate,
+
+		/// The instruction shall be inserted to the backend's (singular) host allocation queue. No lane is assigned. Since at least CUDA serializes the slow
+		/// alloc / free operations anyway to update page tables globally, the added concurrency from multiple thread queues would not increase throughput. The
+		/// separation between alloc_queue and host_queues further enforces a host round-trip between every alloc_instruction and its first successor, which we
+		/// use to inform the executor of the newly allocated pointer for the purpose of accessor hydration.
+		alloc_queue,
```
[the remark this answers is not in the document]
fknorr: cudaMallocAsync is incompatible with RDMA (see https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/) and requires us to think about CUDA-supplied memory pools. I'd rather avoid it until we know all requirements for graph-based memory allocation.

test/out_of_order_engine_tests.cc
fknorr: Pretty much that: Copies are immediately assigned if their dependencies have completed before.

src/out_of_order_engine.cc
fknorr: You got me there. But if I make it const clang-tidy will complain for the same exact reason :P

include/system_info.h
PeterTh: True, but `host_memory_id` etc. aren't even types!

src/out_of_order_engine.cc
PeterTh: Does it complain if you explicitly make it non-copyable?

src/utils.cc
PeterTh: The point is that `__builtin_unreachable()` and its equivalents might transport some information to the compiler which it can use in optimization.

test/test_utils.cc
PeterTh: But it pointlessly touches the history of those lines.

src/utils.cc
fknorr: Sure, but none of its usages are performance sensitive enough to warrant UB in a release build.

src/out_of_order_engine.cc
fknorr: It doesn't!

test/test_utils.cc
fknorr: I'm opinionated on this as you can tell, but not enough to delay cake allocation

PeterTh (approved):
LGTM! (with squashing)

PR #254, opened 2024-06-17
Unconditionally disable reductions for hipSYCL

fknorr:
hipSYCL aka AdaptiveCpp has recently merged [reduction support](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1453) which is not 1:1 compatible with [an earlier attempt](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/578) to implement the API (now defunct), which we were detecting in CMake to conditionally expose it in Celerity.

This PR removes the check and unconditionally disables reduction support on hipSYCL for now to unblock CI. It will be re-enabled as part of the final IDAG merge when all accompanying issues have been resolved.

This PR also silences a deprecation warning to avoid bumping the minimum hipSYCL version.


psalz (approved):
LGTM

PeterTh (approved):

PR #256, opened 2024-06-24
Fix runtime test for task deletion on horizons

fknorr:
This fixes a test by @PeterTh that got broken when epochs were introduced: The task manager must delete tasks behind the effective-epoch, and that condition was awaited in the main thread through `slow_full_sync()`. That function however inserts an epoch which force-deletes all tasks before it, making the test pass even if horizons never delete tasks at all.

This PR replaces the calls to `slow_full_sync()` with manual synchronization to ensure we still test the actual horizons behavior.

### DIFF
--- a/test/runtime_tests.cc
+++ b/test/runtime_tests.cc
@@ -1002,36 +1002,44 @@ namespace detail {
 	TEST_CASE_METHOD(test_utils::runtime_fixture, "tasks behind the deletion horizon are deleted", "[task_manager][task-graph][task-horizon]") {
 		using namespace cl::sycl::access;
 
+		constexpr int horizon_step_size = 2;
+
 		distr_queue q;
 		auto& tm = runtime::get_instance().get_task_manager();
-		tm.set_horizon_step(2);
+		tm.set_horizon_step(horizon_step_size);
 
-		constexpr int extents = 16;
-
-		buffer<int, 1> buf_a(extents);
-		q.submit([&](handler& cgh) {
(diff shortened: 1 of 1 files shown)


PeterTh (approved):

test/runtime_tests.cc
```diff
@@ -1002,36 +1002,42 @@ namespace detail {
...
+				std::unique_lock lock(m);
+				cv.wait(lock, [&] { return completed_step == i; });
 			}
 
-			// need to wait for commands to actually be executed, otherwise no tasks are deleted
-			q.slow_full_sync();
-			CHECK(tm.get_current_task_count() < task_limit);
+			// There are 2 sets of `horizon_step_size` host tasks after the current effective epoch, 2 horizon tasks, plus up to `horizon_step_size` additional
+			// host tasks that will be deleted on the next submission.
+			constexpr int task_limit = 3 * horizon_step_size /* host tasks */ + 2 /* horizons */;
```
GagaLP, line 1039: I don't understand the calculation of the task limit here.

Especially the part with `3 * ...` and the inline code comments seem strange to me. Couldn't we make all the constants 2 and 3 constexpr, giving them descriptive names instead of using inline comments?

GagaLP (approved):
Some minor styling suggestions.

psalz (approved):
LGTM

fknorr:
🍰

test/runtime_tests.cc
fknorr: done.

PR #257, opened 2024-06-25
[IDAG] Optimal nd-memory copy primitives

fknorr:
This PR proposes to merge the nd-memory copy layouting mechanism used by the IDAG executor. It is set to replace `memcpy_strided_host` and the host-side portion of `memcpy_strided_device` that is currently implemented in `buffer_storage`.

Unlike the current implementation, it always detects when a copy between higher-dimensional allocations is "accidentally contiguous" (i.e. can fulfilled by a single memcpy) or when a copy between 3D allocations is "accidentally 2D". The unit contains a full implementation of host-to-host copies (which perform the minimum number of memcpys possible) and the means to implement optimized dispatch to `cudaMemcpy*D` (for this, see [cuda_backend.cc in the IDAG branch](https://github.com/fknorr/celerity-runtime/blob/ed402792e94b0db69fa79c618b7bf80f6c585e7f/src/backend/cuda_backend.cc#L23)).

### DIFF
--- a/include/nd_memory.h
+++ b/include/nd_memory.h
@@ -1,0 +1,122 @@
+#pragma once
+
+#include "grid.h"
+#include "ranges.h"
+
+#include <string.h>
+
+namespace celerity::detail {
+
+/// Describes a box-shaped copy operation between two box-shaped allocations in terms of linear offsets and strides.
+struct nd_copy_layout {
+	struct stride_dimension {
+		size_t source_stride = 1; ///< by how many bytes to advance the source pointer after one step in this dimension.
+		size_t dest_stride = 1;   ///< by how many bytes to advance the destination pointer after one step in this dimension.
+		size_t count = 1;         ///< how many iterations to perform in this dimension.
+
+		friend bool operator==(const stride_dimension& lhs, const stride_dimension& rhs) {
+			return lhs.source_stride == rhs.source_stride && lhs.dest_stride == rhs.dest_stride && lhs.count == rhs.count;
+		}
+
+		friend bool operator!=(const stride_dimension& lhs, const stride_dimension& rhs) { return !(lhs == rhs); }
+	};
+
+	size_t offset_in_source = 0; ///< offset in the source allocation, in bytes, of the first chunk to copy.
+	size_t offset_in_dest = 0;   ///< offset in the destination allocation, in bytes, of the first chunk to copy.
+	int num_complex_strides = 0; ///< number of strides which are not contiguous in either source or destination.
+	stride_dimension strides[2]; ///< in the 3D / 2-complex case, strides[0] is the outer and strides[1] the inner stride.
+	size_t contiguous_size = 0;  ///< number of contiguous bytes in the last dimension.
+
+	friend bool operator==(const nd_copy_layout& lhs, const nd_copy_layout& rhs) {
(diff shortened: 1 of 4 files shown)


fknorr:
I've moved copy-layout generation logic to a separate header so I can re-use it in device copy tests. It doesn't use `GENERATE` to avoid creating thousands of host / device allocations in those tests.

psalz (approved):
Without having fully understood the layouting mechanism, the tests seem comprehensive enough to ensure it does the right thing. API also looks good!

include/nd_memory.h
```diff
@@ -0,0 +1,122 @@
+#pragma once
+
+#include "grid.h"
+#include "ranges.h"
+
+#include <string.h>
+
+namespace celerity::detail {
+
+/// Describes a a box-shaped copy operation between two box-shaped allocations in terms of linear offsets and strides.
```
GagaLP, line 10:
```suggestion
/// Describes a box-shaped copy operation between two box-shaped allocations in terms of linear offsets and strides.
```

include/nd_memory.h
```diff
@@ -0,0 +1,122 @@
...
+		*current_range *= copy_range[d];
+		if(source_range[d] != copy_range[d] || dest_range[d] != copy_range[d]) { contiguous = false; }
+	}
+
+	return layout;
+}
+
+/// For every contiguous chunk in an nd-copy, invoke `f(byte_offset_in_source, byte_offset_in_dest, chunk_bytes)`.
+template <typename F>
+inline void for_each_contiguous_chunk(const nd_copy_layout& layout, F&& f) {
```
GagaLP, line 73: I would rename this to `for_each_contiguous_chunk_do` or `for_each_contiguous_chunk_apply`, but this is just a preference.

include/nd_memory.h
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// From allocation `source_base` spanning `source_box` to allocation `dest_base` spanning `dest_box`, copy `copy_region` elements of `elem_size` bytes.
+inline void nd_copy_host(const void* const source_base, void* const dest_base, const box<3>& source_box, const box<3>& dest_box, const region<3>& copy_region,
+    const size_t elem_size) //
+{
+	for(const auto& copy_box : copy_region.get_boxes()) {
+		nd_copy_host(source_base, dest_base, source_box, dest_box, copy_box, elem_size);
+	}
+}
```
PeterTh, line 120: All of these interface have a user-supplied `elem_size`, rather than being templated on an element type. I guess this feels somewhat natural since they are low-level building blocks.

However, you can always make a size from a type but not the other way around, so if there were any underlying API we ever want to target with these that might benefit from type specialization we lose that option with this interface.

PeterTh (commented):
LGTM, particularly the testing, with one concern regarding the API design.

GagaLP (approved):
Nice job.

The tests seem quite extensive, but I think all of them are needed to catch all cases.

include/nd_memory.h
fknorr: In principle you're right, the most immediate effect this could have is missed optimization due to the lack of alignment information. However all current call sites are in the backend and are thus type-erased already, so this interface would need to continue existing even if we added a templated version.

include/nd_memory.h
fknorr: Prior art: `host_utils::for_each_item()`, `region_map::impl::for_each()`, and `std::for_each()`.

PR #258, opened 2024-06-25
[IDAG] Add new threading facilities

fknorr:
This PR upstreams and tests two threading facilities required for IDAG execution: The in-order `thread_queue` and the `double_buffered_queue` for fast communication between threads.

- `async_event` gains the ability to return results from async operations (in practice always pointers = results from `sycl::malloc`, which is why I opted to make the result typpe `void*` instead of `std::any`)
- `thread_queue` takes functors as submissions and returns `async_events` while providing timing/profiling capabilities for host task, host copy and allocation operations. This will replace the CTPL thread pool, which would require a lot of boilerplate to keep in while providing little benefit, because with OoO dispatch, we don't have a need for pools with > 1 thread anymore.
- `double_buffered_queue` generalizes the queueing mechanism that's currently in place between main- and scheduler thread and ensures the consumer is contended as little as possible. This is a stand-alone component now because it will also be used between scheduler and executor threads.

I do realize there's a staggering number of things called queue now.

### DIFF
--- a/include/double_buffered_queue.h
+++ b/include/double_buffered_queue.h
@@ -1,0 +1,69 @@
+#pragma once
+
+#include <atomic>
+#include <condition_variable>
+#include <mutex>
+#include <vector>
+
+namespace celerity::detail {
+
+/// Clang as of 17.0.0 does not expose std::hardware_destructive_interference_size because it has issues around -march / -mcpu flags among others
+/// (see discussion at https://discourse.llvm.org/t/rfc-c-17-hardware-constructive-destructive-interference-size/48674).
+/// To keep it simple we conservatively pick an alignment of 128 bytes to avoid false sharing across all relevant architectures.
+/// Aarch64 and PowerPC64 have 128-byte cache lines; and x86_64 prefetches 64-byte cache lines in pairs starting from Sandy Bridge
+/// (see https://github.com/crossbeam-rs/crossbeam/blob/e7b5922e/crossbeam-utils/src/cache_padded.rs for a detailed enumeration by architecture).
+constexpr size_t hardware_destructive_interference_size = 128;
+
+/// (Thread-safe) multi-producer single-consumer queue that uses double-buffering to avoid lock contention and keep dequeueing latency as low as possible.
+template <typename T>
+class double_buffered_queue {
+  public:
+	/// Push a single element to the queue. Instead of frequently pushing multiple elements, consider using a vector<T> as the element type.
+	void push(T v) {
+		{
+			std::lock_guard lock(m_write.mutex);
+			// This push might allocate, which is the reason why double_buffered_queue retains ownership of both queues in order to re-use allocated memory and
+			// keep the lock duration as short as possible.
+			m_write.queue.push_back(std::move(v));
+			// Notify the reader that it is worth taking the lock
+			m_write.queue_nonempty.store(true, std::memory_order_relaxed);
+		}
+		m_write.resume.notify_one();
+	}
+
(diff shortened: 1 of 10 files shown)


test/thread_queue_tests.cc
```diff
@@ -0,0 +1,61 @@
...
+
+	auto evt1 = tq.submit([] {});
+	auto evt2 = tq.submit([] { std::this_thread::sleep_for(std::chrono::milliseconds(99)); });
+
+	test_utils::await(evt1);
+	test_utils::await(evt2);
+
+	CHECK(evt1.get_native_execution_time().has_value());
+	CHECK(evt2.get_native_execution_time().value() >= std::chrono::milliseconds(99));
+}
```
psalz, line 61: Should we have a test that throwing inside job causes panic?

include/thread_queue.h
```diff
@@ -0,0 +1,155 @@
...
+		if(m_impl != nullptr) {
+			m_impl->queue.push(job{} /* termination */);
+			m_impl->thread.join();
+		}
+	}
+
+	/// Submit a job to the thread queue.
+	/// `fn` must take no arguments and return either `void` or a type convertible to `void *`, which will be forwarded as the result into the event.
+	template <typename Fn>
+	async_event submit(Fn&& fn) {
```
psalz, line 43:
```suggestion
	async_event submit(Fn&& fn) {
            static_assert(std::is_same_v<std::invoke_result_t<Fn>, void> || std::is_convertible_v<std::invoke_result_t<Fn>, void*>);
```
?

psalz (approved):
LGTM!

GagaLP (approved):
LGTM!

I particularly like the addition of the profiling capability.

PeterTh (commented):
Code LGTM, but I have a question regarding the PR description: why does OoO dispatch alleviate the need for a thread pool?

fknorr:
> Why does OoO dispatch alleviate the need for a thread pool?

Because OoO assignment addresses individual threads as lanes, whereas before, the thread pool decided which job would get assigned to which pool thread. The backend will simply maintain an array of `thread_queues` and explicitly index into it with the lane id.

include/double_buffered_queue.h
```diff
@@ -0,0 +1,69 @@
...
+		if(m_write.queue_nonempty.load(std::memory_order_relaxed) /* opportunistic */) {
+			std::lock_guard lock(m_write.mutex);
+			swap(m_read.queue, m_write.queue);
+			// m_read.queue was cleared before the swap, so m_write.queue is empty now
+			m_write.queue_nonempty.store(false, std::memory_order_relaxed);
+		}
+		return m_read.queue;
+	}
+
+	/// After this function returns, the result of `pop_all` is non-empty as long as there is only exists a single reader thread.
```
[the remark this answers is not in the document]
fknorr: Fixed.

test/thread_queue_tests.cc
fknorr: Propagating an exception out of `thread_main` (as panic in tests would do) would invoke `std::terminate`, so I'm afraid it's not possible at least within this setup. Alternatively we could call `std::promise::set_exception` and re-throw in the executor thread around event completion, but that would make the handler very non-local.

include/thread_queue.h
fknorr: Sure! Also added some `std::decay_t` and `mutable`s while I was at it.

test/thread_queue_tests.cc
psalz: I see, then it's not worth the hassle I'd say.

PR #260, opened 2024-06-29
[IDAG] Pre-Execution Refactoring

fknorr:
This PR performs some small refactorings necessary for me to introduce the new executor- and backend types in another standalone PR _without_ replacing the entire runtime. It is best reviewed per commit.

1. The additional members attached to device-kernel / host-task instructions for bounds checking now also include the task type, which will improve the quality of diagnostic messages.
2. Instructions and outbound pilots are flushed from the IGGen in batches instead of individually, which reduces the frequency with which an implementor of the delegate (= the executor) needs to take locks to move this info to their thread.
3. `accessor` and `closure_hydrator::accessor_info` now refer to "allocations" and "buffers" instead of "virtual" and "backing buffers", and the bounds-checking is improved from using simple arrays to a proper struct.
5. `host_task_launcher` (currently only used as a placeholder) now is independent of `host_queue`, which will disappear soon.
6. As the names `backend` and `executor` will be re-used, these namespaces / components are renamed to `legacy_backend` and `legacy_executor` in the meantime so they can stick around for the next PR.


GagaLP (approved):
Looks good.

src/instruction_graph_generator.cc
```diff
@@ -1547,7 +1547,8 @@ instruction* generator_impl::launch_task_kernel(batch& command_batch, const exec
 		assert(chunk.execution_range.get_area() > 0);
 		assert(chunk.device_id.has_value());
 		return create<device_kernel_instruction>(command_batch, *chunk.device_id, tsk.get_launcher<device_kernel_launcher>(), chunk.execution_range,
-		    std::move(allocation_map), std::move(reduction_map) CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(, tsk.get_id(), tsk.get_debug_name()),
+		    std::move(allocation_map),
+		    std::move(reduction_map) CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(, tsk.get_type(), tsk.get_id(), tsk.get_debug_name()),
```
PeterTh, line 1551: Nitpick: can we have the OOB-check-specific stuff on a separate line?

src/instruction_graph_generator.cc
```diff
@@ -1556,7 +1557,8 @@ instruction* generator_impl::launch_task_kernel(batch& command_batch, const exec
 		assert(chunk.memory_id == host_memory_id);
 		assert(reduction_map.empty());
 		return create<host_task_instruction>(command_batch, tsk.get_launcher<host_task_launcher>(), chunk.execution_range, tsk.get_global_size(),
-		    std::move(allocation_map), tsk.get_collective_group_id() CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(, tsk.get_id(), tsk.get_debug_name()),
+		    std::move(allocation_map),
+		    tsk.get_collective_group_id() CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(, tsk.get_type(), tsk.get_id(), tsk.get_debug_name()),
```
PeterTh, line 1561: Same as above.

include/accessor.h
```diff
@@ -186,12 +185,13 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 		// We currently don't support boundary checking for accessors created using accessor_testspy::make_device_accessor,
 		// which does not set m_oob_indices.
 		if(m_oob_indices != nullptr) {
-			const bool is_within_bounds_lo = all_true(index >= m_accessed_virtual_subrange.offset);
-			const bool is_within_bounds_hi = all_true(index < (m_accessed_virtual_subrange.offset + m_accessed_virtual_subrange.range));
+			const bool is_within_bounds_lo = all_true(index >= m_accessed_buffer_subrange.offset);
```
PeterTh, line 188: I'm not sure if all this renaming is worth it to touch all this code, since I don't think the new names are any more immediately obvious than the old ones. But I also don't think it's the other way around, so it's fine with me.

PeterTh (approved):
LGTM, a few comments.

include/accessor.h
fknorr: I'm trying to get rid of the "virtual buffer" vs "backing buffer" terminology because its a remnant from pre-USM Celerity, and these are more easily confused because of "buffer" appearing in both.

src/instruction_graph_generator.cc
fknorr: We can if I do an explicit `#if CELERITY_ACCESSOR_BOUNDARY_CHECK ... #endif` instead of the wrapping macro. `clang-format` does not want to play ball if I force a newline, it gets confused by the `,` being inside of the macro parentheses.

psalz (approved):
LGTM!

src/instruction_graph_generator.cc
fknorr: Also fixed.

PR #262, opened 2024-07-17
Fix graph printing in TDAG / CDAG generation tests

fknorr:
Previously with `--print-graphs`, dot graphs were written to the log inside TDAG / CDAG tests, but test logs are captured and discarded on success since #234, breaking this command-line option.

This PR updates TDAG / CDAG tests to write dot graphs to stdout like already done for IDAG tests. Test names are included in the dot title, and no additional context information is printed alongside. The user / developer is expected to invoke test executables with [render-graphs.py](https://github.com/celerity/celerity-debug-tools/blob/master/render-graphs.py).

While at it, I've taken the opportunity to rename `test_utils::print_graphs` to `g_print_graphs` to highlight that it's an unprotected global.

### DIFF
--- a/test/instruction_graph_test_utils.h
+++ b/test/instruction_graph_test_utils.h
@@ -507,7 +507,7 @@ class idag_test_context {
 	~idag_test_context() {
 		// instruction-graph-generator has no exception guarantees, so we must not call further member functions if one of them threw an exception
 		if(m_uncaught_exceptions_before == std::uncaught_exceptions()) { finish(); }
-		maybe_log_graphs();
+		maybe_print_graphs();
 	}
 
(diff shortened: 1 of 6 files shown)


test/instruction_graph_test_utils.h
```diff
@@ -694,13 +694,10 @@ class idag_test_context {
 	}
 
 	void maybe_log_graphs() {
```
PeterTh, line 696: Maybe call this `maybe_print_graphs`, I'd otherwise assume that it logs.

PeterTh (approved):
LGTM!

psalz (approved):
LGTM!

PR #263, opened 2024-07-18
[IDAG] Fix re-ordering around eager assignment in out_of_order_engine

fknorr:
Fixes a (very rarely triggered) assertion seen in [this CI run of the IDAG branch](https://github.com/celerity/celerity-runtime/actions/runs/9976972858/job/27570513266).

The bug condition manifests as follows:
1. A host-task or device kernel instruction `k1` is submitted.
2. `k1` is assigned. `lane.last_incomplete_submission` is set to `k1`.
3. A second instruction `k2` is submitted on the same device as `k1` and is marked as `conditional_eager_assignable` because its only predecessor is `k1`.
4. A third instruction `k3` is submitted, also with the single  predecessor `k1` but with a higher piority than `k2`. It is marked  as `conditional_eager_assignable` in the same fashion.
5. `k3` is eagerly assigned. `lane.last_incomplete_submission` is set to `k3`.
6. The executor queries the `async_events` for `k1` and `k3` in exactly this order, and both kernels complete in-between those two event queries, so `out_of_order_engine` is only notified about `k3`'s completion. This out-of-order completion is explicitly permitted by the engine. `lane.last_incomplete_submission` is set to `nullptr`.
7. Eager assignment is attempted on `k2`, which expects `lane.last_incomplete_submission` to be `k1` (success) or `k3` (abort), but finds `nullptr`. `pop_assignable` treats this condition as success, but `assign_one` asserts against it.

Technically, `k2` is transitively unconditional assignable because we know via `nullptr` that there is no ongoing work on our lane. However upgrading this situation to `unconditional_assignable_state` while still waiting for the call to `complete(k1)` would open a whole new can of worms for this incredibly niche situation, so we abort conditional assignment in this case.

### DIFF
--- a/test/out_of_order_engine_tests.cc
+++ b/test/out_of_order_engine_tests.cc
@@ -585,3 +585,32 @@ TEST_CASE("eager assignment is only attempted when the one incomplete dependency
 	CHECK(iq_12.assignment_for(k1).lane != iq_3.assignment_for(k3).lane);
 	CHECK(iq_12.assignment_for(k2).lane != iq_3.assignment_for(k3).lane);
 }
+
+TEST_CASE("multiple conditional-eagerly-assignable instructions with the same predecessor can be re-ordered on assignment",
+    "[out_of_order_engine][regression]") //
+{
+	out_of_order_test_context octx(1);
+	const auto k1 = octx.device_kernel({}, device_id(0));
+
+	const auto iq_k1 = octx.assign_one();
+	REQUIRE(iq_k1.has_value());
+	CHECK(iq_k1->instruction == k1);
+
+	const auto k2 = octx.device_kernel({k1}, device_id(0));      // conditional_eagerly_assignable with expected_last_submission == k1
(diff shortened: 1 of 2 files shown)


test/out_of_order_engine_tests.cc
```diff
@@ -585,3 +585,30 @@ TEST_CASE("eager assignment is only attempted when the one incomplete dependency
 	CHECK(iq_12.assignment_for(k1).lane != iq_3.assignment_for(k3).lane);
 	CHECK(iq_12.assignment_for(k2).lane != iq_3.assignment_for(k3).lane);
 }
+
+TEST_CASE("multiple conditional-eagerly-assignable instructions with the same predecessor be re-ordered on assignment", "[out_of_order_engine][regression]") {
```
PeterTh, line 589: "**can** be reordered on assignment"?

PeterTh (approved):
LGTM

PR #264, opened 2024-07-18
Fix: Do not value-capture buffers into host tasks in system_benchmarks

fknorr:
It is illegal to value-capture buffers etc into host task lambdas, and actually breaks with the new runtime. Two instances of this pattern, both in `system_benchmarks.cc`, were missed in #244.

### DIFF
--- a/test/system_benchmarks.cc
+++ b/test/system_benchmarks.cc
@@ -23,12 +23,14 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
 #endif
 
 	celerity::distr_queue queue;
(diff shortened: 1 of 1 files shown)


psalz (approved):
So the current failure mode is that it just hangs? Will this be improved with the new executor?

fknorr:
Currently (buffer-manager runtime) this _just works_ because lifetime extension makes sure that the final reference is dropped on TDAG pruning. The IDAG runtime detects the attempt to destroy a buffer from the wrong thread and aborts, but for some reason [the most recent benchmark run ](https://github.com/celerity/celerity-runtime/actions/runs/9995587277/job/27628077350) timed out instead of aborting. Not sure why that was, but it _worked on my machine_.

PR #265, opened 2024-07-24
[IDAG] Switch to Instruction-Graph Scheduling

fknorr:
This is the final PR in the IDAG series. It switches to the new IDAG-based runtime and drops all newly unused legacy components.

new runtime architecture

- `runtime` now manages multiple devices, and the `distr_queue` API has been updated to reflect the fact.
- `buffer_manager`, `reduction_manager` and `host_object_manager` are now gone. ID assignment for these types is now handled by the runtime directly, and all components interacting with buffers, reductions and host objects (graph generators, executor and recorders) track the relevant state themselves as instructed by `notify_*_created`/`_destroyed` introduced in #246. As a result, tasks do not need to keep strong references to buffers and host objects around anymore (`lifetime_extending_state`).
- `scheduler` now generates both the command- and the instruction graph in the same thread, and maintains ownership of both structures. The CDAG is pruned at generation time (since commands never leave the scheduler thread), and the IDAG is pruned once the scheduler is notified of epoch completion. Command serialization is gone, and with it, the command `is_flushed` marker.
- The `runtime` now uses `live_executor` (replacing `legacy_executor` and `worker_job`) together with a `communicator` and `backend` instance to execute instructions. `communicator` together with `receive_arbiter` replace `buffer_transfer_manager`. `backend` implementations replace `legacy_backend`, `host_queue` and `device_queue`.
- Runtime destruction is now delayed until the last buffer / queue / host object is destroyed. With this change, we can stop distinguishing between a non-existing and a shut-down runtime. For backwards compatibility, `~distr_queue` will continue to epoch-synchronize. `runtime` asserts that non-thread-safe functions are only called from the application thread, which will trigger onaccidental value-captures of buffers / host objects into host tasks.
- Reductions are now available on all SYCL implementations since hipSYCL has added support. This allows us to drop a lot of `#ifdefs` in tests and frontend code.
- `log_context`, which was only used by `worker_job`, is removed.
- `vendor/ctpl`, which was only used by `host_queue`, is removed.

Since one node now addresses multiple GPUs, scheduling becomes more expensive (IDAG generation is maybe ~4x as expensive as CDAG generation). This will be visible in benchmark results.


include/distr_queue.h
```diff
@@ -116,16 +112,36 @@ class distr_queue {
...
   private:
-	std::shared_ptr<detail::distr_queue_tracker> m_tracker;
-
-	void init(detail::device_or_selector device_or_selector) {
-		if(!detail::runtime::is_initialized()) { detail::runtime::init(nullptr, nullptr, device_or_selector); }
-		try {
-			detail::runtime::get_instance().startup();
-		} catch(detail::runtime_already_started_error&) {
-			throw std::runtime_error("Only one celerity::distr_queue can be created per process (but it can be copied!)");
+	struct tracker {
```
PeterTh, line 115: I think we'd want a brief comment here on what the purpose of the `distr_queue::tracker` struct is. It's not user-facing, but for internal documentation.

include/fence.h
```diff
@@ -70,42 +70,34 @@ namespace celerity::detail {
 template <typename T>
 class host_object_fence_promise final : public detail::fence_promise {
   public:
-	explicit host_object_fence_promise(const experimental::host_object<T>& obj) : m_host_object(obj) {}
+	explicit host_object_fence_promise(const T* instance) : m_instance(instance) {}
```
PeterTh, line 73: What's the background for this change?

include/log.h
```diff
@@ -25,40 +19,4 @@ namespace celerity::detail {
 
 using log_level = spdlog::level::level_enum;
 
-struct log_context {
```
PeterTh, line 28: What happened to the log context? Was it unused?

include/task_manager.h
```diff
@@ -77,11 +76,6 @@ namespace detail {
 
 			auto unique_tsk = into_task(std::move(cgh));
 
-			// Require the collective group before inserting the task into the ring buffer, otherwise the executor will try to schedule the collective host
-			// task on a collective-group thread that does not yet exist.
-			// The queue pointer will be null in non-runtime tests.
-			if(m_queue) m_queue->require_collective_group(unique_tsk->get_collective_group_id());
```
PeterTh, line 83: I guess this issue doesn't happen in instruction graph execution?

src/runtime.cc
```diff
@@ -105,233 +105,323 @@ namespace detail {
...
+
+		m_user_bench = std::make_unique<experimental::bench::detail::user_benchmarker>(*m_cfg, m_local_nid);
 
 		cgf_diagnostics::make_available();
 
-		m_h_queue = std::make_unique<host_queue>();
-		m_d_queue = std::make_unique<device_queue>();
+		auto devices = std::visit(
+		    [&](const auto& value) { return pick_devices(m_cfg->get_host_config(), value, sycl::platform::get_platforms()); }, user_devices_or_selector);
+		assert(!devices.empty());
```
PeterTh, line 156: I think this shouldn't be an assertion but an error check with reporting to the user, since it could be a somewhat common setup error.

src/runtime.cc
```diff
@@ -105,233 +105,323 @@ namespace detail {
...
+		require_call_from_application_thread();
+		return *m_task_mngr;
+	}
 
-	std::string runtime::gather_command_graph() const {
-		assert(m_command_recorder.get() != nullptr);
-		const auto graph_str = print_command_graph(m_local_nid, *m_command_recorder);
+	std::string gather_command_graph(const std::string& graph_str, const size_t num_nodes, const node_id local_nid) {
+		const auto comm = MPI_COMM_WORLD;
+		const int tag = 'CDAG'; // Celerity does not perform any other peer-to-peer communication over MPI_COMM_WORLD
```
PeterTh, line 283: This is cute, but I locally get a `warning: multi-character character constant [-Wmultichar]` for it.

PeterTh (commented):
At last, a negative LoC balance.

Also, that's a beautiful picture, and it (and its source) should probably go in docs.

include/fence.h
fknorr: The fence promise is owned by the TDAG, so copying the host object here would lifetime-extend the instance, which we generally don't do anymore.

include/log.h
fknorr: It's only use was in `worker_job`, which is now gone. `executor` has full control of submissions an does all the relevant logging.

include/task_manager.h
fknorr: The IDAG RT does not tie collective host tasks to particular threads, so this is not relevant anymore.

But I also think the comment was somewhat misleading, since `host_queue` could always just have spawned new CG threads as-needed. I believe the original reason was that doing this from the TM enforces that all `MPI_Comm_dup`s happen in a single global total order. This is now guaranteed through IDAG dependencies on the `clone_collective_group_instruction` we emit for the first use of each new CG.

src/runtime.cc
fknorr: This exception is already thrown in in `pick_devices`, so the assertion merely checks a postcondition.

src/runtime.cc
fknorr: Outsmarted again. I've increased the level of l33t.

psalz (approved):
R.I.P. BM, BTM, et al. LGTM!

src/config.cc
```diff
@@ -82,28 +72,9 @@ size_t parse_validate_dry_run_nodes(const std::string_view str) {
...
-		    "Process has local rank {}, but CELERITY_DEVICES only includes {} device(s)", host_cfg.local_rank, devices.empty() ? 0 : devices.size() - 1)};
-	}
-	if(static_cast<long>(devices.size()) - 1 > static_cast<long>(host_cfg.node_count)) {
-		throw env::validation_error{fmt::format(
-		    "CELERITY_DEVICES contains {} device indices, but only {} worker processes were spawned on this host", devices.size() - 1, host_cfg.node_count)};
-	}
-
-	return devices;
+	throw env::validation_error{
+	    "Support for CELERITY_DEVICES has been removed with Celerity 0.6.0. Please use SYCL or vendor specific means to limit device visibility."};
```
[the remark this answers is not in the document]
fknorr: Good catch!

src/scheduler.cc
```diff
@@ -1,103 +1,118 @@
...
 #include "distributed_graph_generator.h"
-#include "frame.h"
-#include "graph_serializer.h"
-#include "legacy_executor.h"
+#include "instruction_graph_generator.h"
+#include "log.h"
 #include "named_threads.h"
+#include "recorders.h"
 
-#include <matchbox.hh>
```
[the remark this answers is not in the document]
fknorr: It shouldn't have been.

test/test_utils.cc
```diff
@@ -233,11 +233,13 @@ namespace celerity::test_utils_detail {
 
 // These error and warning messages will appear depending on the system the (runtime) tests are executed on, so we must not fail tests because of them.
 
-const char* const expected_runtime_init_warnings_regex = "Celerity has detected that only .* logical cores are available to this process.*";
+// NOCOMMIT why are these separate definitions? Can we inline them into runtime_fixture()?
```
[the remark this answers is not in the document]
fknorr: Guess so, doesn't hurt to have all of them in one place anyway.

src/legacy_executor.cc
```diff
@@ -1,176 +0,0 @@
...
-		    m_metrics.device_idle.get().count(), m_metrics.starvation.get().count());
-	}
-
-	void legacy_executor::run() {
-		closure_hydrator::make_available();
-		bool done = false;
-
-		while(!done || !m_jobs.empty()) {
-			// Bail if a device error ocurred.
-			if(m_running_device_compute_jobs > 0) { m_d_queue.get_sycl_queue().throw_asynchronous(); }
```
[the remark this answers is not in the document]
fknorr: I'm now doing a periodic check every 256 executor loop iterations.

PR #266, opened 2024-07-29
Fix out-of-bounds access in double_buffered_queue_tests

fknorr:
Found by chance in an [ASan run](https://github.com/celerity/celerity-runtime/actions/runs/10140967533/job/28037265641): Calling `vector::back()` requires `!empty()`.


GagaLP (approved):
Great catch. Interesting bug. LGTM!

PR #267, opened 2024-08-02
Integrate with Tracy profiler

fknorr:
This integrates the runtime (particularly scheduler and executor) with [Tracy](https://github.com/wolfpld/tracy) version 0.11 for profiling support. This will allow us to optimize the runtime, and users to find what kernels and other operations their application is spending the most time in.

screenshot

To compile Celerity with Tracy support, pass `-DCELERITY_TRACY_SUPPORT=ON` (default `OFF`) to CMake. To enable tracing at runtime, populate the environment variable `CELERITY_TRACY=off|fast|full` (default `off`). `fast` emits the same zones as `full`, but avoids attaching debug information that takes measurable time to stringify inside Celerity. Leaving the runtime setting to `off` should have no visible performance impact.

Inside Celerity, tracing is made available through the `CELERITY_DETAIL_TRACY_*` macros from tracy.h, which expand to Tracy code when support is enabled, or to an empty token list when disabled. Inside `live_executor.cc`, additional context is maintained to track the asynchronous completion of instructions as Fibers. Also, the arguments formerly passed to the `CELERITY_TRACE` macro can be captured and copied into zone text to avoid duplicating the debug stringification.

The PR performs four additional, minor changes to the runtime:
- Move `host_config` (an MPI dependency) out of `config` in order to parse environment and enable / disable Tracy before the call to `MPI_Init`.
- Auto-enable kernel profiling when `CELERITY_TRACY=full`
- Add `estimated_global_memory_traffic_bytes` attribute to `device_kernel_instruction`. It estimates the SM <-> DRAM traffic inside a kernel from accessor data types and range mappers, which allows us to print a throughput estimate on device kernel zones, which is relevant for weak-scaling problems.
- Add `backend::init`, which triggers any potential backend device initialization on executor thread startup, which would otherwise delay (and distort the execution time of) the first instructions executed per device.

In CI, Tracy support is enabled in (and only in) the SimSYCL release build, so we do get notified about compile errors but not much else. Let me know if you can come up with a better automation setup.

Please experiment with this integration a bit and feel free to push fixups for zone texts and the somewhat arbitrary zone "color scheme" I have come up with.


fknorr:
I've bumped Tracy to a post-release version and since wolfpld/tracy#854 was merged. This allows us to explicitly specify the order of threads within the profiler view.

We might want to add documentation on how to run a Celerity application against Tracy.

CMakeLists.txt
```diff
@@ -123,6 +123,8 @@ else()
 endif()
 option(CELERITY_USE_MIMALLOC "Use the mimalloc memory allocator" ${CELERITY_USE_MIMALLOC_DEFAULT})
 
+option(CELERITY_TRACY_SUPPORT "Enable Tracy profiler integration" OFF)
```
PeterTh, line 126: We should have some documentation for the compile time options (like we have for env vars). Probably in `docs/installation.md`

CMakeLists.txt
```diff
@@ -176,6 +178,14 @@ if(CELERITY_USE_MIMALLOC)
   fetch_content_from_submodule(mimalloc vendor/mimalloc)
 endif()
 
+if(CELERITY_TRACY_SUPPORT)
+  set(TRACY_FIBERS ON CACHE BOOL "" FORCE)
+  set(TRACY_STATIC ON CACHE BOOL "" FORCE)
+  set(TRACY_NO_SAMPLING ON CACHE BOOL "" FORCE)
+  set(TRACY_NO_SYSTEM_TRACING ON CACHE BOOL "" FORCE)
```
PeterTh, line 185: Do we want to FORCE these? The user might want them?

include/print_utils.h
```diff
@@ -193,6 +213,21 @@ struct as_sub_second {
...
 };
 
+/// Wrap a byte count in this to auto-format it as KiB / MiB / etc.
+struct as_binary_size {
+	as_binary_size(double bytes) : bytes(bytes) {}
+	as_binary_size(size_t bytes) : bytes(static_cast<double>(bytes)) {}
+	double bytes;
+};
+
+/// Wrap a byte-per-second ratio in this to auto-format it as KiB/s, MiB/s, ...
```
PeterTh, line 223: Bikeshedding: do we want power-of-2 units for transfer speeds? HW capabilities are usually specified with metric prefixes.

src/live_executor.cc
```diff
@@ -8,20 +8,262 @@
...
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 #include <matchbox.hh>
 
+
 namespace celerity::detail::live_executor_detail {
 
+#if CELERITY_TRACY_SUPPORT
```
PeterTh, line 26: Could this be moved to a separate file for readability purposes (i.e. to, as much as possible, keep the core functionality of the live executor separate from concerns related to Celerity integration).

include/tracy.h
```diff
@@ -0,0 +1,105 @@
...
+#define CELERITY_DETAIL_TRACY_ZONE_NAME(...)                                                                                                                   \
+	CELERITY_DETAIL_IF_TRACY_ENABLED_FULL(::celerity::detail::tracy_detail::apply_string([&](const auto& n) { ZoneName(n.data(), n.size()); }, __VA_ARGS__))
+#define CELERITY_DETAIL_TRACY_ZONE_TEXT(...)                                                                                                                   \
+	CELERITY_DETAIL_IF_TRACY_ENABLED_FULL(::celerity::detail::tracy_detail::apply_string([&](const auto& t) { ZoneText(t.data(), t.size()); }, __VA_ARGS__))
+
+#define CELERITY_DETAIL_TRACY_ZONE_SCOPED_V(TAG, COLOR_NAME, ...)                                                                                              \
+	CELERITY_DETAIL_TRACY_ZONE_SCOPED(TAG, COLOR_NAME);                                                                                                        \
+	CELERITY_DETAIL_TRACY_ZONE_NAME(__VA_ARGS__);
+
+#define CELERITY_DETAIL_TRACY_SET_THREAD_NAME_AND_ORDER(NAME, ORDER) CELERITY_DETAIL_IF_TRACY_ENABLED(::tracy::SetThreadNameWithHint(NAME, ORDER))
```
PeterTh, line 105: Is this redundant with `celerity::detail::set_thread_name`? Should it just use this information, or should on macro do both (and the Tracy part only if there is Tracy)?

src/runtime.cc
```diff
@@ -147,16 +187,26 @@ namespace detail {
 		}
 #endif
 
+		if(!s_test_mode && m_cfg->get_tracy_mode() != tracy_mode::off) {
+			if constexpr(CELERITY_TRACY_SUPPORT) {
+				CELERITY_WARN("Profiling with Tracy is enabled. Performance may be negatively impacted.");
+			} else {
+				CELERITY_WARN("CELERITY_TRACY is set, but Celerity was compiled without Tracy support. Ignoring.");
+			}
+		}
```
PeterTh, line 196: Arguably, this should be in the environment variable validator.

test/utils_tests.cc
```diff
@@ -49,3 +49,13 @@ TEST_CASE("name strings are correctly extracted from types", "[utils][get_simpli
...
 }
+
+TEST_CASE("replace_all replaces all occurrences of a pattern in string", "[utils][replace_all]") {
+	CHECK(utils::replace_all("hello world", " ", "_") == "hello_world");
+	CHECK(utils::replace_all("hello world", "l", "") == "heo word");
+	CHECK(utils::replace_all("hello world", " ", "<-->") == "hello<-->world");
+	CHECK(utils::replace_all("hello world", "l", "_") == "he__o wor_d");
+	CHECK(utils::replace_all("hello world", "o", "<->") == "hell<-> w<->rld");
+	CHECK(utils::replace_all("hello world", "foo", "bar") == "hello world");
+	CHECK(utils::replace_all("hello world", "", "bar") == "hello world");
```
PeterTh, line 60: For symmetry, maybe also test the empty input string case.

PeterTh (commented):
Awesome stuff.

Some inline notes, also, this PR should update [`CHANGELOG.md`](https://github.com/celerity/celerity-runtime/blob/master/CHANGELOG.md)!

CMakeLists.txt
fknorr: done.

CMakeLists.txt
fknorr: AFAIK `FORCE` is needed to override the default that Tracy passes to these `option()`s.

test/utils_tests.cc
fknorr: done.

src/runtime.cc
fknorr: I disagree. Warning 1) should not be in `config.cc` because potential slowness is not a concern of the config parser. I had thought about moving 2) in there before, but for consistency I should then make `config::get_tracy_mode()` etc. conditionally available, which adds a lot of preprocessor lines for little gain.

src/live_executor.cc
fknorr: I had considered this in the beginning, but since I ended up with quite a lot of Tracy-conditional code within executor member functions that reference these structures directly anyway, I decided against it.

include/tracy.h
fknorr: `detail::set_thread_name` sets the POSIX thread name, whereas `TracySetThreadName` just updates the Tracy structures. Tracy in principle will detect the POSIX thread name, but I've observed this to be somewhat unreliable / racy.

Not sure what is to be gained from merging these since on the API side, `order` is also Tracy-specific.

include/print_utils.h
fknorr: Done.

README.md
```diff
@@ -122,6 +122,14 @@ Celerity's runtime behavior:
   at the end of execution (requires log level `info` or higher).
 - `CELERITY_DRY_RUN_NODES` takes a number and simulates a run with that many nodes
   without actually executing the commands.
+<<<<<<< HEAD
```
PeterTh, line 125: I think you might have committed a merge conflict?

include/tracy.h
PeterTh: My idea regarding merging is that we want every thread that has a POSIX thread name to also have a Tracy thread name, and those names to be the same.

README.md
fknorr: Oops!

psalz (approved):
Very cool, ship it!

include/tracy.h
PeterTh: I made a note to look into this later.

PR #270, opened 2024-08-05
Enforce instant submission in AdaptiveCpp

fknorr:
Previously we would call `dag::flush_async()` after every submit, but AdaptiveCpp supports instant submission with forward progress guarantees for some time, and since AdaptiveCpp/AdaptiveCpp#1542 we have `-DACPP_FORCE_INSTANT_SUBMISSION` to assert from Celerity that this actually happens.

This PR bumps AdaptiveCpp to the latest release, removes queue flushing and adds `-DACPP_FORCE_INSTANT_SUBMISSION` .


PeterTh (approved):
LGTM.

psalz (approved):
Apart from the error in the integration tests (and SimSYCL?), LGTM.

Since this is a proper release, should we set it as the minimum required version in our CMakeLists?

fknorr:
Well, the `FORCE` check definitely works....

fknorr:
So, turns out making this work with OpenMP is impossible at the moment - see the corresponding [ACpp issue](https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1555). This means we have to keep the flush-hack around for the time being. We might decide to conditionally enable the `FORCE` define in CI (outside integration tests) as an automated sanity check though.

PR #272, opened 2024-08-06
Clean up SYCL legacy

fknorr:
This PR

- Replaces all uses of `<CL/sycl>` and `cl::sycl::` with `<sycl/sycl>` and `sycl::` as per 2020 (DPC++ produces deprecation warnings for this)
- Removes the few remaining (unconditional) workarounds we had introduced for ComputeCpp
- Commits to the hipSYCL -> AdaptiveCpp name change and bumps to v24.06 for good measure.


docs/platform-support.md
```diff
@@ -8,7 +8,7 @@ The most recent version of Celerity aims to support the following environments.
 
 ## SYCL Implementations
 
-* hipSYCL ≥ revision [`db000681`](https://github.com/illuhad/hipSYCL/commit/db000681), with
+* AdaptiveCpp ≥ revision [`v24.06.0`](https://github.com/AdaptiveCpp/AdaptiveCpp/tree/v24.06.0), with
```
psalz, line 11: Should we add a note that we currently don't support the single pass compilation flow?

psalz (approved):
Feels good - FGTM!

docs/platform-support.md
fknorr: Done!

PR #283, opened 2024-09-09
Shutdown runtime `atexit()`, introduce `celerity::queue` , deprecate `distr_queue`

fknorr:
As it turns out, moving the runtime shutdown trigger from the current refcount approach to using `atexit` is relatively straight-forward. This change will allow a user to create multiple queues, and to create new Celerity objects after the last active one has gone out of scope.

To add back some manual control, `celerity::runtime::shutdown` is added.

We had an existing bug where `~buffer()` would not necessarily wait for all tasks that could read from the host-init-pointer, leading to potential use-after-free behavior. I fixed this by submitting an epoch when destroying a buffer that references user memory, and applied the same fix for `host_object<T&>` and `host_object<void>`, both of which semantically reference user memory.

The PR further adds `celerity::queue` as a replacement for `distr_queue` and deprecates the latter. `queue` is more closely aligned to SYCL queue, and has two significant differences to `distr_queue`:
- There are no device-selecting constructors, because their existence would communicate that there can be multiple queues addressing disjoint subsets of devices, which the runtime does not support yet.
- `wait()`, the replacement for `slow_full_sync()`, does not trigger an `MPI_Barrier`. This is always enough to guarantee local synchronization, and we should find a better mechanism for user-side benchmarking anyway.

### Open questions
- [ ] Can we downgrade the epoch emitted for the `~buffer` fix to a fence instruction (without the preceding "snapshot" instruction)? This could avoid a full sync.

### DIFF
--- a/include/buffer.h
+++ b/include/buffer.h
@@ -107,6 +107,7 @@ class buffer {
 			if(host_init_ptr != nullptr) {
 				const auto user_ptr = const_cast<void*>(host_init_ptr); // promise: instruction_graph_generator will never issue a write to this allocation
 				user_aid = detail::runtime::get_instance().create_user_allocation(user_ptr);
+				references_user_memory = true;
 			}
 			id = detail::runtime::get_instance().create_buffer(detail::range_cast<3>(range), sizeof(DataT), alignof(DataT), user_aid);
 		}
@@ -119,11 +120,15 @@ class buffer {
 		~tracker() {
 			CELERITY_DETAIL_TRACY_ZONE_SCOPED("buffer::~buffer", DarkCyan);
 			detail::runtime::get_instance().destroy_buffer(id);
+			// The user must guarantee liveness of the user pointer only until the buffer instance goes out of scope
+			// TODO This is more synchronization than necessary - consider issuing a fence-like task that does not block concurrent tasks.
+			if(references_user_memory) { detail::runtime::get_instance().sync(detail::epoch_action::none); }
 		}
 
 		detail::buffer_id id;
 		celerity::range<Dims> range;
 		std::string debug_name;
+		bool references_user_memory = false;
 	};
 
 	std::shared_ptr<tracker> m_tracker;
--- a/include/fence.h
+++ b/include/fence.h
@@ -4,7 +4,6 @@
 #include <memory>
 #include <type_traits>
 
-#include "distr_queue.h"
 #include "host_object.h"
 #include "runtime.h"
 #include "task_manager.h"
@@ -20,7 +19,7 @@ class buffer_fence_promise;
 namespace celerity {
 
 /**
- * Owned representation of buffer contents as captured by celerity::distr_queue::fence.
+ * Owned representation of buffer contents as captured by celerity::queue::fence.
  */
 template <typename T, int Dims>
 class buffer_snapshot {
@@ -103,14 +102,10 @@ class buffer_fence_promise final : public detail::fence_promise {
 	std::promise<buffer_snapshot<DataT, Dims>> m_promise;
 };
 
-} // namespace celerity::detail
-
-namespace celerity {
-
 template <typename T>
-std::future<T> distr_queue::fence(const experimental::host_object<T>& obj) {
+std::future<T> fence(const experimental::host_object<T>& obj) {
 	static_assert(std::is_object_v<T>, "host_object<T&> and host_object<void> are not allowed as parameters to fence()");
-	CELERITY_DETAIL_TRACY_ZONE_SCOPED("distr_queue::fence", Green2);
+	CELERITY_DETAIL_TRACY_ZONE_SCOPED("queue::fence", Green2);
 
 	detail::side_effect_map side_effects;
 	side_effects.add_side_effect(detail::get_host_object_id(obj), experimental::side_effect_order::sequential);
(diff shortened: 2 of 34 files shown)


fknorr:
We should benchmark this. I'm still not convinced that `slow_full_sync` is slow enough today to deserve the name. With distributed graph generation, it does not add network latency besides the `MPI_Barrier` (which was on the order of 10µs for 64 nodes on Marconi 100).  With the barrier gone, it simply waits for execution to catch up locally, which should be universally understood from the name `wait`.

User-side benchmarking: I agree. Maybe the following API would work for the meantime?
```c++
queue.wait(celerity::experimental::barrier);
```

test/system/distr_tests.cc
```diff
@@ -122,7 +122,7 @@ namespace detail {
 	    "[reductions][print_graph][smoke-test]") //
 	{
 		env::scoped_test_environment test_env(print_graphs_env_setting);
-		// init runtime early so the distr_queue ctor doesn't override the log level set by log_capture
+		// init runtime early so the queue ctor doesn't override the log level set by log_capture
```
psalz, line 125: This comment is outdated (we still need to initialize it early though, for the `runtime_testspy` calls below)

include/fence.h
```diff
@@ -137,16 +132,33 @@ std::future<buffer_snapshot<DataT, Dims>> distr_queue::fence(const buffer<DataT,
...
 
 template <typename T, int Dims>
 using buffer_snapshot [[deprecated("buffer_snapshot is no longer experimental, use celerity::buffer_snapshot")]] = celerity::buffer_snapshot<T, Dims>;
 
-template <typename... Params>
-[[deprecated("fence is no longer experimental, use celerity::distr_queue::fence")]] [[nodiscard]] auto fence(celerity::distr_queue& q, const Params&... args) {
-	return q.fence(args...);
+template <typename T, int Dims>
+[[deprecated("fence is no longer experimental, use celerity::queue::fence")]] [[nodiscard]] auto fence(
+    celerity::distr_queue& /* q */, const buffer<T, Dims>& buf, const subrange<Dims>& sr) {
```
psalz, line 148: Any particular reason for spelling the overloads out explicitly? (Not that I'm against it, just curious)

include/queue.h
```diff
@@ -0,0 +1,96 @@
...
+	/// Waits for all tasks submitted to the queue to complete.
+	///
+	/// Since waiting will stall the scheduling of more work, this should be used sparingly - more so than on a single-node SYCL program.
+	void wait() { // NOLINT(readability-convert-member-functions-to-static)
+		CELERITY_DETAIL_TRACY_ZONE_SCOPED("queue::wait", Red2);
+		[[maybe_unused]] const auto tid = detail::runtime::get_instance().sync(detail::epoch_action::none);
+		CELERITY_DETAIL_TRACY_ZONE_NAME("T{} wait", tid);
+	}
+
+	/**
```
psalz, line 34: Inconsistent docblock style (above uses `///`)

psalz (approved):
I agree that we still need an option for user benchmarking; other than that LGTM!

include/fence.h
fknorr: Reason is backwards compatibility: The explicit parameter types allow the second argument of `experimental::fence(buf, {{0, 0}, {10, 10}})` to be deduced as a `subrange<2>`, but that would not be possible on `fence(Params &&...args)`.

test/system/distr_tests.cc
fknorr: Good catch, fixed!

include/queue.h
fknorr: Fixed.

fknorr:
I have added `queue::wait(experimental::barrier)` in case someone really wants to barrier-synchronize like `slow_full_sync` did. I believe this not to be too useful for time measurements though.

psalz (approved):

include/buffer.h
```diff
@@ -119,11 +120,14 @@ class buffer {
 		~tracker() {
 			CELERITY_DETAIL_TRACY_ZONE_SCOPED("buffer::~buffer", DarkCyan);
 			detail::runtime::get_instance().destroy_buffer(id);
+			// The user must guarantee liveness of the user pointer only until the buffer instance goes out of scope
+			if(references_user_memory) { detail::runtime::get_instance().sync(detail::epoch_action::none); }
```
PeterTh, line 124: This is more synchronization than strictly necessary. It might be possible to do this more efficiently.

Also, as discussed offline, it would be neat to have an integration test for this.

src/runtime.cc
```diff
@@ -247,7 +250,7 @@ namespace detail {
 
 	void runtime::require_call_from_application_thread() const {
 		if(std::this_thread::get_id() != m_application_thread) {
-			utils::panic("Celerity runtime, distr_queue, handler, buffer and host_object types must only be constructed, used, and destroyed from the "
+			utils::panic("Celerity runtime, queue, handler, buffer and host_object types must only be constructed, used, and destroyed from the "
```
PeterTh, line 253: This should probably be "From one application thread" (the application might be multithreaded)

PeterTh:
Regarding `slow_full_sync` vs. `wait` -- we decided offline to go with the latter for similarity to SYCL, but to add a basic performance bug detection feature which informs users if they perform a suspiciously large number of waits compared to the overall number of tasks.

src/runtime.cc
fknorr: We have multiple occurrences of "the application thread" in docs because Celerity is not user-thread-safe. Maybe "user thread" is an alternative? Although I'm not sure that's more helpful.

include/buffer.h
fknorr: I've added an integration test.

The fine-grained synchronization needs a refactor of fences first, so I'm postponing this.

fknorr:
I've added a warning heuristic + test to the task manager.

PeterTh (approved):

PR #288, opened 2024-10-02
Disambiguate `parallel_for(Integral, ...)`

fknorr:
With this PR, `handler` accepts the following shorthands as required (in parts) by SYCL:
- `parallel_for(Integral, ...)` => `parallel_for(range<1>, ...)`
- `parallel_for(Integral, Integral, ...)` => `parallel_for(range<1>, id<1>, ...)`

I've used the opportunity of having to write new tests for this to move all `handler`-related tests that do not require runtime instantiation to `handler_tests`.


psalz (approved):
LGTM

PeterTh (commented):
No problems with the signature change, and I also like moving the handler tests to a dedicated file, but I'd prefer if the hint-specific tests remained in their own hint tests file (I think that makes them easier to find, and generally having a finer granularity there is better).

fknorr:
I've split the tests back up.

Also, due to [what's likely a compiler bug](https://godbolt.org/z/1zWq3Gj5P) the changes messed up the DPC++ build, which I've fixed by constraining `nd_range` conversion operators. To make overload resolution more traceable, I've added a type trait that matches the `ReductionsAndKernel` argument lists.

PeterTh (approved):

PR #290, opened 2024-10-09
Generate "fat" push commands

psalz:
This changes command generation to create a single "fat" push command for each buffer and task, instead of a separate push for each last writer and target node, similar to how await pushes already work.

This will enable future optimizations in the IDAG, and also reduces the number of commands that need to be generated, particularly in scenarios where buffer contents are highly fragmented.

Even though the `command_graph_generator` has been designed to work for arbitrary numbers of chunks per node, since oversubscription was moved to the IDAG, there was no way of creating multiple chunks and this logic wasn't actually being tested. I've introduced a new `test_chunk_multiplier` that can be set to create multiple chunks per node for this purpose.

Note: This builds on top of #289, only the second to last commit is relevant to the review.


src/command_graph_generator.cc
```diff
@@ -199,379 +183,289 @@ void command_graph_generator::generate_distributed_commands(const task& tsk) {
...
+		for(auto& [nid, region] : scratch.target_regions) {
+			combined_region = region_union(combined_region, region);
+			target_regions.push_back({nid, std::move(region)});
 		}
 
-		// In case this buffer was in a pending reduction state but the result was discarded, remove the pending reduction.
-		if(buffer.pending_reduction.has_value()) {
-			m_completed_reductions.push_back(buffer.pending_reduction->rid);
-			buffer.pending_reduction = std::nullopt;
+		auto* cmd = create_command<push_command>(transfer_id(tsk.get_id(), bid, no_reduction_id), std::move(target_regions),
```
fknorr, line 407:
```suggestion
		auto* const cmd = create_command<push_command>(transfer_id(tsk.get_id(), bid, no_reduction_id), std::move(target_regions),
```

src/instruction_graph_generator.cc
```diff
@@ -1749,57 +1749,57 @@ void generator_impl::compile_execution_command(batch& command_batch, const execu
...
+			for(auto& [original_writer, boxes] : individual_send_boxes) {
+				concurrent_send_regions.push_back(region(std::move(boxes)));
+			}
 		}
-	}
 
-	allocate_contiguously(command_batch, trid.bid, host_memory_id, std::move(required_host_allocation));
-	establish_coherence_between_buffer_memories(command_batch, trid.bid, host_memory_id, concurrent_send_regions);
+		allocate_contiguously(command_batch, trid.bid, host_memory_id, std::move(required_host_allocation));
+		establish_coherence_between_buffer_memories(command_batch, trid.bid, host_memory_id, concurrent_send_regions);
```
fknorr, line 1785: These must be outside the loop, otherwise we get reallocations / staging copies for every target node instead of once

include/command_graph_generator.h
```diff
@@ -105,15 +106,108 @@ class command_graph_generator {
 
 	command_graph& get_command_graph() { return m_cdag; }
 
+	/// Only for testing: Instead of (at most) a single chunk per node, generate `multiplier` chunks per node for each task.
+	/// This is to ensure that CDAG generation logic works for arbitrary numbers of chunks, even though we don't provide
+	/// a way for users to specify more than one chunk per node... yet.
+	void set_test_chunk_multiplier(const size_t multiplier) {
```
fknorr, line 112:
```suggestion
	void test_set_chunk_multiplier(const size_t multiplier) {
```
So that "intrusive" test functions have the `test_` prefix like they do everywhere else.

fknorr (changes requested):
Another step closer to RDMA :rocket:

Some remarks below - please also add an IDAG test that verifies that no unnecessary resizes / staging copies happen in case of a broadcast push (same region is sent to multiple nodes), both with d2d copies enabled and disabled.

include/command_graph_generator.h
psalz: If by "everywhere" you mean runtime.h - ok, done!

psalz:
> please also add an IDAG test that verifies that no unnecessary resizes / staging copies happen in case of a broadcast push (same region is sent to multiple nodes) and for a push with partially overlapping regions.

Thanks for the review! So far, the introduction of fat pushes hasn't changed anything from the perspective of the IDAG - the number of allocations / coherence copies is the same as for individual pushes. I'm wondering whether it makes sense to invest time into optimizing this, if everything is going to be replaced with host-side staging (and ultimately device side staging + RDMA) in the near future anyway?

fknorr:
> I'm wondering whether it makes sense to invest time into optimizing this, if everything is going to be replaced with host-side staging (and ultimately device side staging + RDMA) in the near future anyway?

True, all that will have to be re-worked anyway. I'm fine with a TODO comment in `iggen.cc`!

src/print_graph.cc
```diff
@@ -173,48 +113,134 @@ std::string print_command_graph(const node_id local_nid, const command_recorder&
 		return fmt::format("id_{}_{}", local_nid, id);
 	};
 
-	const auto print_vertex = [&](const command_record& cmd) {
-		static const char* const colors[] = {"black", "crimson", "dodgerblue4", "goldenrod", "maroon4", "springgreen2", "tan1", "chartreuse2"};
+	const auto get_subgraph = [&](const task_command_record& task_cmd) {
+		if(task_subgraph_dot.count(task_cmd.tid) == 0) {
```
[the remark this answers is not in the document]
fknorr: The future is now

src/instruction_graph_generator.cc
psalz: Out of scope for this PR, as discussed below.

PeterTh (approved):

fknorr (approved):

PR #292, opened 2024-10-11
Generalize range mappers to regions, add `neighborhood(along_axes)` RM

fknorr:
3/5/7-point stencils with 2D splits are currently inefficient because they produce boundary exchanges with diagonal neighbor nodes / devices, since range mapper outputs must be (bounding) boxes, meaning that the `neighborhood` range mapper will include the unnecessary diagonal boundary elements.

This PR generalizes the range mapper interface to return regions (which are implicitly convertible from subranges) and adds the `direct_neighborhood` range mapper, which returns the "tight" read regions for 3/5/7-point stencil codes. `direct_neighborhood` is the only stable public API to expose complex-region range mappers and `region` remains in the `detail` namespace, so there is no documented API to write similar user-defined mappers yet.

I've used this opportunity to move pure range mapper tests to `range_mapper_tests.cc`.

## Open Questions

1. Semantics of `direct_neighborhood(1, 1)` are obvious, but what is `direct_neighborhood(2, 2)` supposed to be? Currently it excludes all diagonal elements (you get an "ever taller plus sign" when increasing the boundary size parameter), but there might be codes who want to use the [von Neumann neighborhood](https://en.wikipedia.org/wiki/Von_Neumann_neighborhood) (i.e. Manhattan distance) instead. For the communication issue at hand this doesn't matter too much since as soon as you need boundary exchanges with a diagonal neighbor, you're not losing much by just sticking with the overestimated `neighborhood` mapper.

2. Is `direct_neighborhood` the right name? Or should the new funcitonality be integrated into the existing `neighborhood`, which might receives a `enum metric` ctor param instead?

3. Should `direct_neighborhood` be `experimental`?

4. The OOB check still operates on one bounding box per accessor. This will cause OOB accesses to remain unreported if they fall inside the bounding box (= the allocation), but outside the mapped range. Is it worth tracking this more precisely?


test/instruction_graph_memory_tests.cc
```diff
@@ -80,6 +80,37 @@ TEMPLATE_TEST_CASE_SIG("multiple overlapping accessors trigger allocation of the
...
+    "[instruction_graph_generator][instruction-graph][memory]", ((int Dims), Dims), 1, 2, 3) //
+{
+	const auto full_range = test_utils::truncate_range<Dims>({256, 256, 256});
+	const auto accessed_region =
+	    region<Dims>({test_utils::truncate_box<Dims>({{0, 0, 0}, {64, 64, 64}}), test_utils::truncate_box<Dims>({{128, 128, 128}, {192, 192, 192}})});
+	const auto access_bounding_box = bounding_box(accessed_region);
+
+	test_utils::idag_test_context ictx(1 /* nodes */, 0 /* my nid */, 1 /* devices */);
+	auto buf = ictx.create_buffer(full_range);
+	ictx.device_compute(range(1)).discard_write(buf, [=](const auto& /* chnk*/) { return accessed_region; }).submit();
```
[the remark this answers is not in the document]
fknorr: I really hope it would be illegal to RVO out of a mutable lambda capture...

include/range_mapper.h
```diff
@@ -54,50 +55,49 @@ namespace detail {
...
-		if(BufferDims > 1 && end[1] > buffer_size[1]) { sr.range[1] = sr.offset[1] <= buffer_size[1] ? buffer_size[1] - sr.offset[1] : 0; }
-		if(BufferDims > 2 && end[2] > buffer_size[2]) { sr.range[2] = sr.offset[2] <= buffer_size[2] ? buffer_size[2] - sr.offset[2] : 0; }
-		return sr;
+	region<BufferDims> clamp_region_to_buffer_size(const region<BufferDims>& r, const range<BufferDims>& buffer_size) {
+		return region_intersection(r, box<BufferDims>::full_range(buffer_size));
 	}
 
 	template <int BufferDims, typename Functor>
-	subrange<BufferDims> invoke_range_mapper(int kernel_dims, Functor fn, const celerity::chunk<3>& chunk, const range<BufferDims>& buffer_size) {
+	region<BufferDims> invoke_range_mapper(int kernel_dims, Functor fn, const celerity::chunk<3>& chunk, const range<BufferDims>& buffer_size) {
```
psalz, line 63: While hacking around on range mappers recently I discovered that this function is actually not being used anywhere!

psalz (changes requested):
Good change! As discussed in person, I would prefer having an option to control the shape of the neighborhood on the existing mapper, instead of introducing another similar-named one. Maybe it could be as simple as `range_mapper(const size_t dim0, ..., bool include_diagonals = true)`.

PeterTh (approved):
Regarding naming, I generally think having more names makes for an easier to use interface than having more options on existing names. As such, I actually don't mind `direct_neighborhood` at all.

As discussed offline, one of my concerns with this is that people would see it and then implement e.g. a periodic boundary condition range mapper in a similar way, inadvertently ballooning their local allocations.

Other than that, I like it.
I also really like moving range mapper tests to their own TU!

psalz:
> Regarding naming, I generally think having more names makes for an easier to use interface than having more options on existing names. As such, I actually don't mind `direct_neighborhood` at all.

If the names are descriptive, yes. But what does "direct" mean? To me it implies "right next to". By that definition, the first element along the diagonal is also a "direct" neighbor, with an euclidean distance of 1. You could say it's Manhattan distance, but then what about a `direct_neighborhood{2, 2}` - how are the outer elements "direct" neighbors of the current element by any definition?

include/range_mapper.h
fknorr: Good catch. Removed!

fknorr:
I like the idea of merging `neighborhood` and `direct_neighborhood` because I think additional ctor arguments are more discoverable when the user is looking on a variation of the default neighborhood.

Live, we've also been discussing a generic `metric` parameter where we could eventually offer an L1 or L2 metric neighborhood. However I realized this is ill-defined with non-square neighborhood sizes: An `{1, 2}` neighborhood makes sense with maximum metric and the "axis aligned direct metric" discussed here, but not with L1 / L2.

Given this I propose scrapping generic metrics and having ctors of the following style:
```c++
enum classs diagonals { no, yes };
neighborhood::neighborhood(id<Dims> extent, diagonals diag = diagnonals::yes);
```
or
```c++
neighborhood::neighborhood(id<Dims> extent);
neighborhood::neighborhood(id<Dims> extent, no_diagonals_tag /* no_diagonals */);
```
both with the existing `(dim0, ... dimN)` forwarding to diagonals = yes.

PeterTh:
Of the two options, I prefer the enum class, but (at the risk of bikeshedding) I am not sure about the "diagonals" name. It's rather clear when we are talking about a "1"-sized neighborhood, but for "2" and above it becomes a bit more vague.

It might make more sense to define what's included rather than what is excluded

Options:
```c++
// advantage: kind of intuitive for the 2D case
// disadvantage: not really well-specified; it's actually a "minus" for 1D :P
enum class included_elements { full_range, plus_only };
// advantage: accurate
// disadvantage: slightly wordy, perhaps less intuitive?
enum class included_elements { full_range, primary_axes_only };

neighborhood::neighborhood(id<Dims> extent, included_elements elems = included_elements::full_range);
```

The first enum entry could also be `complete_hyperrectangle` if we really wanted to be super precise...

Also, should this be an error in 1D or just a no-op?
(If it was a compile-time param we could make it only compile for 2D and 3D)

psalz:
More bikeshedding:

```c++
enum class neighborhood_shape {
        along_axes,
        bounding_box
};
```

PeterTh:
I like that.

fknorr:
Implemented as suggested.

include/range_mapper.h
```diff
@@ -189,27 +175,63 @@ namespace access {
...
+	///
+	/// This is typically used in stencil applications. For bounding-box neighborhoods,
+	/// - `neighborhood(1)` declares the read for a 1D 3-point stencil,
+	/// - `neighborhood(1, 1)` for a 2D 9-point stencil, and
+	/// - `neighborhood(1, 1, 1)` for a 3D 27-point stencil.
+	///
+	/// For neighborhoods defined along axes only,
+	/// - `neighborhood(range(1), along_axes)` declares the read for a 1D 3-point stencil,
+	/// - `neighborhood(range(1, 1), along_axes)` for a 2D 5-point stencil, and
+	/// - `neighborhood(range(1, 1, 1), along_axes)` for a 3D 7-point stencil.
```
psalz, line 195: Hm, what's the reasoning for adding a new `range` overload instead of adding the default argument to the existing constructors?

test/instruction_graph_memory_tests.cc
```diff
@@ -1059,3 +1090,45 @@ TEST_CASE("device kernels report global memory traffic estimate based on range m
 		CHECK(reduce->estimated_global_memory_traffic_bytes == buffer_size / num_devices * sizeof(float));
 	}
 }
+
+TEST_CASE("5-point stencil program with 2D split does not exchange boundaries with diagonal neighbor devices",
```
psalz, line 1094: I guess we should also have a CDAG test:

```c++
TEST_CASE("5-point stencil program with 2D split does not exchange boundaries with diagonal neighbor nodes", "[command_graph_generator][command-graph]") {
	cdag_test_context cctx(4 /* num nodes */);

	constexpr range<2> buffer_size(256, 256);
	auto buf_a = cctx.create_buffer<2>(buffer_size);

	cctx.device_compute(range(buffer_size)).name("init").hint(experimental::hints::split_2d()).discard_write(buf_a, acc::one_to_one()).submit();
	cctx.device_compute(range(buffer_size))
	    .name("read")
	    .hint(experimental::hints::split_2d())
	    .read(buf_a, acc::neighborhood(range(1, 1), acc::along_axes))
	    .submit();

	const std::vector<std::pair<node_id, node_id>> diagonals{{0, 3}, {1, 2}, {2, 1}, {3, 0}};
	for(auto [nid, diag] : diagonals) {
		CHECK(std::ranges::none_of(cctx.query<push_command_record>().on(nid).assert_count(1)->target_regions, [&](auto& tr) { return tr.first == diag; }));
	}
}
```

PeterTh:
I'm not sure I like the enum instead of enum class in terms of dev experience for new developers. It's a bit shorter, but it means that whenever any user now writes `celerity::access::` then their code completion will offer the individual enum entries as recommendations, at the same level as any of the prebuilt range mappers. IMHO the words `bounding_box` and `along_axis` on their own don't give much indication that they are only relevant in the narrow context of specializing the neighborhood (for a new user at least).

I.e. when I as a Celerity newb am writing `accessor r(result, cgh, access::` and then my IDE tells me `along_axes` is an option, and that sounds like I what I want, then I'm tempted to pick that. Of course that will be a type error eventually, but still.

PeterTh:
I have no objection to `enum class celerity::neighborhood_shape`.

test/instruction_graph_memory_tests.cc
fknorr: Added this.

include/range_mapper.h
fknorr: The existing ones are now deprecated to notify our users that they should look at the optional shape parameter for their code, which might have used the bounding-box neighborhood before without intending it.

PeterTh (approved):

psalz (approved):

fknorr:
So unfortunately (but not entirely unsurprising), using regions instead of boxes in the requirement analysis hot path has a measurable performance impact on scheduling performance, especially for TDAG generation which does little else.

I've managed to improve this somewhat by optimizing `region_intersection` for the common case of box-shaped regions, but didn't manage to eliminate it entirely.

fknorr:
I've added the `region_builder` type for the common pattern of accumulating boxes / regions into their union in a loop. Unlike repeated calls to `region_union` or explicit uses of `box_vector`, this eliminates any unnecessary region normalization steps. Unfortunately this doesn't show up as significant in the benchmarks, but I've decided to keep it anyway because it transports the semantics better than the `box_vector` approach we had in many places before.

I've also drive-by fixed a bug introduced in #289, where the cggen would detect uninitialized reads only at a bounding-box granularity instead of per-element.

PR #293, opened 2024-10-14
Only poll events of instructions that are actively executing

fknorr:
We currently poll events of all assigned instructions, even if they are only eagerly assigned an not currently executing yet. Polling SYCL (CUDA) events has measurable latency. This PR optimizes polling by exposing the "completion front" on the out-of-order-engine, which exposes the set of instructions without incomplete predecessors.


src/out_of_order_engine.cc
```diff
@@ -94,6 +94,9 @@ struct engine_impl {
 	/// is assumed to have completed earlier (triggering its removal from the map).
 	std::unordered_map<instruction_id, incomplete_instruction_state> incomplete_instructions;
 
+	/// The subset of incomplete_instructions that to not have incomplete predecessors themselves.
+	std::unordered_set<instruction_id> completion_front;
```
PeterTh, line 98: Minor point, but I think it would make more sense to name this something like `incomplete_front`, or maybe `pending_instructions`. (I would assume that a "completion_front" is a set of complete instructions)

PeterTh (approved):
LGTM

src/out_of_order_engine.cc
psalz: Since the `out_of_order_engine` only really deals with incomplete instructions, I think it's quite clear from context. That being said, how about 'execution_front`?

psalz (approved):
LGTM!

src/out_of_order_engine.cc
fknorr: I'm ok with `execution_front`, but `incomplete_front` and `pending_instructions` are too unspecific IMO.

fknorr: I've renamed it to `execution_front` and also added some explicit tests on that.

PR #294, opened 2024-10-17
Experimental support for AdaptiveCpp SSCP flow

fknorr:
Single-pass compilation cannot provide `__SYCL_DEVICE_ONLY__` detection macro, so we need to guard accessor hydration with `__acpp_if_target_host()`. This works because AdaptiveCpp does not verify that kernels are actually device copyable.

For DPC++ (and anything else) we still need to conditionally remove definitions for accessor copy and move ctors.

This now works on my machine :tm:, but there is no CI integration yet.

### DIFF
--- a/include/accessor.h
+++ b/include/accessor.h
@@ -154,7 +154,12 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	accessor(accessor&&) noexcept = default;
 	accessor& operator=(accessor&&) noexcept = default;
 
-#if !defined(__SYCL_DEVICE_ONLY__)
+	// On the host, we provide a custom copy constructor that performs accessor hydration before a kernel launch. This code path must not exist on the device
+	// (accesses host global state / thread locals). With DPC++ and ACpp multi-pass compilation we can distinguish these cases at preprocessing time
+	// with CELERITY_DETAIL_COMPILE_TIME_TARGET_DEVICE_ONLY. With the ACpp single-pass compiler this information does not exist before JIT time (runtime
+	// from the C++ perspective), so CELERITY_DETAIL_COMPILE_TIME_TARGET_DEVICE_ONLY is always false and we rely on a branch in copy_and_hydrate(). This works
+	// because fortunately for us ACpp doesn't verify that kernels / accessors are device copyable (= trivially copyable in our case).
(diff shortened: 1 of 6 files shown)


include/workaround.h
```diff
@@ -51,14 +51,33 @@
...
-#else
-#define CELERITY_DETAIL_ASSERT_ON_HOST(...)
-#endif
-
 #if CELERITY_ACCESSOR_BOUNDARY_CHECK
 #define CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(...) __VA_ARGS__
 #else
 #define CELERITY_DETAIL_IF_ACCESSOR_BOUNDARY_CHECK(...)
 #endif
+
```
PeterTh, line 59: I think a short comment about the bigger picture of the how and why of this PR would be nice, and probably the best place to put it would be here before the block introducing the associated preprocessor machinery?

fknorr: Added some doc comments.

PeterTh (approved):
LGTM

include/workaround.h
```diff
@@ -51,14 +51,38 @@
...
+// SYCL implementations (compiler frontends) provide different means of distinguishing host and device code. Multi-pass compliers define __SYCL_DEVICE_ONLY__ in
```
psalz, line 60:
```suggestion
// SYCL implementations (compiler frontends) provide different means of distinguishing host and device code. Multi-pass compilers define __SYCL_DEVICE_ONLY__ in
```

include/accessor.h
```diff
@@ -154,7 +154,7 @@ class accessor<DataT, Dims, Mode, target::device> : public detail::accessor_base
 	accessor(accessor&&) noexcept = default;
 	accessor& operator=(accessor&&) noexcept = default;
 
-#if !defined(__SYCL_DEVICE_ONLY__)
+#if !CELERITY_DETAIL_COMPILE_TIME_TARGET_DEVICE_ONLY // host- or universal single pass
```
psalz, line 157: This only works because ACPP does not actually check `is_device_copyable` (which implies `trivially_copyable`) - consider adding a comment!

psalz (approved):
LGTM!

include/workaround.h
fknorr: Very compliant!

PR #296, opened 2024-10-28
Bump authoritative clang-format version to 18

psalz:
`clang-format-18` is the default version shipping with Ubuntu 24.04 LTS.

This also changes the `clang-tidy` step to run on the `intel` partition: The lint container is now built on top of SimSYCL, and our `intel` partition currently has the better CPU. Both of these measures should help to reduce the time required for the clang-tidy checks.

Since we are already re-formatting a bunch of files, this is a good time to review whether we want to make some changes to our clang-format config, particularly to incorporate new features / options that have been introduced since clang-format-14.

[List of clang-format options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

Changelogs:
- [clang-format-15](https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#clang-format)
- [clang-format-16](https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#clang-format)
- [clang-format-17](https://releases.llvm.org/17.0.1/tools/clang/docs/ReleaseNotes.html#clang-format)
- [clang-format-18](https://releases.llvm.org/18.1.0/tools/clang/docs/ReleaseNotes.html#clang-format)


include/ranges.h
```diff
@@ -117,13 +117,13 @@ class coordinate {
 #undef CELERITY_DETAIL_DEFINE_COORDINATE_BINARY_COPY_LHS_OPERATOR
 
 #define CELERITY_DETAIL_DEFINE_COORDINATE_BINARY_INPLACE_LHS_OPERATOR(op)                                                                                      \
-	friend constexpr Interface& operator op(Interface& lhs, const Interface& rhs) {                                                                            \
+	friend constexpr Interface& operator op(Interface & lhs, const Interface & rhs) {                                                                          \
```
PeterTh, line 120: Any idea what's happening here?  
`Interface &` is weird in the first place, but it's very weird to have both that and `Interface&` in the same line?

Seems to happen only in the Macro definition. We could of course just ignore this.

include/utils.h
```diff
@@ -207,7 +207,8 @@ template <std::integral Integral>
 }
 
 template <typename Void>
-requires(std::is_void_v<Void>) [[nodiscard]] constexpr Void* offset(Void* const ptr, const size_t offset_bytes) {
+    requires(std::is_void_v<Void>)
```
PeterTh, line 210: I guess this is one of the things we could make a decision on, between the C++20 switch and the new clang format version.

I am happy with this default, FWIW.

test/backend_tests.cc
```diff
@@ -416,8 +416,8 @@ TEST_CASE("backends report execution time iff profiling is enabled", "[backend]"
 
 	SECTION("on host tasks") {
 		event = backend->enqueue_host_task(
-		    /* lane */ 0, [&](const box<3>&, const communicator*) { std::this_thread::sleep_for(std::chrono::milliseconds(1)); }, {}, box_cast<3>(box<0>()),
-		    nullptr);
+		    /* lane */
+		    0, [&](const box<3>&, const communicator*) { std::this_thread::sleep_for(std::chrono::milliseconds(1)); }, {}, box_cast<3>(box<0>()), nullptr);
```
PeterTh, line 420: This is very silly. I couldn't find a way to configure this behaviour with a quick search.

PeterTh (approved):

fknorr:
I would love to see
```
NamespaceIndentation: Never
```
This way we could collapse all `namespace a { namespace b {` to `namespace a::b` without having gradual formatting changes over time in unrelated PRs like we do now.

Also I think we can safely set
```
Standard: Cpp20
```

include/ranges.h
fknorr: I believe that clang is unable to parse this properly because the context is not clear from inside the macro definition.

include/utils.h
fknorr: +1 new default, these are easy to miss if not on a separate line.

test/backend_tests.cc
PeterTh: One idea: would putting the comment *behind* the 0 make it behave less silly? If so, we could standardize on doing that.

PeterTh:
Agreed on the standard, but I like `NamespaceIndentation: Inner`.

We could use this opportunity to collapse all instances of `namespace a { namespace b {`.

fknorr (approved):

psalz:
I replaced our clang-tidy CI action with https://github.com/platisd/clang-tidy-pr-comments, since the old one apparently couldn't properly handle nested `.clang-tidy` configuration files. The new action allows us to run clang-tidy with `-j48`, which should make it a whole lot faster as well!

PR #297, opened 2024-10-28
Refactor command_graph

fknorr:
**Note:** Based on #295.

The upcoming lookahead feature requires queueing commands independently of IDAG generation, so command instances must remain live for a longer period of time - currently they are pruned immediately when a new horizon / epoch is generated. I've used this opportunity for some simplification / refactoring.

- `command_graph` is now an `epoch_partitioned_graph`, and information only relevant for building the CDAG (like the execution front) is kept in the CDAG generator instead.
- `class abstract_command` is renamed to `class command`, constructors are public and `get_type` member is gone, whose only remaining use was the Tracy integration.
- CDAG pruning is moved out to the scheduler, identical to IDAG pruning.
- The CDAG generator maintains command pointers instead of looking up commands by id. The `batch` is kept as a local variable instead of a class member.


include/command_graph_generator.h
```diff
@@ -23,48 +22,52 @@ constexpr size_t max_num_nodes = 256;
...
 class command_graph_generator {
 	friend struct command_graph_generator_testspy;
 
-	inline static const write_command_state no_command = write_command_state(static_cast<command_id>(-1));
+	inline static const write_command_state no_command = {};
 
 	struct buffer_state {
-		buffer_state(region_map<write_command_state> lw, region_map<std::bitset<max_num_nodes>> rr)
-		    : local_last_writer(std::move(lw)), replicated_regions(std::move(rr)), pending_reduction(std::nullopt) {}
+		explicit buffer_state(const range<3>& range, const write_command_state initial_wcs, const node_bitset& replicated_on_nodes)
```
PeterTh, line 69: Is there a specific reason to make this 3-parameter constructor `explicit`?

include/command_graph_generator.h
```diff
@@ -218,24 +252,20 @@ class command_graph_generator {
 	 * Generates command(s) that need to be processed by every node in the system,
 	 * because they may require data transfers.
 	 */
-	void generate_distributed_commands(const task& tsk);
+	void generate_distributed_commands(batch& current_batch, const task& tsk);
```
PeterTh, line 255: Above this point, the name of the `batch` parameter is `batch`, starting from here and below it's `current_batch`. Is there actually a semantic difference or is this arbitrary? (If the latter, I suggest settling on one, preferably `current_batch`)

src/scheduler.cc
```diff
@@ -13,13 +13,25 @@
...
+		    cmd,                                                    //
+		    [](const epoch_command&) { return "epoch"; },           //
+		    [](const horizon_command&) { return "horizon"; },       //
+		    [](const execution_command&) { return "execution"; },   //
+		    [](const push_command&) { return "push"; },             //
+		    [](const await_push_command&) { return "await push"; }, //
+		    [](const reduction_command&) { return "reduction"; },   //
+		    [](const fence_command&) { return "fence"; });
+	}
+
```
PeterTh, line 27: This being in `scheduler.cc` seems a bit off.
Obviously it's currently only used here, but I'd expect to find it in print utils or so.

fknorr: Moved to `print_utils.h`.

include/command_graph_generator.h
fknorr: Renamed to `current_batch` everywhere.

include/command_graph_generator.h
fknorr: I just think explicit as the default is good practice.

psalz (approved):
LGTM!

PeterTh (approved):

PR #299, opened 2024-11-04
Move completed-epoch handling out of task_manager

fknorr:
Replaces `epoch_monitor` with an atomic counter in `runtime` for the purpose of TDAG pruning, and re-uses the fence promise mechanism for suspending the main thread when waiting on an epoch task.

This consolidates the "previous horizon is applied as an epoch" logic, which was previously duplicated, in a single location in `runtime`.

Furthermore, `task_graph` ownership is moved out of `task_manager`, making it a single-threaded graph generator like the others. A rename to `task_graph_generator` should follow eventually but was omitted here to keep diff noise low.


src/runtime.cc
```diff
@@ -311,15 +327,14 @@ namespace detail {
...
 		m_exec.reset();
 
+		// task_manager references the scheduler as its delegate, so we destroy it first.
+		m_task_mngr.reset();
+
 		// ~executor() joins its thread after notifying the scheduler that the shutdown epoch has been reached, which means that this notification is
 		// sequenced-before the destructor return, and `runtime` can now stop functioning as an executor_delegate (which would require m_schdlr to be live).
-		// m_schdlr references the task instances managed m_task_mngr, so we destroy it first. m_task_mngr uses m_schdlr as a delegate, but does not call to the
-		// delegate from its destructor.
+		// m_schdlr references the task instances managed m_task_mngr, so we destroy it first.
```
psalz, line 335: This looks outdated, the task instances are managed by `task_graph`, and this is not being destroyed first?

include/task_manager.h
```diff
@@ -164,10 +109,10 @@ namespace detail {
 
 		task_manager::delegate* m_delegate;
 
-		const size_t m_num_collective_nodes;
+		size_t m_num_collective_nodes;
 		policy_set m_policy;
 
-		task_graph m_tdag;
+		task_graph* m_tdag;
```
psalz, line 115:
```suggestion
		task_graph& m_tdag;
```

Should this be a reference? We don't need to copy `task_manager` and this can never be null.

test/dag_benchmarks.cc
```diff
@@ -74,9 +75,9 @@ TEST_CASE("benchmark task handling", "[benchmark][group:task-graph]") {
 			// create simplest possible host task
 			const auto highest_tid = tm.submit_command_group([](handler& cgh) { cgh.host_task(on_master_node, [] {}); });
 			// start notifying once we've built some tasks
-			if(i % report_interval == 0 && i / report_interval > 1) {
+			if(i % report_interval == 0 && i / report_interval > 2) {
 				// every other generated task is always a horizon (step size 0)
-				tm.notify_horizon_reached(highest_tid + 1);
+				tdag.erase_before_epoch(highest_tid - 1);
```
psalz, line 80: I don't quite follow this change

psalz (approved):
I like it, LGTM!

src/runtime.cc
fknorr: You're right!

test/dag_benchmarks.cc
fknorr: Well, this test has a horizon step size of 1, so every other task is a horizon. Reaching horizon $t_{n+1}$ means epoch $t_{n-1}$ has been applied. Previously `task_manager` did duplicate the "the last horizon becomes an epoch" logic itself, but this function is now gone and we need to erase from the TDAG manually.

There is more than one test that expects behavior like this, and it's probably right to ask if they're still testing anything meaningful that I'm not just writing out manually in the tests after the refactoring.

PeterTh (approved):

include/task_manager.h
fknorr: The lint goes away when I delete all copy ctors and assignment operators, so I did that with a comment to eventually PIMPL this class.

PR #300, opened 2024-11-04
Refactor `buffer_access_map`

psalz:
This changes the `buffer_access_map` to work in terms of consumed and produced regions instead of access modes, with the latter now only required for recording.

Additional changes:
- The access map now knows its parent tasks' geometry, so global size and dimensionality no longer need to be passed into its various member functions.
- Task level requirements are computed once upon construction.
- Execution ranges are now passed as `box<3>` instead of `subrange<3>`.
- Task/command graph printing now lists buffer accesses in declaration order, and accesses with same buffer and mode are no longer merged.

Note: This includes #292 to see whether we can compensate for the increased cost of region range mappers.


include/cgf_diagnostics.h
```diff
@@ -90,11 +91,11 @@ class cgf_diagnostics {
 		m_is_checking = false;
 		m_expected_target = std::nullopt;
 
-		for(size_t i = 0; i < m_expected_buffer_accesses->get_num_accesses(); ++i) {
+		for(size_t i = 0; i < m_expected_buffer_accesses->size(); ++i) {
 			if(!m_registered_buffer_accesses[i]) {
 				throw std::runtime_error(fmt::format("Accessor {} for buffer {} is not being copied into the kernel. This indicates a bug. Make sure "
 				                                     "the accessor is captured by value and not by reference, or remove it entirely.",
-				    i, m_expected_buffer_accesses->get_nth_access(i).first));
+				    i, m_expected_buffer_accesses->operator[](i).bid));
```
fknorr, line 98: You could just use `buffer_accesses` here to avoid the odd `->operator[]()` call.

include/task.h
```diff
@@ -15,51 +15,72 @@
...
+	};
+
 	class buffer_access_map {
 	  public:
-		void add_access(buffer_id bid, std::unique_ptr<range_mapper_base>&& rm) { m_accesses.emplace_back(bid, std::move(rm)); }
+		struct buffer_access {
+			buffer_id bid;
+			access_mode mode;
+			std::unique_ptr<range_mapper_base> range_mapper;
+		};
```
fknorr, line 37: `buffer_access` is often used outside the BAM, so it might make code simpler to move this struct to the enclosing scope.

Also consider initializing all members.

test/range_mapper_tests.cc
```diff
@@ -128,102 +128,85 @@ TEST_CASE("slice built-in range mapper behaves as expected", "[range-mapper]") {
...
+		REQUIRE(sr.range == range<1>{30});
+	}
+	{
+		range_mapper rm{acc::neighborhood<2>({10, 10}), range<2>{128, 128}};
+		auto sr = rm_result_to_subrange(rm.map_2(chunk<2>{{5, 100}, {10, 20}, {128, 128}}));
+		REQUIRE(sr.offset == id<2>{0, 90});
+		REQUIRE(sr.range == range<2>{25, 38});
+	}
+	{
+		range_mapper rm{acc::neighborhood<3>({3, 4, 5}), range<3>{128, 128, 128}};
```
fknorr, line 164: You seem to have reverted the changes from #292 here?

fknorr (commented):
I like this as an intermediate step, although I believe we should integrate the BAM into the task structure eventually. This would deduplicate the task geometry and allow us to report reduction outputs as a produced buffer region like we do with regular accesses.

include/cgf_diagnostics.h
psalz: Right!

include/task.h
psalz: I wouldn't have to spell it out everywhere if `emplace_back` would work... 😑 Done!

test/range_mapper_tests.cc
psalz: Oops, since the tests moved to a new file I missed that while copying over the changes.

fknorr (approved):

include/task.h
```diff
@@ -12,54 +13,77 @@
...
+		/// Computes the union of all produced regions (across multiple accesses) for a given execution range.
+		region<3> compute_produced_region(const buffer_id bid, const box<3>& execution_range) const;
+
+		/// Returns a set of bounding boxes, one for each accessed region, that must be allocated contiguously.
+		box_vector<3> compute_required_contiguous_boxes(const buffer_id bid, const box<3>& execution_range) const;
 
 	  private:
-		std::vector<std::pair<buffer_id, std::unique_ptr<range_mapper_base>>> m_accesses;
+		std::vector<buffer_access> m_accesses;
+		std::unordered_set<buffer_id> m_accessed_buffers; ///< Cached set of buffer ids found in m_accesses
```
PeterTh, line 83: Since this is only ever inserted into and has a handful of elements used for iteration, a small flat set (based on a deduped vector) implementation would probably be better. But we don't have such a vocabulary type, and I this probably doesn't matter enough for it to be worth it.

That said, we have so many tiny sets and maps all over the place that at some point we might consider introducing some suitable containers.

src/command_graph_generator.cc
```diff
@@ -425,7 +418,13 @@ void command_graph_generator::generate_await_pushes(batch& current_batch, const
 }
 
 void command_graph_generator::update_local_buffer_fresh_regions(const task& tsk, const std::unordered_map<buffer_id, region<3>>& per_buffer_local_writes) {
-	auto requirements = get_buffer_requirements_for_mapped_access(tsk, subrange<3>(tsk.get_global_offset(), tsk.get_global_size()), tsk.get_global_size());
+	buffer_requirements_list requirements;
+	for(const auto bid : tsk.get_buffer_access_map().get_accessed_buffers()) {
+		buffer_requirements reqs{bid, {}, {}};
+		reqs.consumed = tsk.get_buffer_access_map().get_task_consumed_region(bid);
+		reqs.produced = tsk.get_buffer_access_map().get_task_produced_region(bid);
+		requirements.emplace_back(std::move(reqs));
```
PeterTh, line 426:
```suggestion
		const auto& tm = tsk.get_buffer_access_map();
		requirements.emplace_back(bid, tm.get_task_consumed_region(bid), tm.get_task_produced_region(bid));
```

include/task.h
```diff
@@ -12,54 +13,77 @@
...
+
+		/// Computes the union of all produced regions (across multiple accesses) for a given execution range.
+		region<3> compute_produced_region(const buffer_id bid, const box<3>& execution_range) const;
+
+		/// Returns a set of bounding boxes, one for each accessed region, that must be allocated contiguously.
+		box_vector<3> compute_required_contiguous_boxes(const buffer_id bid, const box<3>& execution_range) const;
 
 	  private:
-		std::vector<std::pair<buffer_id, std::unique_ptr<range_mapper_base>>> m_accesses;
+		std::vector<buffer_access> m_accesses;
```
PeterTh, line 82: `std::vector<buffer_access>` is used here and also in some signatures.  
Maybe we should have an alias for it so that it's also easier to evaluate potential replacements

PeterTh (approved):

src/command_graph_generator.cc
psalz: Done!

psalz: ...I had to switch to `push_back` because Clang 14 doesn't support `emplace_back` for structs w/o constructor.

PR #302, opened 2024-11-06
Use `region_map::update_box` instead of `update_region` where possible

fknorr:
As we saw in #292, using single-element regions instead of boxes is surprisingly expensive, even though the underlying storage is a `small_vector`. I've discovered a few uses of `region_map::update_region` where the argument is just an implicitly-cast single box. Replacing these calls with `region_map::update_box` actually returns a few percent of graph generation performance.

At least half the instruction-graph generation time in 4 GPU / 4x oversubscribed `wave_sim` can be attributed to region map reads and writes, so if we run into scheduler-bound problems, we should re-visit the region map again. I would imagine a batch-update functionality to be helpful, where we can defer the merging / normalization step to after all updates have been queued.

**Edit:** The last benchmarks did not show any perf difference, but this change is still _correct_ so I'm going ahead with the merge.


psalz (approved):
I did some exploratory work to see if there are any low-hanging fruit left in `region_map`, but alas everything I tried only made things slower.

PR #303, opened 2024-11-06
Introduce per-device submission threads in the SYCL backend

PeterTh:
We recently discovered that codes with fine-grained kernels (at roughly `N * ~7us` runtime or below for `N` GPUs) can be executor-thread-limited due to the fixed dispatch overhead of kernels being serviced from a single executor thread for all devices.

Luckily, extending the backend to support one dispatch worker thread per device was possible without any changes to external interfaces -- by introducing a new "wrapper" type of asynchronous event, `delayed_async_event`. The implementation could also leverage the existing high-level thread queue primitive. As such, the overall amount of code added is small.

This PR makes this new mechanism optional and configurable via an environment variable, as it could be interesting to observe its impact on various code and HW combinations. To avoid ugly interfaces, a new `sycl_backend::configuration` type is introduced.

Note that the full performance advantage of this change will materialize in combination with a subsequent thread pinning PR, but it already reduces the benchmark times of some system benchmarks substantially (2D oversubscribed stencil is cut almost in half).

A note regarding coverage: the decrease in coverage isn't really accurate -- it's because coverage is reported based on the SimSYCL build, and that is not thread safe and so doesn't execute some threaded BE paths; they are generally exercised in the unit tests for any other backend.


src/backend/sycl_backend.cc
```diff
@@ -9,6 +9,14 @@
 
 #include <fmt/ranges.h>
 
+namespace {
+void wait_synchronously_for(const celerity::detail::async_event& ev) {
```
fknorr, line 16: I don't believe you have to wait for any of the events below. They are executed in-order with other submissions to the worker thread anyway.

src/backend/sycl_backend.cc
```diff
@@ -207,16 +253,18 @@ async_event sycl_backend::enqueue_host_task(size_t host_lane, const host_task_la
...
-		auto& hydrator = closure_hydrator::get_instance();
-		hydrator.arm(target::device, std::move(accessor_infos));
-		const auto launch_hydrated = hydrator.hydrate<target::device>(sycl_cgh, launch);
-		launch_hydrated(sycl_cgh, execution_range, reduction_ptrs);
+	return enqueue_device_work(device, [=, this] {
+		auto& queue = m_impl->get_device_queue(device, lane);
+		CELERITY_DETAIL_TRACY_ZONE_SCOPED("sycl::submit", Orange2);
+		auto event = queue.submit([&](sycl::handler& sycl_cgh) {
+			auto& hydrator = closure_hydrator::get_instance();
+			hydrator.arm(target::device, std::move(accessor_infos));
```
fknorr, line 266: This move is ineffective without also moving accessor_infos into the lambda capture (and making the lambda mutable).

include/backend/sycl_backend.h
```diff
@@ -78,6 +105,9 @@ class sycl_backend : public backend {
 
 	system_info& get_system_info(); // mutable system_info is filled by sycl_cuda_backend constructor
 
+	// Enqueues a task on the worker thread corresponding to the given device, and wraps the async_event returned by the task in a delayed_async_event
+	async_event enqueue_device_work(device_id device, std::function<async_event()> work);
```
fknorr: This implementation detail should not be in the header.
Also consider dropping the `std::function` and having a templated functor instead, no need for an allocation here.

include/backend/sycl_backend.h
```diff
@@ -39,7 +58,15 @@ namespace celerity::detail {
 /// This abstract class implements all `backend` functions except copies, which not subject to platform-dependent specialization.
 class sycl_backend : public backend {
   public:
-	explicit sycl_backend(const std::vector<sycl::device>& devices, bool enable_profiling);
+	struct configuration {
+		// If `per_device_worker_threads` is true, operations on each device will be enqueued on a separate worker thread.
+		// If false, all operations will be enqueued on the executor thread.
+		bool per_device_worker_threads = true;
```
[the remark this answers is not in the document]
PeterTh: I strongly disagree. In terms of actual usability during performance exploration, I'd **much** rather have a non-compile-time option.

PeterTh:
I agreed with the suggestion above to use `std::future` instead of the manual mechanism, but this turned into an interesting sidetrack that I want to briefly document:

* Using `std::future` instead turned out to be **59.7%** slower in system benchmarks, which is obviously massive. This was very unexpected, the current working theory is that it is due to the overhead of using `wait_for(0, ...` as an implementation of `is_ready`.
* Since it seemed like this is performance critical, I tried a few other implementation options:
  *  Having a `submit_and_forget` operation on the worker, since the event is not used in this case. No significant performance change.
  * Locking with a spinlock rather than a `std::mutex`. No significant perf impact.
  * Using an `atomic_bool` flag to signal availability of the event. ~7% faster over 5 runs each with a relative stddev of 4%. Not very exciting but I'll take it since the code isn't really more difficult to understand.

include/backend/sycl_backend.h
```diff
@@ -23,6 +23,25 @@ class sycl_event final : public async_event_impl {
...
 };
 
+// asynchronous event which can be filled with an actual event from another thread later
+class delayed_async_event final : public async_event_impl {
+  public:
+	struct data_block {
+		std::optional<async_event> event = std::nullopt;
+		std::mutex mutex;
+	};
+	using shared_data_block = std::shared_ptr<data_block>;
```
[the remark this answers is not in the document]
PeterTh: See longer comment below.

include/backend/sycl_backend.h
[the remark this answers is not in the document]
fknorr: I was convinced offline that this is a good idea after all because it can be set in cleared in tests. Go ahead!

fknorr (approved):
Implemenation LGTM ✔️ 

I have one final bike shed to paint about the term "backend worker thread". Can we come up with something that better conveys the distinction between threads used to hide SYCL submission latency, vs those backend threads we already allocate for host-execution lanes (h2h copies / host tasks)? "Worker" is a bit non-descriptive in that regard. I propose "device submission threads" instead.

Two things we can look into in the future (and in another PR):

- Generalizing the hand-rolled std-future-like logic in `delayed_async_event` - this might also be useful for thread queue events (but not fences, which need to be able to suspend a thread).
- Using our new threads to speed up device initialization (currently still done in the executor thread directly). @PeterTh's preliminary experiments seem to show that this is internally serialized in a CUDA lock or similar, but we should consult Tracy on this.

PeterTh:
Renamed & squashed!

src/backend/sycl_backend.cc
```diff
@@ -124,34 +145,56 @@ struct sycl_backend::impl {
...
+	if(m_impl->config.per_device_submission_threads) {
+		for(device_id did = 0; did < m_impl->system.devices.size(); ++did) {
+			m_impl->devices[did].submission_thread.emplace(fmt::format("cy-be-submission-{}", did.value), m_impl->config.profiling);
+			// no need to wait for the event -> will happen before the first task is submitted
+			(void)m_impl->devices[did].submission_thread->submit([did] {
+				closure_hydrator::make_available();
+				// TODO: Set BE submission thread affinity
+			});
+		}
+	}
```
psalz, line 175:
```suggestion
	} else {
	    closure_hydrator::make_available();
	}
```

Could we move this call from the `live_executor` to here? If submission threads are enabled then we don't need it to make it available it in the executor. Besides, it's somewhat confusing to have it called in two places.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - `handler::parallel_for(size_t, [size_t,] ...)` now acts as a shorthand for `parallel_for(range<1>, [id<1>,] ...)` (#288)
 - Experimental support for the AdaptiveCpp generic single-pass compiler (#294)
 - Constructor overloads to the `access::neighborhood` range mapper for reads in 3/5/7-point stencil codes (#292)
+- The SYCL backend now uses per-device submission threads to dispatch commands for better performance. 
+  This new behaviour is enabled by default, and can be disabled via "CELERITY_BACKEND_DEVICE_SUBMISSION_THREADS" (#303)
```
psalz, line 21:
```suggestion
- The SYCL backend now uses per-device submission threads to dispatch commands for better performance in multi-device configurations. 
  This new behavior is enabled by default, and can be disabled via "CELERITY_BACKEND_DEVICE_SUBMISSION_THREADS" (#303)
```

Come to think of it, should we disable the thread if there's only a single device..?

psalz (commented):
Nice work!

As an aside: I saw that you renamed the PR to "submission threads" but the commit messages still use "dispatch/backend worker"!

src/backend/sycl_backend.cc
PeterTh: As discussed offline, still required in some testing / non-live-execution scenarios. Documented.

src/backend/sycl_backend.cc
```diff
@@ -226,18 +270,51 @@ async_event sycl_backend::enqueue_host_copy(size_t host_lane, const void* const
...
+
+	// Note: this mechanism is quite similar in principle to a std::future/promise,
+	//       but implementing it with that caused a 50% (!) slowdown in system-level benchmarks
+	sycl_backend_detail::delayed_async_event::shared_state async_event_state = std::make_shared<sycl_backend_detail::delayed_async_event::state>();
+	auto async_event = make_async_event<sycl_backend_detail::delayed_async_event>(async_event_state);
+
+	(void)submission_thread->submit([this, device, lane, work, async_event_state] {
+		auto event = work(m_impl->get_device_queue(device, lane));
+		async_event_state->event = std::move(event);
+		async_event_state->is_ready.store(true, std::memory_order_release);
```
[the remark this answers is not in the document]
PeterTh: Good point, adjusted as suggested.

CHANGELOG.md
PeterTh: Evaluated this on the thread pinning branch: even with just 1 device, having the BE submission thread is actually slightly faster (i.e. 600 ms vs 650 ms in the 2D stencil OS3 case), so I don't think we should complicate the defaults.

PR #304, opened 2024-11-08
Benchmark improvements

PeterTh:
This PR overhauls the "system" benchmark set to better capture effects that are highly significant in some real-world workloads.  
To that end, the following new benchmarks are added:

- 2D and oversubscribed variants of the stencil benchmark
- A new benchmark which seeks to reproduce the core performance features of the "RSIM" application

The implementation of the system benchmarks in terms of the unit testing/benchmarking framework is also changed. The benchmarks are now performed in functions which are simply invoked several times from basic test cases, rather than using a templated design. This greatly reduces the number of individual kernels that need to be compiled, and also makes it easier to vary individual aspects (e.g. running the 2D tiled stencil at lower iteration counts).

There are 2 additional, benchmark-related changes:

- Minor updates to the `ci/check-perf-impact.rb` script to make it compatible with Ruby 3.2+
- An adaptation to the taskset core list in `ci/run-benchmarks.sh`, which was needlessly restrictive, distorting the real-world impact of potential multithreading optimizations on these benchmark runs.

This change should be merged before #303, and probably also before #298 (and most everything that aims at system-level performance improvements).


test/system_benchmarks.cc
```diff
@@ -139,3 +142,117 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
 	});
 	CHECK(*queue.fence(success_buffer).get() == true);
 }
+
+constexpr auto stencil_tags = "[benchmark][group:system][stencil]";
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 1D  50 iters oversub 1", stencil_tags) { run_stencil_benchmark(50, false, 1); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 1D 500 iters oversub 1", stencil_tags) { run_stencil_benchmark(500, false, 1); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 1D  50 iters oversub 3", stencil_tags) { run_stencil_benchmark(50, false, 2); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 1D 500 iters oversub 3", stencil_tags) { run_stencil_benchmark(500, false, 2); }
```
psalz, line 150: It says `3` but does `2` (same below).

test/system_benchmarks.cc
```diff
@@ -8,41 +8,37 @@
...
-class bench_runtime_fixture : public test_utils::runtime_fixture {};
-
-TEMPLATE_TEST_CASE_METHOD_SIG(
-    bench_runtime_fixture, "benchmark independent task pattern with N tasks", "[benchmark][group:system][indep-tasks]", ((int N), N), 100, 1000, 5000) {
-	constexpr size_t num_tasks = N;
-	constexpr size_t num_repeats = 2;
-	constexpr size_t items_per_task = 256;
+// This benchmark represents a set of parallel tasks working independently on a shared buffer
 
+void run_indep_task_benchmark(size_t num_tasks) {
```
psalz, line 15:
```suggestion
void run_indep_task_benchmark(const size_t num_tasks) {
```

test/system_benchmarks.cc
```diff
@@ -68,39 +64,46 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+TEST_CASE_METHOD(fixture, "benchmark independent task pattern with 2500 tasks", indep_task_tags) { run_indep_task_benchmark(2500); }
 
-TEMPLATE_TEST_CASE_METHOD_SIG(
-    bench_runtime_fixture, "benchmark stencil pattern with N time steps", "[benchmark][group:system][stencil]", ((int N), N), 50, 1000) {
-	constexpr size_t num_iterations = N;
-	constexpr int side_length = 128; // sufficiently small to notice large-scale changes in runtime overhead
 
+// This benchmark represents a basic 2D stencil, executed with 1D and 2D splits and varying levels of oversubscription
+
+void run_stencil_benchmark(size_t num_iter, bool split2d, size_t oversub) {
```
psalz, line 77:
```suggestion
void run_stencil_benchmark(const size_t num_iter, const bool split2d, const size_t oversub) {
```

test/system_benchmarks.cc
```diff
@@ -139,3 +142,117 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 1D 500 iters oversub 3", stencil_tags) { run_stencil_benchmark(500, false, 2); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 2D  30 iters oversub 1", stencil_tags) { run_stencil_benchmark(30, true, 1); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 2D 300 iters oversub 1", stencil_tags) { run_stencil_benchmark(300, true, 1); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 2D  30 iters oversub 3", stencil_tags) { run_stencil_benchmark(30, true, 2); }
+TEST_CASE_METHOD(test_utils::runtime_fixture, "benchmark stencil: 2D 300 iters oversub 3", stencil_tags) { run_stencil_benchmark(300, true, 2); }
+
+
+// This benchmark represents the core "RSIM" compute step, notable for its growing buffer access pattern
+
+void run_rsim_benchmark(size_t n_tris, size_t num_iter) {
```
psalz, line 159:
```suggestion
void run_rsim_benchmark(const size_t n_tris, const size_t num_iter) {
```

test/system_benchmarks.cc
```diff
@@ -139,3 +142,117 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+					val /= (float)t;
+					write_rad[{t, item.get_id(0)}] = val;
+				});
+			});
+		}
+		queue.submit([&](handler& cgh) {
+			accessor read_rad{rad, cgh, access::one_to_one(), read_only};
+			accessor write_rad_result{rad_result, cgh, access::one_to_one(), write_only};
+			cgh.parallel_for(rad_size, [=](item<2> item) { write_rad_result[item] = read_rad[item]; });
+		});
```
psalz, line 231: Instead of doing this copy in each iteration, you could move the body of the `BENCHMARK` into a lambda that accepts a `bool verify` and call that once before running the actual benchmark.

psalz (changes requested):

test/system_benchmarks.cc
PeterTh: Nicely spotted, thanks.

test/system_benchmarks.cc
PeterTh: Do you think that's worth it? It's just one copy of a relatively small buffer that happens once after 50 / 500 iterations of the kernel.

PeterTh: Discussed offline, and I was convinced this makes sense. Implemented.

test/system_benchmarks.cc
```diff
@@ -139,3 +142,126 @@ TEMPLATE_TEST_CASE_METHOD_SIG(
...
+				const float expected = 1.f;
+				constexpr float epsilon = 0.01f;
+				if(std::fabs(r[item] - expected) > epsilon) {
+					fmt::print("Mismatch at {}/{}: {} !~= {} +/- {}\n", item.get_id(0), item.get_id(1), r[item], expected, epsilon);
+					succ = false;
+				}
+			});
+		});
+	});
+	CHECK(*queue.fence(success_buffer).get() == true);
```
fknorr, line 260: Could be a lot simpler if we just fence'd `rad_result` and did the comparison in the main thread.

fknorr:
I've run #298 against the new benchmarks:

```
1.00 <= benchmark stencil: 1D  50 iters oversub 1 / iterations
1.20 <= benchmark stencil: 1D  50 iters oversub 3 / iterations
1.00 <= benchmark stencil: 2D  30 iters oversub 1 / iterations
1.49 <= benchmark stencil: 2D  30 iters oversub 3 / iterations
1.01 <= benchmark stencil: 1D 500 iters oversub 1 / iterations
1.21 <= benchmark stencil: 1D 500 iters oversub 3 / iterations
1.00 <= benchmark stencil: 2D 300 iters oversub 1 / iterations
1.48 <= benchmark stencil: 2D 300 iters oversub 3 / iterations
0.72 <= benchmark rsim:   64 tris  50 iters / iterations
0.50 <= benchmark rsim: 1024 tris  50 iters / iterations
0.70 <= benchmark rsim:   64 tris 500 iters / iterations
0.18 <= benchmark rsim: 1024 tris 500 iters / iterations
```

Lookahead does what it's supposed to for the RSim pattern.

It's interesting how much the oversubscribed stencil consistently slows down however. Without looking into this further it appears to be scheduler bound? I don't actually need to care about the fine-split resulting from oversubscription for figuring out the allocation bounds, so maybe there is still optimization potential.

ci/check-perf-impact.rb
```diff
@@ -46,7 +46,7 @@
 MESSAGE_FN = "#{ENV['GITHUB_WORKSPACE']}/check_perf_message.txt"
 
 # check if the expected files and env variables are present
-if !File.exists?(BENCH_FN)
+if !File.exist?(BENCH_FN)
```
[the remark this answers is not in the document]
PeterTh: Implemented support for `GITHUB_BASE_REF=remote/branch`, and the script no longer writes the reference csv anywhere.

test/system_benchmarks.cc
PeterTh: I'm now doing the verification in the lambda as per the original idea of @psalz .

PR #308, opened 2024-11-15
Do not expose Celerity internals in installed headers

fknorr:
We have a considerable number of header files now, most of which are concerned with implementation details only that should not be exposed to the user, if only in order to cut compile times somewhat. This is a refactoring PR following up on #300 and #299 that removes direct and transitive includes of all DAG related logic from `celerity.h`.

- `handler` now stores the results from CGF invocation in `struct raw_command_group` instead of creating a task instance directly. This allows it to be oblivious of the future task id and the number of collective nodes (both are task-manager parameters). This also simplifies some test code as "standalone" CGF invocation is possible for the first time.
- The global range is now passed into host task launchers instead of being captured in them, which is necessary to type-erase host task lambdas without knowledge of the number of collective nodes.
- `runtime` and `task_manager` now take a `raw_command_group` instead of a generic lambda to invoke. Task creation, previously spread throughout `handler`, now happens in `create_command_group_task` inside task.cc. This helps to separate TDAG generation from SYCL semantics more.
- command-group "parameter types" such as `struct buffer_access` are moved to the new "cgf.h" alongside `struct raw_command_group`.
- `runtime` is now an abstract base class which still exposes all static members, but is implemented by `runtime_impl` which performs all internal includes.
- `print_utils.h` is split into two files, one for public-facing types, one for internals.
- As a drive-by, `task_manager::generate_fence_task` does not take BAMs and SEMs anymore but has a overload for each variant. A `struct host_object_effect` is added as a companion to `struct buffer_access`.

All in all, this shrinks the set of headers reachable from `celerity.h` considerably. We now assert that this set does not grow unnoticed by explicitly enumerating the set of public headers which will be installed by CMake, while internal headers are only available in-tree (and thus for unit tests). Eventually we will want to move internal headers to src/ or a subdirectory, but I'm skipping this step for now to avoid merge conflicts.


include/runtime.h
```diff
@@ -44,45 +35,41 @@ namespace detail {
...
+		virtual reduction_id create_reduction(std::unique_ptr<reducer> reducer) = 0;
 
-		bool is_dry_run() const { return m_cfg->is_dry_run(); }
+		virtual bool is_dry_run() const = 0;
 
-		void set_scheduler_lookahead(experimental::lookahead lookahead);
+		virtual void set_scheduler_lookahead(experimental::lookahead lookahead) = 0;
 
-		void flush_scheduler();
+		virtual void flush_scheduler() = 0;
```
PeterTh, line 70: As discussed offline, it would be nice to find a solution here that does not require us to make these public interfaces virtual.

psalz (approved):
Neat, LGTM!

include/device.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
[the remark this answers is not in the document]
fknorr: Is now `device_selectors.h`, and `select_devices.h`.

psalz (approved):
Approving again for new testspy-pattern, looks very clean now!

src/named_threads.cc
```diff
@@ -49,7 +49,13 @@ void set_current_thread_name(const thread_type t_type);
...
 	set_current_thread_name(t_type);
 	thread_pinning::pin_this_thread(t_type);
-	CELERITY_DETAIL_TRACY_SET_THREAD_NAME_AND_ORDER(t_type);
+
+#if CELERITY_TRACY_SUPPORT
+	const char* name = tracy_detail::leak_name(named_threads::thread_type_to_string(t_type));
+	const int32_t order = tracy_detail::lane_order::thread + static_cast<int32_t>(t_type);
+	assert(order <= static_cast<int32_t>(tracy_detail::lane_order::thread_max));
+	::tracy::SetThreadNameWithHint(name, order);
+#endif
```
PeterTh, line 58: I would prefer if this remained contained in the tracy module.

PeterTh (changes requested):

PeterTh (commented):
Looks great overall!

Since this makes some (minor, but still) implementation changes to the DAG benchmarks, did you run those to make sure that there is no performance change?

Edit: one final thing -- this is mostly an internal change, but it does change the (exposed) API surface, so maybe we should have a line for it in the changelog?

PeterTh:
I ran the benchmarks on the branch, everything looks ok.

:heavy_check_mark: No significant performance change in the microbenchmark set. You are good to go!

test/test_utils.h
```diff
@@ -379,47 +317,48 @@ namespace test_utils {
...
 	detail::task_id add_nd_range_compute_task(detail::task_manager& tm, CGF cgf, celerity::nd_range<KernelDims> execution_range = {{1, 1}, {1, 1}}) {
 		// (See above).
 		detail::cgf_diagnostics::teardown();
-		return tm.submit_command_group([&, er = execution_range](handler& cgh) {
+		auto cg = detail::invoke_command_group_function([&, er = execution_range](handler& cgh) {
 			cgf(cgh);
 			cgh.parallel_for<KernelName>(er, [](nd_item<KernelDims>) {});
 		});
+		return tm.generate_command_group_task(std::move(cg));
 		detail::cgf_diagnostics::make_available();
```
fknorr, line 337: Fixing an unreachable code warning now, how did that ever work?!

psalz: Oops! Well this line not being executed just means that the CGF diagnostics are disabled for the remainder of the test. So it just wouldn't have caught any accessor-misue later on. But I doubt that there's any test that uses both `add_compute_task` et al., as well as a real command group where the diagnostics would matter. In fact, I just removed the call to `teardown` and it looks like we have exactly 1 test that uses `add_compute_task` with a real buffer, producing that warning (all others use mock buffers).

fknorr: Ironicallly, the test in question `get_access can be called on const buffer` does not call `get_access` on a const buffer. I'm refactoring it and splitting it into a RT test and a RT deprecation test to give it some meaning.

fknorr: I've pushed another commit that also moves non-RT sibling tests out to `task_graph_tests`. We should get rid of `add_*_task`, but this will have to be alongside a TDAG generator / TDAG test context refactoring.

src/named_threads.cc
fknorr: I found a way to move this code back to tracy.h by stringifying / casting `thread_type` in `named_threads`, and applying them in `tracy.h`.

fknorr:
> Edit: one final thing -- this is mostly an internal change, but it does change the (exposed) API surface, so maybe we should have a line for it in the changelog?

It doesn't touch the public API, unless I missed exporting something.

PeterTh (approved):
LGTM!

PR #309, opened 2024-11-18
Introduce a thread pinning mechanism

PeterTh:
This is a followup to #303.

As the backend executor requires low-latency communication with the submission threads, it is advantageous for these to be close in terms of cache hierarchy. On larger systems, we've observed another 50%+ performance increase due to this change in extreme situations.

In terms of interface design, the goal was to provide a very simple entry point (`pin_this_thread`), that is safe to use from any thread at any time, and does not require polluting any other modules with state related to thread pinning.  
Any pinning of the user thread also needs to be safe in terms of restoration, and the whole mechanism needs to work in the presence of process-level affinity masking (e.g. from MPI).

This PR also:
- Removes the ad-hoc core number check and replaces it with a better mechanism
- Adds extensive unit testing for this tool
- Pins the threads of non-runtime (micro-)benchmarks, to hopefully increase result consistency
- Takes this opportunity to improve documentation, unifying and fixing missing installation documentation, and providing full configuration documentation in a separate file


src/affinity.cc
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if(start < str.size()) result.push_back(std::string(str.substr(start)));
+		return result;
+	}
+} // namespace
+
+environment_configuration parse_validate_env(const std::string_view str) {
+	using namespace std::string_view_literals;
+	constexpr const char* error_msg =
+	    "Cannot parse CELERITY_THREAD_PINNING setting, needs to be either 'auto', 'from:#', comma-separated core list, or bool: {}";
```
fknorr, line 48: Just a note that we should document the core-list somewhere, i.e. which index is which thread and also how this interacts with legacy mode.

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,206 @@
...
+			// find the next sequential core we may use
+			auto next_available_core = [&]() {
+				if(!g_state.config.enabled) return 0u;
+				while(CPU_ISSET(current_core, &g_state.available_cores) == 0 && current_core < CPU_SETSIZE) {
+					current_core++;
+				}
+				if(current_core >= CPU_SETSIZE) {
+					CELERITY_WARN("Ran out of available cores for thread pinning after assigning {} cores, will disable pinning.", assigned_cores);
+					g_state.config.enabled = false;
+					return 0u;
```
fknorr, line 99: This is oddly non-local, maybe `throw std::out_of_range` here and surround L105-L115 with a try-catch? Would also get rid of warning spam.

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,206 @@
...
+	if(ret != 0) {
+		CELERITY_WARN("Error retrieving initial process affinity mask, will disable pinning.");
+		g_state.config.enabled = false;
+		return;
 	}
 
-} // namespace detail
-} // namespace celerity
+	// pinned threads per process: user, scheduler, executor, 1 backend worker per device
+	uint32_t pinned_threads_per_process = 3 + g_state.config.num_devices;
```
fknorr, line 61: Devices should only be enumerated when backend workers / device submission threads are enabled, right?

test/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ file(GLOB_RECURSE TEST_INCLUDES *.h)
 
 set(TEST_TARGETS
   accessor_tests
+  affinity_tests
```
fknorr, line 32: None of these tests will work on Windows, we should either make the test target conditional or have a `SKIP` in every test case.

fknorr (commented):
I love that we're finally integrating this into Celerity itself, and that you managed to do so without involving hwloc!

It would be neat to also pin threads for non-system benchmarks - AFAICT this should be as simple as keeping a `thread_pinner` instance in every `benchmark_context` class.

include/affinity.h
```diff
@@ -1,15 +1,83 @@
...
 
-	uint32_t affinity_cores_available();
+constexpr uint32_t thread_type_step = 10000;
 
-	/* a priori we need 3 threads, plus 1 for parallel-task workers and at least one more for host-task.
-	 This depends on the application invoking celerity. */
-	constexpr static uint64_t min_cores_needed = 5;
+// The threads Celerity interacts with ("user") and creates (everything else), identified for the purpose of pinning.
+// Note: this is not an enum class to make interactions such as specifying `first_backend_worker+i` easier
+enum thread_type : uint32_t {
```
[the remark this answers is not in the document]
PeterTh: Disabling this, it's very silly.

include/affinity.h
```diff
@@ -1,15 +1,83 @@
...
+	user = 0 * thread_type_step,
```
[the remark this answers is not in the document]
PeterTh: Changed to `application` in this PR, and also changed all existing "main thread" mentions in the code.

src/affinity.cc
```diff
@@ -0,0 +1,89 @@
...
+	// "from:" case
+	constexpr auto from_prefix = "from:"sv;
+	if(str.starts_with(from_prefix)) {
+		try {
+			const auto from = env::default_parser<uint32_t>{}(std::string(str.substr(from_prefix.size())));
+			return {true, from, {}};
+		} catch(const env::parser_error& e) {
+			CELERITY_ERROR(error_msg, e.what());
+			return {};
+		}
```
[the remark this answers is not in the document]
PeterTh: Absolutely correct, changed. I should really know this.

src/config.cc
```diff
@@ -121,8 +122,10 @@ namespace detail {
 			}
 #endif // CELERITY_WORKAROUND(SIMSYCL)
 
+			m_thread_pinning_config = parsed_and_validated_envs.get_or(env_thread_pinning, {});
+
 			const auto has_dry_run_nodes = parsed_and_validated_envs.get(env_dry_run_nodes);
-			if(has_dry_run_nodes) { m_dry_run_nodes = *has_dry_run_nodes; }
+			if(has_dry_run_nodes) { m_dry_run_nodes = static_cast<int>(*has_dry_run_nodes); }
```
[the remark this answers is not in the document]
PeterTh: Agreed, changed.

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,206 @@
...
+	}
+
+	// allow the affinity of this thread to be restored on teardown
+	{
+		const pinned_thread_state ps{
+		    .previous_cpuset = previous_cpuset,
+		    .pinned_to_cpuset = new_cpuset,
+		};
+		g_state.pinned_threads.emplace(this_thread, ps);
+		if(!t_remover.has_value()) t_remover.emplace();
```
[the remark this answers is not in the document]
PeterTh: No, in the unlikely but supported case of two or more consecutive pinning regions with the same involved thread, this branch is taken (and works as designed, added a test for that).

src/runtime.cc
```diff
@@ -325,6 +331,9 @@ namespace detail {
 		// Create and await the shutdown epoch
 		sync(epoch_action::shutdown);
 
+		// This only affects non-functional behaviour, so we can safely do it first
+		m_thread_pinner.reset();
```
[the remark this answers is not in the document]
PeterTh: Good point, removed!

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,206 @@
...
+	uint32_t pinned_threads_per_process = 3 + g_state.config.num_devices;
+	// total number of threads to be pinned across processes (legacy mode)
+	uint32_t total_threads = pinned_threads_per_process * g_state.config.num_legacy_processes;
+
+	if(g_state.config.enabled) {
+		// build our pinning plan
+		if(!cfg.hardcoded_core_ids.empty()) {
+			// just use the provided hardcoded IDs
+			if(static_cast<uint32_t>(cfg.hardcoded_core_ids.size()) != total_threads) {
+				CELERITY_WARN("Hardcoded core ID count ({}) does not match the number of threads to be pinned ({}), will disable pinning.",
```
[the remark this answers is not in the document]
PeterTh: I'd rather keep this behaviour, if the user goes to the trouble of providing a full core list they probably really want that.

src/platform_specific/affinity.unix.cc
PeterTh: Refactored this whole thing to be much better, so this no longer exists.

src/platform_specific/affinity.unix.cc
PeterTh: Very good point. Handled+ tested this case.

test/CMakeLists.txt
PeterTh: Done. (A few tests should work, i.e. parsing and stringification, so I used skip)

src/platform_specific/affinity.unix.cc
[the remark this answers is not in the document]
fknorr: Yes the warning should definitely remain here, I just think it would be less surprising if we revert to autopinning on failure (i.e. the default with no environment specified) instead of disabling pinning altogether.

src/affinity.cc
PeterTh: Done in the upcoming documentation overhaul. However, since this is a very low-level feature I didn't duplicate the information in the docs, since it would probably just get outdated. I rather just point to the relevant code.

test/test_utils.h
```diff
@@ -156,6 +156,20 @@ namespace detail {
...
+		benchmark_thread_pinner() {
+			const detail::thread_pinning::runtime_configuration cfg{
+			    .enabled = true,
+			    .use_backend_device_submission_threads = false,
+			};
+			m_thread_pinner.emplace(cfg);
+			detail::thread_pinning::pin_this_thread(detail::thread_pinning::thread_type::application);
+		}
+
+		std::optional<detail::thread_pinning::thread_pinner> m_thread_pinner;
```
fknorr, line 170:
```suggestion
		benchmark_thread_pinner() {
			detail::thread_pinning::pin_this_thread(detail::thread_pinning::thread_type::application);
		}

		detail::thread_pinning::thread_pinner m_thread_pinner{{
			.enabled = true,
			.use_backend_device_submission_threads = false,
		}};
```
Just out of curiosity, can designated initializers be used in this fashion?

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,215 @@
...
+				    cfg.hardcoded_core_ids.size(), total_threads);
+			} else {
+				selected_core_ids = cfg.hardcoded_core_ids;
+			}
+		} else {
+			// otherwise, sequential core assignments for now; it is most important that each of the threads is "close"
+			// to the ones next to it in this sequence, so that communication between them is fast
+			selected_core_ids = get_available_sequential_cores(g_state.available_cores, total_threads, cfg.standard_core_start_id);
```
psalz, line 104:
```suggestion
			selected_core_ids = get_available_sequential_cores(g_state.available_cores, total_threads, cfg.standard_core_start_id);
			if(selected_core_ids.empty()) {
			  CELERITY_WARN("...")
			}
```

Instead of printing the warning in `get_available_sequential_cores`, we should print it here and include the number of cores that would be required. Otherwise this carries less information than the warning for when pinning is disabled!

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+		    fmt::format("Celerity has detected that only {} logical cores are available to this process. It is recommended to assign at least {}", 3, 6)));
+	}
+}
+
+TEST_CASE("do not plan for device submission threads if they are unused", "[affinity]") {
+	skip_unsupported_configurations();
+	raii_affinity_masking mask({1, 2, 3, 4});
+	const detail::thread_pinning::runtime_configuration cfg = {.enabled = true, .num_devices = 10, .use_backend_device_submission_threads = false};
+	detail::thread_pinning::thread_pinner pinner(cfg);
+	// no additional check, a warning will make the test fail if we do not handle this case correctly
```
psalz, line 173:
```suggestion
	SUCCEED(); // no additional check, a warning will make the test fail if we do not handle this case correctly
```

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+	CHECK(test_utils::log_contains_substring(
+	    detail::log_level::warn, fmt::format("only {} logical cores are available to this process. It is recommended to assign at least", 1)));
+	CHECK_FALSE(test_utils::log_contains_substring(detail::log_level::debug, "Affinity"));
+}
+
+TEST_CASE_METHOD(test_utils::runtime_fixture, "the application thread is actually pinned when pinning is enabled", "[affinity][runtime]") {
+	skip_unsupported_configurations();
+	env::scoped_test_environment ste("CELERITY_THREAD_PINNING", "auto");
+
+	const core_set process_mask = {3, 4, 5, 6, 7};
```
psalz, line 236: Why do we need this specific set, starting from 3? Couldn't this test also run on a 6-core machine? Same below.

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+		CPU_ZERO(&cpuset);
+		CPU_SET(4, &cpuset);
+		REQUIRE(pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset) == 0);
+	}
+
+	CHECK(test_utils::log_contains_substring(
+	    detail::log_level::warn, fmt::format("Thread affinity of thread {} was changed unexpectedly, skipping restoration.", pthread_self())));
+}
+
+TEST_CASE("multiple subsequent non-overlapping pinning ranges are handled correctly", "[affinity]") {
```
psalz, line 295: Wait, what is this testing? If the thread pinner goes out of scope, it should restore the previous state, right? Why is the "non-overlapping" part relevant?

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,215 @@
...
+namespace {
 
-	uint32_t affinity_cores_available() {
-		cpu_set_t available_cores;
-		[[maybe_unused]] const auto ret = pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &available_cores);
-		assert(ret == 0 && "Error retrieving affinity mask.");
-		return CPU_COUNT(&available_cores);
+using namespace celerity::detail::thread_pinning;
+
+std::vector<uint32_t> get_available_sequential_cores(const cpu_set_t& available_cores, const uint32_t count, const uint32_t starting_from_core) {
```
psalz, line 18: Don't we want sequential as in next to each other? As far as I can tell this function allows for gaps?

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,215 @@
...
+	if(g_state.config.enabled) {
+		// select the core set to use
+		std::vector<uint32_t> selected_core_ids = {};
+		if(!cfg.hardcoded_core_ids.empty()) {
+			// just use the provided hardcoded IDs
+			if(static_cast<uint32_t>(cfg.hardcoded_core_ids.size()) != total_threads) {
+				CELERITY_WARN("Hardcoded core ID count ({}) does not match the number of threads to be pinned ({}), will disable pinning.",
+				    cfg.hardcoded_core_ids.size(), total_threads);
+			} else {
+				selected_core_ids = cfg.hardcoded_core_ids;
```
psalz, line 99: Don't we also have to check whether the specific cores are even available to the process?

psalz (commented):
Good stuff! A couple of notes.

Also, I remember we talked about whether we should detect whether the application thread is already pinned, and not do anything in that case. Did you end up deciding against that?

include/affinity.h
```diff
@@ -1,15 +1,87 @@
...
-	/* a priori we need 3 threads, plus 1 for parallel-task workers and at least one more for host-task.
-	 This depends on the application invoking celerity. */
-	constexpr static uint64_t min_cores_needed = 5;
+// The threads Celerity interacts with ("user") and creates (everything else), identified for the purpose of pinning.
+// Note: this is not an enum class to make interactions such as specifying `first_backend_worker+i` easier
+enum thread_type : uint32_t {
+	application = 0 * thread_type_step,
+	scheduler = 1 * thread_type_step,
+	executor = 2 * thread_type_step,
+	first_backend_worker = 3 * thread_type_step,
```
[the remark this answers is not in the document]
PeterTh: Since it's `scheduler` and `executor` I'll go with `first_device_submitter`

src/platform_specific/affinity.unix.cc
PeterTh: It's as sequential as we can make it. It's not necessarily *contiguous*, but it's also not called that ;)

src/platform_specific/affinity.unix.cc
PeterTh: It will fail later when pinning so I didn't want to add extra code here for this edge case.

src/platform_specific/affinity.unix.cc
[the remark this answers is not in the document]
PeterTh: Done

include/affinity.h
```diff
@@ -1,15 +1,87 @@
...
+	uint32_t num_legacy_processes = 1;
+	// Process index of current process running in legacy mode
+	uint32_t legacy_process_index = 0;
+
+	// The core to start pinning "standard" threads to
+	uint32_t standard_core_start_id = 1;
+
+	// If set, this list of core ids will be used for pinning instead of the default round-robin scheme
+	// The list must contain exactly as many elements as there are standard threads
+	std::vector<uint32_t> hardcoded_core_ids = {}; // NOLINT(readability-redundant-member-init) -- to allow partial designated init elsewhere
```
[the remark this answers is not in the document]
PeterTh: Yes. I'm not sure I'd label this a clang-tidy "bug" as such though. It's an unfortunate interaction between 2 rules.

test/affinity_tests.cc
PeterTh: Added. Out of curiosity, what does this change?

include/affinity.h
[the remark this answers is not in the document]
fknorr: The comment still says "user".

ci/run-benchmarks.sh
```diff
@@ -13,6 +13,8 @@ fi
 
 # we use taskset to pin the process to a specific set of cores and their HTs to reduce benchmark result variance
 # this set is chosen to all be located on the second die of the Threadripper 2920X CPU in our current CI benchmark system "gpuc2"
+# since this is a very rare system with only 3 cores per CC, we use a hardcoded thread pinning strategy which places the 4 BE worker threads on HTs
+export CELERITY_THREAD_PINNING=7,8,9,10,11,22,23
 bash /root/capture-backtrace.sh taskset -c 6-11,18-23 test/benchmarks \
```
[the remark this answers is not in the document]
PeterTh: No, removed. (until the results tell me otherwise at least)

src/affinity.cc
```diff
@@ -0,0 +1,83 @@
...
+			} catch(const env::parser_error& e) { throw env::parser_error{fmt::format(error_msg, e.what())}; }
+		}
+		return {true, 0, core_ids};
+	}
+
+	// if all else fails, assume we have a boolean
+	try {
+		return {env::default_parser<bool>{}(str), auto_start_from_core, {}};
+	} catch(const env::parser_error& e) { throw env::parser_error{fmt::format(error_msg, e.what())}; }
+	return {};
```
fknorr, line 80: This line is unreachable.

src/affinity.cc
```diff
@@ -0,0 +1,83 @@
...
+			result.push_back(std::string(str.substr(start, end - start)));
+			start = end + 1;
+			end = str.find(delim, start);
+		}
+		if(start < str.size()) result.push_back(std::string(str.substr(start)));
+		return result;
+	}
+} // namespace
+
+environment_configuration parse_validate_env(const std::string_view str) {
```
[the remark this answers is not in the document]
PeterTh: It was when I started implementing it, but I purposefully moved it since I think it's better for this level of parsing to be in the module it interacts with rather than the generic config file (where it detracts from that code's main purpose).

I don't think including libenvpp in two TUs is a problem, we're not talking about a header.

src/platform_specific/affinity.unix.cc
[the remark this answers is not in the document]
fknorr: Behavior hasn't changed: When the warning prints (still says "disable" btw),  `selected_core_ids` remains empty which will take the branch in L180 and set `enabled=false`.

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+bool have_cores(const core_set& desired_set) {
+	auto current_set = get_current_cores();
+	return std::ranges::all_of(desired_set, [&](uint32_t core) { return current_set.contains(core); });
+}
+
+} // namespace
+
+TEST_CASE("semantic thread type enum entries can be turned into strings", "[affinity]") {
+	using namespace detail::thread_pinning;
+	auto to_tt([&](uint32_t t) { return static_cast<thread_type>(t); });
```
fknorr, line 90:
```suggestion
	const auto to_tt = [&](uint32_t t) { return static_cast<thread_type>(t); };
```
This is very odd to read.

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+	core_set cores;
+	for(uint32_t i = 0; i < CPU_SETSIZE; ++i) {
+		if(CPU_ISSET(i, &mask)) { cores.insert(i); }
+	}
+	return cores;
+}
+
+bool have_cores(const core_set& desired_set) {
+	auto current_set = get_current_cores();
+	return std::ranges::all_of(desired_set, [&](uint32_t core) { return current_set.contains(core); });
```
fknorr, line 83:
```suggestion
	const auto current_set = get_current_cores();
	return std::ranges::all_of(desired_set, [&](const uint32_t core) { return current_set.contains(core); });
```

src/platform_specific/affinity.unix.cc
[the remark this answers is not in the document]
PeterTh: I haven't committed this stuff yet, sorry. (that comment was just me ticking off everything I've done locally)

ci/run-benchmarks.sh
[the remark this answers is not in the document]
PeterTh: And the answer is in, we still need it. Otherwise the non-runtime benchmarks run on a different set of cores, unless either (i) we hardcode that in the benchmark pinner, or (ii) we manually interpret the environment variable in the non-runtime benchmarks.

I think both of those are worse than keeping the taskset, so I'll do that.

test/affinity_tests.cc
```diff
@@ -0,0 +1,323 @@
...
+	~raii_test_runtime() { celerity::runtime::shutdown(); }
+	raii_test_runtime(const raii_test_runtime&) = delete;
+	raii_test_runtime(raii_test_runtime&&) = delete;
+	raii_test_runtime& operator=(const raii_test_runtime&) = delete;
+	raii_test_runtime& operator=(raii_test_runtime&&) = delete;
+};
+
+constexpr void skip_unsupported_configurations() {
+#ifdef _WIN32
+	SKIP("Affinity is not supported on Windows");
```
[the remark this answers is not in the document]
PeterTh: Changed to a macro.

test/affinity_tests.cc
PeterTh: That is odd indeed, must have been late last Saturday.

include/affinity.h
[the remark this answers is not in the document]
PeterTh: Yeah psalz also spotted that, fixed now

src/platform_specific/affinity.unix.cc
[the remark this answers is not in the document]
PeterTh: Same reason as the other one, was done locally but not committed yet, please re-check.

PeterTh:
All updates should now be in, including the previously discussed (but not implemented) feature which prevents the application thread from being pinned if it has already had its affinity mask modified by something else (+test).

ci/run-benchmarks.sh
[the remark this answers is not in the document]
fknorr: Aren't we now instantiating a thread pinner for all benchmarks anyway?

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,226 @@
...
+			}
+		}
+		if(selected_core_ids.empty()) {
+			// otherwise, sequential core assignments for now; it is most important that each of the threads is "close"
+			// to the ones next to it in this sequence, so that communication between them is fast
+			selected_core_ids = get_available_sequential_cores(g_state.available_cores, total_threads, cfg.standard_core_start_id);
+			if(selected_core_ids.empty()) {
+				CELERITY_WARN("Insufficient available cores for thread pinning (required {}, {} available), disabling pinning.", //
```
psalz, line 106:
```suggestion
				CELERITY_WARN("Insufficient logical cores available for thread pinning (required {}, {} available), disabling pinning. Performance may be negatively impacted.", //
```

To bring it a bit more in line with the other warning.

src/affinity.cc
```diff
@@ -0,0 +1,83 @@
...
+
+std::string thread_type_to_string(const thread_type t_type) {
+	switch(t_type) {
+	case thread_type::application: return "application";
+	case thread_type::scheduler: return "scheduler";
+	case thread_type::executor: return "executor";
+	default: break;
+	}
+	if(t_type >= thread_type::first_backend_worker && t_type < thread_type::first_host_queue) {
+		return fmt::format("backend_worker_{}", t_type - thread_type::first_backend_worker);
```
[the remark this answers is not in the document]
psalz: I shortened this to "submit" yesterday because I somehow thought this was about naming threads (as in, setting their name using the POSIX API, where we have a 16 character limit). While these are different names, unfortunately the actual thread names are too long as well since the "worker -> submission" renaming in #303:
image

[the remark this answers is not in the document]
psalz: Come to think of it, it's kind of confusing that we now have two different concepts of thread names, between what the pinner prints, and what we set using the POSIX API (with the "cy-" prefix).

src/platform_specific/affinity.unix.cc
```diff
@@ -1,20 +1,226 @@
...
+			g_state.thread_pinning_plan.emplace(thread_type::executor, selected_core_ids[current_core_id++]);
+			if(g_state.config.use_backend_device_submission_threads) {
+				for(uint32_t i = 0; i < g_state.config.num_devices; ++i) {
+					const auto device_tid =
+					    static_cast<thread_type>(thread_type::first_device_submitter + i); // NOLINT(clang-analyzer-optin.core.EnumCastOutOfRange)
+					g_state.thread_pinning_plan.emplace(device_tid, selected_core_ids[current_core_id++]);
+				}
+			}
+		}
+	} else {
```
psalz, line 126:
```suggestion
	} else {
	CELERITY_DEBUG("Thread pinning is disabled.");
```

Probably wouldn't hurt to have, in case someone sets `CELERITY_THREAD_PINNING=off` in their environment and forgets about it.

src/platform_specific/affinity.unix.cc
psalz: Actually the warning can get nonsensical when using the `from:` syntax. I just got this:

image

src/platform_specific/affinity.unix.cc
psalz: I just tried `CELERITY_THREAD_PINNING="4,5,6,7,8,9,10" taskset -c 49 ./examples/matmul/matmul` and didn't get any errors. In fact I think it simply ignored the process mask. Is that working as intended?

src/platform_specific/affinity.unix.cc
psalz: Hmm, ok. It would probably be preferable to find a contiguous range, but in the end the whole thing is a simple heuristic anyway. Speaking of: We should probably include a comment here that this function assumes that cores that are close together in the cpu set are also close in terms of cache, that we currently can't properly handle hyperthreads and that doing anything more fancy would require something like hwloc!

test/affinity_tests.cc
psalz: Kind of, "overlapping lifetime" is clear to me, but "non-overlapping pinning ranges" makes it sound that its about the ranges (which are in fact overlapping).

include/affinity.h
```diff
@@ -1,15 +1,90 @@
...
+	std::vector<uint32_t> hardcoded_core_ids; // starts empty, which means no hardcoded IDs are used
+};
+
+// Parses and validates the environment variable string, returning the corresponding configuration
+environment_configuration parse_validate_env(const std::string_view str);
+
+// Configures the pinning mechanism
+// For now, only "standard" threads are pinned
+//   these are threads that benefit from rapid communication between each other,
+//   i.e. applciation -> scheduler -> executor -> device submission threads
```
psalz, line 46:
```suggestion
//   i.e. application -> scheduler -> executor -> device submission threads
```

ci/run-benchmarks.sh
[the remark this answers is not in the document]
PeterTh: We are, I'm talking about configuring the thread pinner.
By default it would take a different core than what used to be the case.

test/affinity_tests.cc
PeterTh: Ah, I'll change the wording.

src/platform_specific/affinity.unix.cc
PeterTh: I think the function doesn't really need that comment, the fact that we use it as the "auto" core pinning strategy does. But we have a comment to that effect already there.

src/platform_specific/affinity.unix.cc
PeterTh: Absolutely not intended, thanks for checking. I wrongly assumed that the `pthread_setaffinity_np` call would fail.

The behaviour for your example is now this:
```
[2024-11-21 11:28:59.479] [0] [warning] Not all hardcoded core IDs are available, downgrading to auto-pinning.
[2024-11-21 11:28:59.479] [0] [warning] Insufficient available cores for thread pinning (required 7, 1 available), disabling pinning.
```
There's also a new test for it.

src/platform_specific/affinity.unix.cc
PeterTh: I made it a bit more meaningful.

ci/run-benchmarks.sh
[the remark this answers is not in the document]
fknorr: Mhh we don't parse the environment in non-RT tests. Makes sense.

fknorr (approved):

psalz (approved):

PR #310, opened 2024-11-21
Refactoring and unifying various thread-related code

PeterTh:
This refactoring PR performs 2 main tasks:

* Introduce a single source of truth for thread names, and a unified interface to both name and (if enabled and required) pin threads
* Also unify this with tracy thread ordering, so that there is only a single interface for all of it, and it takes a `named_threads` enum entry as its parameter

There are no functional changes, test coverage is slightly extended.

Note that the implementation of how the try/catch block in threads is designed and what happens in response to exceptions had already slightly diverged (presumably unintentionally).

Check-perf-impact result (no new file required):
:heavy_check_mark: No significant performance change in the microbenchmark set. You are good to go!

### DIFF
--- a/include/affinity.h
+++ b/include/affinity.h
@@ -1,14 +1,14 @@
 #pragma once
 
 #include <cstdint>
-#include <string>
 #include <string_view>
 #include <vector>
 
+#include "named_threads.h"
 
 // The goal of this thread pinning mechanism, when enabled, is to ensure that threads which benefit from fast communication
 // are pinned to cores that are close to each other in terms of cache hierarchy.
-// It currently accomplishes this by pinning threads to cores in a round-robin fashion according to their order in the `thread_type` enum.
+// It currently accomplishes this by pinning threads to cores in a round-robin fashion according to their order in the `named_threads::thread_type` enum.
 //
 // In terms of interface design, the goal is to provide a very simple entry point (`pin_this_thread`), that is safe to use from any thread at any time,
 // and does not require polluting any other modules with state related to thread pinning. The `thread_pinner` RAII class offers the only way to manage the
@@ -16,20 +16,6 @@
 //
 // TODO: A future extension would be to respect NUMA for threads performing memory operations, but this requires in-depth knowledge of the system's topology.
 namespace celerity::detail::thread_pinning {
-
-constexpr uint32_t thread_type_step = 10000;
-
-// The threads Celerity interacts with ("application") and creates (everything else), identified for the purpose of pinning.
-// Note: this is not an enum class to make interactions such as specifying `first_backend_worker+i` easier
-enum thread_type : uint32_t {
-	application = 0 * thread_type_step,
-	scheduler = 1 * thread_type_step,
-	executor = 2 * thread_type_step,
-	first_device_submitter = 3 * thread_type_step,
-	first_host_queue = 4 * thread_type_step,
-	max = 5 * thread_type_step,
-};
-std::string thread_type_to_string(const thread_type t_type);
 
 // User-level configuration of the thread pinning mechanism (set by the user via environment variables)
 struct environment_configuration {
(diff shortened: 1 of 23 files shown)


include/thread_queue.h
```diff
@@ -1,19 +1,38 @@
...
 #include <type_traits>
 #include <variant>
 
 namespace celerity::detail {
 
+// A helper to implement the body of a worker thread while taking care of all boilerplate code
+void thread_body(const named_threads::thread_type t_type, const std::invocable auto& fun) {
+	thread_pinning::name_and_pin_thread(t_type);
+	const auto name = named_threads::thread_type_to_string(t_type);
+	CELERITY_DETAIL_TRACY_SET_THREAD_NAME_AND_ORDER(tracy_detail::leak_name(name), tracy_detail::thread_order::thread_queue);
```
PeterTh, line 23: This is wrong, should just change the tracy macro to take a `thread_type`.

include/affinity.h
```diff
@@ -83,8 +70,12 @@ class thread_pinner {
 	bool m_successfully_initialized = false;
 };
 
+// Performs both naming and pinning (if enabled for this thread) of the invoking thread
+// This should be the first thing called in any thread that is part of the Celerity runtime
+void name_and_pin_thread(const named_threads::thread_type t_type);
```
fknorr, line 75:
```suggestion
void name_and_pin_this_thread(const named_threads::thread_type t_type);
```
For consistency with `pin_this_thread`?

include/thread_queue.h
```diff
@@ -1,19 +1,39 @@
...
+	CELERITY_DETAIL_TRACY_SET_THREAD_NAME_AND_ORDER(t_type);
+	try {
+		fun();
+	}
+	// LCOV_EXCL_START
+	catch(std::exception& e) { //
+		utils::panic("exception in {}: {}", name, e.what());
+	} catch(...) { //
+		utils::panic("exception in {}", name);
+	}
```
fknorr, line 33: It was probably me who came up with this pattern, but I'm having second thoughts. There's one or two hard to reproduce bugs in the runtime right now (`std::get wrong index for variant`) which produce a crash dump that shows the `panic` as the reason of fault instead of the stack trace for the exception causing it. If we let the exception terminate the program, we would get a proper trace.

Maybe a try-catch would be better suited around invocations of user code such as host tasks.

src/named_threads.cc
```diff
@@ -0,0 +1,42 @@
...
+std::string thread_type_to_string(const thread_type t_type) {
+	switch(t_type) {
+	case thread_type::application: return "cy-application";
+	case thread_type::scheduler: return "cy-scheduler";
+	case thread_type::executor: return "cy-executor";
+	case thread_type::alloc: return "cy-alloc";
+	default: break;
+	}
+	if(t_type >= thread_type::first_device_submitter && t_type < thread_type::first_host_queue) { //
+		return fmt::format("cy-be-sub-{}", static_cast<uint32_t>(t_type) - static_cast<uint32_t>(thread_type::first_device_submitter));
```
fknorr, line 31: How about `cy-submit-{}`? "be" == backend is not even part of the enum variant anymore.

fknorr (commented):
I like the centralized naming!

include/thread_queue.h
PeterTh: That's a good point, maybe we should just not catch it and let the program fail?
(In the distant future we could use standardized mechanisms to get a proper stack trace...)

psalz (approved):
LGTM!

src/named_threads.cc
PeterTh: Went with `dev-sub` since you make a good point about trying to match the enum.

include/thread_queue.h
PeterTh: We decided offline that for debugging purposes the best thing to do is to just let the default handler happen in all circumstances.

src/runtime.cc
```diff
@@ -1,29 +1,5 @@
...
-#include "reduction.h"
-#include "scheduler.h"
-#include "system_info.h"
-#include "task.h"
-#include "task_manager.h"
-#include "tracy.h"
-#include "types.h"
-#include "utils.h"
-#include "version.h"
-
```
fknorr, line 26: Automerge gone wrong?

fknorr (approved):
Mostly LGTM!

PR #311, opened 2024-11-22
[clang-tidy] Disable lints that frequently cause false positives

### LINKED ISSUE #127: Decide on which clang-tidy checks we want to enable
I've [parsed](https://gist.github.com/psalz/5f936894a451f25e2b9d1d6f66915307) all clang-tidy diagnostics generated with the following checks enabled:
```
-*,
bugprone-*,
clang-analyzer-*,
clang-diagnostic-*,
cppcoreguidelines-*,
mpi-*,
performance-*,
readability-*
```

and those are the results:
```
Found 2137 diagnostics (out of 13487 parsed; 11350 were duplicates). 
  787: cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers 
  678: readability-identifier-length
  104: readability-function-cognitive-complexity
   92: readability-qualified-auto
   41: cppcoreguidelines-avoid-c-arrays
   37: cppcoreguidelines-pro-bounds-pointer-arithmetic
   36: cppcoreguidelines-init-variables
   30: cppcoreguidelines-special-member-functions
   23: bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions
   22: cppcoreguidelines-macro-usage
   22: performance-unnecessary-value-param
   19: cppcoreguidelines-avoid-non-const-global-variables
   17: readability-implicit-bool-conversion
   16: readability-isolate-declaration
   16: readability-named-parameter
   16: bugprone-macro-parentheses
   15: bugprone-easily-swappable-parameters
   14: cppcoreguidelines-pro-bounds-constant-array-index
   12: bugprone-exception-escape
   12: cppcoreguidelines-pro-bounds-array-to-pointer-decay
   11: cppcoreguidelines-pro-type-member-init
   10: clang-analyzer-deadcode.DeadStores
   10: bugprone-implicit-widening-of-multiplication-result
   10: cppcoreguidelines-pro-type-static-cast-downcast
    7: cppcoreguidelines-pro-type-vararg
    7: performance-move-const-arg
    7: cppcoreguidelines-non-private-member-variables-in-classes
    6: readability-braces-around-statements
    6: readability-else-after-return
    6: cppcoreguidelines-pro-type-reinterpret-cast
    4: readability-redundant-access-specifiers
    4: cppcoreguidelines-owning-memory
    4: readability-avoid-const-params-in-decls
    3: clang-analyzer-core.UndefinedBinaryOperatorResult
    3: performance-faster-string-find
    3: cppcoreguidelines-pro-type-const-cast
    2: readability-simplify-boolean-expr
    2: performance-unnecessary-copy-initialization
    2: performance-for-range-copy
    2: readability-container-size-empty
    2: readability-inconsistent-declaration-parameter-name
    2: readability-static-accessed-through-instance
    2: clang-analyzer-optin.mpi.MPI-Checker
    2: performance-noexcept-move-constructor
    2: bugprone-lambda-function-name
    2: readability-convert-member-functions-to-static
    2: clang-diagnostic-error
    1: performance-inefficient-vector-operation
    1: readability-suspicious-call-argument
    1: clang-analyzer-core.CallAndMessage
    1: readability-make-member-function-const
    1: cppcoreguidelines-prefer-member-initializer
```

You can find the raw clang-tidy output [here](https://gist.githubusercontent.com/psalz/d3934f47c0010d24de0cafd20b7b8898/raw/7bc19ac8cd641e0cd3ecc9f20fa3b6c2504a7628/gistfile1.txt) (useful for checking what some of these diagnostics even do).

Based on this, I would suggest the following configuration:
```
-*,
bugprone-*,
clang-analyzer-*,
clang-diagnostic-*,
cppcoreguidelines-*,
-cppcoreguidelines-avoid-c-arrays,
-cppcoreguidelines-macro-usage,
-cppcoreguidelines-pro-bounds-pointer-arithmetic,
mpi-*,
performance-*,
readability-*,
-readability-avoid-const-params-in-decls
-readability-identifier-length,
-readability-magic-numbers,
-readability-uppercase-literal-suffix,
```

Thoughts?

fknorr:
Resolves #127.

There's two lints we would like to have but can't enable fully due to FPs:
- `misc-const-correctness` will suggest `const accessor` and `const std::lock_guard`, which is just noise. I've enabled this in the runtime, but disabled it for examples and tests - we don't constuct too many accessors in runtime code itself.
- `misc-include-cleaner` doesn't understand "interface headers" like `<sycl/sycl.hpp>` and will insist we include `<hipSYCL/sycl/handler.hpp>` et al. It also removes includes that are necessary for trait implementations, such as those of `fmt::formatter`.

I've manually performed autofixes for both these lints and added sensible corrections resulting from them as individual commits in this PR to clean up the codebase somewhat.

I also used the include-cleaner mayhem as an opportunity to set a consistent `#include` schema:

```c++
#pragma once // ONLY in my_interface.h
#include "my_interface.h" // ONLY in my_interface.cc

// 1 blank line, then all other headers form my module
#include "sibling_header_1.h"
#include "sibling_header_2.h"

// 1 blank line, then everything from the stdlib
#include <stdlib_header>
#include <stdlib_header_2>

// 1 blank line, then all 3rdparty deps
#include <3rdparty1/header.h>
#include <3rdparty2/header.h>

// 1 blank line, then any conditional includes (and maybe defines)
#ifdef ANY_CONDITIONAL_INCLUDES
#include <whatever.h>
#endif


// 2 blank lines, then the first actual token in the file
namespace celerity::detail {
...
```

I'm including my own headers before the stdlib or 3rdparty dependencies, because this is more likely to lead to a compiler error when another header forgets to include something. clang-format is smart enough to alphabetically sort includes within those groups.

I didn't bother with updating includes in tests in this PR.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -11,20 +11,29 @@
 #    `git diff -U0 --no-color | clang-tidy-diff.py -p1 -path path/to/compile_commands.json`
 #
 InheritParentConfig: false
+
 # See https://clang.llvm.org/extra/clang-tidy/checks/list.html for a full list of available checks.
-# Note: We would like to enable `misc-const-correctness` (introduced with Clang 15), but it currently
-#       seems to be somewhat buggy still (producing false positives) => revisit at some point.
+# We disable a number of checks that cause frequent false positives, including:
+#   -bugprone-unchecked-optional-access treats std::optional::value as "unchecked"
+#   -misc-include-cleaner can't deal with "interface headers" such as sycl.hpp
+#   -misc-unused-using-decls complains about "using sub_group = sycl::sub_group" (API export)
+#   -readability-convert-member-functions-to-static lints against implementations of fmt::formatter
 Checks: -*,
   bugprone-*,
   -bugprone-easily-swappable-parameters,
   -bugprone-lambda-function-name,
   -bugprone-macro-parentheses,
+  -bugprone-unchecked-optional-access,
   misc-*,
-  -misc-const-correctness,
+  -misc-include-cleaner,
+  -misc-misplaced-const,
   -misc-no-recursion,
   -misc-non-private-member-variables-in-classes,
   -misc-unused-parameters,
+  -misc-unused-using-decls,
+  -misc-use-anonymous-namespace,
   clang-analyzer-*,
+  -clang-analyzer-optin.mpi.MPI-Checker,
   clang-diagnostic-*,
   cppcoreguidelines-*,
   -cppcoreguidelines-avoid-c-arrays,
@@ -32,19 +41,28 @@ Checks: -*,
   -cppcoreguidelines-avoid-magic-numbers,
   -cppcoreguidelines-macro-usage,
   -cppcoreguidelines-non-private-member-variables-in-classes,
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
+  -cppcoreguidelines-pro-bounds-constant-array-index,
   -cppcoreguidelines-pro-bounds-pointer-arithmetic,
+  -cppcoreguidelines-pro-type-const-cast,
+  -cppcoreguidelines-pro-type-reinterpret-cast,
   mpi-*,
   performance-*,
   -performance-enum-size,
   readability-*,
   -readability-avoid-const-params-in-decls,
+  -readability-convert-member-functions-to-static,
+  -readability-else-after-return,
   -readability-function-cognitive-complexity,
   -readability-identifier-length,
   -readability-magic-numbers,
   -readability-qualified-auto,
+  -readability-redundant-inline-specifier,
   -readability-uppercase-literal-suffix,
 
 CheckOptions:
+  - key: misc-const-correctness.WarnPointersAsValues
+    value: true
   # Naming conventions
   - key: readability-identifier-naming.ClassCase
     value: lower_case
--- a/include/recorders.h
+++ b/include/recorders.h
@@ -4,7 +4,6 @@
 #include "grid.h"
 #include "instruction_graph.h"
 #include "intrusive_graph.h"
-#include "matchbox.hh"
 #include "nd_memory.h"
 #include "pilot.h"
 #include "ranges.h"
@@ -23,6 +22,8 @@
 #include <utility>
 #include <variant>
 #include <vector>
+
+#include <matchbox.hh>
 
 
 namespace celerity::detail {
--- a/include/recorders.h
+++ b/include/recorders.h
@@ -1,11 +1,26 @@
 #pragma once
 
 #include "command_graph.h"
+#include "grid.h"
 #include "instruction_graph.h"
+#include "intrusive_graph.h"
+#include "matchbox.hh"
+#include "nd_memory.h"
 #include "pilot.h"
+#include "ranges.h"
+#include "sycl_wrappers.h"
 #include "task.h"
+#include "types.h"
 
+#include <algorithm>
+#include <cassert>
+#include <cstddef>
 #include <functional>
+#include <memory>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <utility>
 #include <variant>
 #include <vector>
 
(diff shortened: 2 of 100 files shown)


.clang-tidy
```diff
@@ -11,40 +11,58 @@
...
   -cppcoreguidelines-pro-bounds-pointer-arithmetic,
+  -cppcoreguidelines-pro-type-const-cast,
+  -cppcoreguidelines-pro-type-reinterpret-cast,
   mpi-*,
   performance-*,
   -performance-enum-size,
   readability-*,
   -readability-avoid-const-params-in-decls,
+  -readability-convert-member-functions-to-static,
+  -readability-else-after-return,
```
PeterTh, line 55: I'm not sure about getting rid of this one. I ran into it 3 times so far, but each time I think the code after resolving it "naturally" was better. Do you have any scenario where it really fails?

.clang-tidy
```diff
@@ -11,40 +11,58 @@
...
   -readability-function-cognitive-complexity,
   -readability-identifier-length,
   -readability-magic-numbers,
   -readability-qualified-auto,
+  -readability-redundant-inline-specifier,
```
PeterTh, line 60: What does this fail on? Never saw it.

psalz (approved):
I expected the diff to be a lot worse! Thanks!

> I also used the include-cleaner mayhem as an opportunity to set a consistent #include schema:

Should we add that guide to CONTRIBUTING.md? 🤔

include/recorders.h
```diff
@@ -1,11 +1,26 @@
 #pragma once
 
 #include "command_graph.h"
+#include "grid.h"
 #include "instruction_graph.h"
+#include "intrusive_graph.h"
+#include "matchbox.hh"
```
PeterTh, line 7: Shouldn't this be in a different group?

src/backend/sycl_cuda_backend.cc
```diff
@@ -1,15 +1,29 @@
 #include "backend/sycl_backend.h"
 
-#include <cuda_runtime.h>
```
PeterTh, line 3: I'm confused at this not requiring a cuda header.

src/instruction_graph_generator.cc
```diff
@@ -2,23 +2,46 @@
 
 #include "access_modes.h"
 #include "command_graph.h"
+#include "dense_map.h"
+#include "fmt/base.h"
```
PeterTh, line 6: Wrong goup.

PeterTh (changes requested):
Nicely done, I really appreciate 2 main points of this:

* I love standardizing the header ordering, and I also like the reasoning for the 3 groups.
* We really do need to reduce the clang-tidy spam, and I agree with the vast majority of removals. I added notes on a small number of them.

What I'm ambivalent about is the sheer amount of includes that are the result of strict include what you use. But we decided on that earlier. I guess in some cases it might indicate which headers have too much responsibility or not enough encapsulation.

What I really don't like is (bikeshedding warning!) the extra newline. It also seems like this is the opposite of formalizing the established practice of our codebase, since it seems like there are far more files where it is added than there are others.

src/backend/sycl_cuda_backend.cc
fknorr: It split up the include into multiple internal cuda headers above. fixed.

.clang-tidy
fknorr: I get this somewhat frequently. A good example is `out_of_order_engine_detial::engine_impl::pop_assignable`, which has two of these.

I believe there are two distinct situations for else vs return:
- The `then` branch is an early exit, because the operation is a noop with the parameters specified, a cached value can be returned, or similar. In this case, the remainder of the function (after the if+return) is the common or complex case for the fuction being invoked
- The `then` and `else` branches have similar or equal weight, and the function would be just as good if we flipped the two. In that case I want to see two blocks that are similarly scoped / indented to highlight that there are two equally valid paths for execution to take.

The lint doesn't distinguish between these cases.

fknorr:
Re newline: I don't feel strongly about this, but visually it does help me in files with huge include lists like `instruction_graph_generator.cc`.

And I agree the number of includes is a bit excessive, but I have trouble coming up with a much better rule dictating which includes must be present. We could though, for example, omit all includes from a .cc file that have already happened in the .h file of the same name.

PeterTh (approved):

PR #313, opened 2024-11-27
Extend lookahead to cover host-buffer allocations for send / receive

fknorr:
Multi-node RSim surfaced the fact that we still generate frequent small allocations because we do not `anticipate()` push and await-push commands, which do allocate (buffer) host memory.

We want to replace strided-host-buffer staging of MPI transfers with linearized `copy_plan` staging eventually, but for the meantime this PR avoids frequent host-buffer reallocations (and with it, potential runaway scheduling complexity) around sends and receives by accumulating all boxes they potentially touch during `anticipate()` as well.


PeterTh (approved):

psalz (approved):
LGTM!

PR #314, opened 2024-11-27
Exclude src/testspy from coverage

fknorr:
Since #308, scheduler / runtime testspies are located in `src`, which was included in coverage reports. Since it's test code, it shouldn't be.

I've also removed one leftover branch from #283 I found in the coverage report.


GagaLP (approved):
LGTM!

psalz (approved):

PR #315, opened 2024-11-27
Re-define `access_mode` explicitly, remove access_modes.h header

fknorr:
The `celerity::access_mode = sycl::access_mode` alias was a leftover from pre-USM backed Celerity. Re-defining the enum explicitly allows us to simplify the include tree somewhat. This also updates tests that were previously using the SYCL type.

This has one "breaking change" that I'm not sure if we need to communicate: Users can't just pass SYCL access modes anymore (was undocumented), and `access_mode::atomic` has been removed (wasn't supported on accessors anyway, but could be enumerated).

Removing `access_modes.h` (which contained functionality that was in parts only used by tests) and `struct mode_traits` resolves the ambiguity between namespaces celerity::access and celerity::detail::access (now removed).


GagaLP (approved):
Nicely done. 
This was one thing, which always confused me. 
LGTM!

psalz (approved):
LGTM!

PR #317, opened 2024-12-04
Small optimizations on iggen buffer handling

fknorr:
`perform_task_buffer_accesses` updates last-writers twice to gracefully handle overlapping writes, which is an edge case. This PR quickly checks if overlapping writes are present, and sticks to a single update if there are not. By transposing the loop nest from `chunk -> bid` to `bid -> chunk`, we can also save avoid constructing another `unordered_map`.

Results are not looking too impressive in the benchmark report, but I do get a consistent 4% speedup for RSim `room_small`, which is scheduler bound on gpuc3.


GagaLP (approved):
Nicely done.
LGTM! 👍

src/instruction_graph_generator.cc
```diff
@@ -1865,80 +1867,76 @@ instruction* generator_impl::launch_task_kernel(batch& command_batch, const exec
...
+		for(auto& allocation : memory.allocations) {
+			add_dependencies_on_last_writers(command_instructions[i], allocation, region_intersection(concurrent_reads[i], allocation.box));
+			add_dependencies_on_last_concurrent_accesses(command_instructions[i], allocation, region_intersection(concurrent_writes[i], allocation.box),
+			    instruction_dependency_origin::write_to_allocation);
 		}
 	}
 
-	// 3. Clear tracking structures for all regions that are being written to. We gracefully handle overlapping writes by treating the set of all conflicting
-	// writers as last writers of an allocation.
+	// 3. To gracefully handle overlapping writes, clear tracking structures for all regions that are being written to, so we can treat the set of all
```
psalz, line 1910:
```suggestion
	// 3. To gracefully handle overlapping writes if detection is disabled via the error policy, clear tracking structures for all regions that are being written to, so we can treat the set of all
```

src/instruction_graph_generator.cc
```diff
@@ -1865,80 +1867,76 @@ instruction* generator_impl::launch_task_kernel(batch& command_batch, const exec
...
+		for(auto& alloc : buffer.memories[concurrent_chunks[i].memory_id].allocations) {
+			alloc.track_concurrent_read(region_intersection(alloc.box, concurrent_reads[i]), command_instructions[i]);
+			if(has_overlapping_writes) {
+				alloc.track_concurrent_write(region_intersection(alloc.box, concurrent_writes[i]), command_instructions[i]);
+			} else {
+				alloc.track_atomic_write(region_intersection(alloc.box, concurrent_writes[i]), command_instructions[i]);
 			}
-			buffer.track_original_write(rw.writes, command_instructions[i], concurrent_chunks[i].memory_id);
 		}
+		buffer.track_original_write(concurrent_writes[i], command_instructions[i], concurrent_chunks[i].memory_id);
```
psalz, line 1939:
```suggestion
		// On the buffer level, there is no special handling of overlapping writes: The last chunk that writes to a given
		// region becomes its last writer, and subsequent reads on other devices will require a copy.
		buffer.track_original_write(concurrent_writes[i], command_instructions[i], concurrent_chunks[i].memory_id);
```

psalz (approved):
LGTM! I've suggested two comment changes that I've added for my understanding while investigating how to implement replicated writes!

PR #319, opened 2024-12-10
Cache the set of active transfers in receive arbiter

fknorr:
`receive_arbiter` tracks a `transfer` for every transfer id it learns through an inbound pilot, so long-running programs like `wave_sim -T 10000` will fill the transfer map quickly. It's an `unordered_map` so lookups shouldn't suffer, but we iterate over it in `poll_communicator`, which has linear complexity (initially found by @psalz).

We only need to poll transfers for which we have already seen the `receive_instruction`. This PR avoids the linear growth by caching the set of active transfers in a vector, and iterating over that.

I've used the opportunity to refactor the code a bit and use `std::erase_if` in the many places we had erase-remove idiom before.


psalz (approved):
Good stuff!

GagaLP (approved):
Nice

PR #321, opened 2024-12-19
Improve wording of uninitialized read warning

psalz:
The previous warning suggested to use `no_init` even for pure-consumer (`read_only`) accesses, which never made any sense.

### DIFF
--- a/src/task_manager.cc
+++ b/src/task_manager.cc
@@ -89,9 +89,20 @@ namespace detail {
 					}
 				}
 				if(!uninitialized_reads.empty()) {
-					utils::report_error(m_policy.uninitialized_read_error,
-					    "{} declares a reading access on uninitialized {} {}. Make sure to construct the accessor with no_init if possible.",
-					    print_task_debug_label(tsk, true /* title case */), print_buffer_debug_label(bid), std::move(uninitialized_reads).into_region());
+					bool is_pure_consumer_access = true;
+					for(size_t i = 0; i < access_map.get_num_accesses(); ++i) {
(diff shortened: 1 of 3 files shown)


PeterTh (approved):

src/task_manager.cc
```diff
@@ -89,9 +89,20 @@ namespace detail {
...
-					    "{} declares a reading access on uninitialized {} {}. Make sure to construct the accessor with no_init if possible.",
-					    print_task_debug_label(tsk, true /* title case */), print_buffer_debug_label(bid), std::move(uninitialized_reads).into_region());
+					bool is_pure_consumer_access = true;
+					for(size_t i = 0; i < access_map.get_num_accesses(); ++i) {
+						const auto [b, mode] = access_map.get_nth_access(i);
+						if(b == bid && is_producer_mode(mode)) {
+							is_pure_consumer_access = false;
+							break;
+						}
+					}
```
psalz, line 99: This is also just a heuristic btw; there could be an uninitialized `read` and an initialized `write` on the same buffer and it would print the `no_init` advice. Unfortunately I don't think this can be avoided unless we do the check on a per-accessor basis (which would mean separately evaluating each range mapper, followed by a region map query).

GagaLP (approved):
The new wording makes it more readable. Nice

src/task_manager.cc
fknorr: This is still better than what we have now! ... and we should at some point improve the `buffer_access_map` interface so this can just be a `std::any_of`.

PR #322, opened 2024-12-20
Measure and report executor starvation time

psalz:
We previously had a rudimentary Tracy zone for executor starvation, however this was only semi-correct, because if the scheduler is not currently producing any instructions (for example if the user is doing something other than submitting tasks to Celerity), then the executor is not really starving.

I've extended the scheduler to report its idle state to the runtime, which in turn feeds it to the executor for measuring starvation. If the total starvation time exceeds a percentage threshold of the time spent doing actual work (i.e., processing instructions), we print a warning indicating that the application might be scheduler-bound.

Since the scheduler looks at task and command queue sizes to decide whether it is busy or idle, the lookahead mechanism can cause the scheduler to remain busy even if it is not actually producing any instructions. I consider this a feature rather than a bug, because it means that applications that attempt to interleave Celerity kernels with other work will produce a starvation warning if the kernels aren't actually being executed. I've opted to include a hint for this in the warning itself.


PeterTh (approved):
LGTM

fknorr (changes requested):
Nice, another simple but very useful debugging tool!

I agree on how starvation time is measured, but I believe that the design can be improved by moving the actual tracking from `live_executor` to `runtime`.

Proposal:
- Move `total_starvation_time` and `total_active_time` into `class runtime::impl`
- Have `on_executor_busy` and `on_executor_idle` in `executor_delegate`, mirroring `scheduler_delegate`
- If either scheduler or executor change between busy/idle states, take a lock and record which of the 4 possible states we are in (each event also means suspending / resuming a thread, so I suppose the overhead is manageable)
- Remove getters (and dry-run dummy implementations) from executors

This would have the following advantages:
- executor would not need to concern itself with scheduler state conceptually (they are fully separated atm)
- executor public API would not expose any query functions, which makes reasoning about its behavior in the multithreaded context easier
- Cross-thread counter updates would be centralized in runtime (precedence: `runtime::impl::m_latest_epoch_reached`)
- `idle_state_change` would not need to traverse the executor queue, so its reporting would be more accurate in time.

The only downside I see in this is that the Tracy integration will not be able to distinguish between executor starvation and idleness.

PR #323, opened 2024-12-20
Change split functions to work on box instead of chunk

psalz:
`chunk` is a purely user-facing type (and of questionable usefulness at that), and since all of our internal tracking is built around `box`, splitting should follow suit.

### DIFF
--- a/include/command_graph_generator.h
+++ b/include/command_graph_generator.h
@@ -217,7 +217,7 @@ class command_graph_generator {
 
 	struct assigned_chunk {
 		node_id executed_on = -1;
-		chunk<3> chnk;
+		box<3> chnk;
 	};
 
 	struct buffer_requirements {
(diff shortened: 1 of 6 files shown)


GagaLP (approved):
Nice

fknorr (changes requested):
I like it!

If I'm not mistaken you could also define `command_graph_generator::assigned_chunk` in terms of boxes and get rid of even more casts without much hassle - `buffer_access_map` already operates on boxes.

Also, a drive-by suggestion:
```patch
diff --git a/src/command_graph_generator.cc b/src/command_graph_generator.cc
index da56682e..3cfb6d22 100644
--- a/src/command_graph_generator.cc
+++ b/src/command_graph_generator.cc
@@ -117,11 +117,9 @@ std::vector<const command*> command_graph_generator::build_task(const task& tsk)
 }

 void command_graph_generator::report_overlapping_writes(const task& tsk, const box_vector<3>& local_chunks) const {
-       const chunk<3> full_chunk{tsk.get_global_offset(), tsk.get_global_size(), tsk.get_global_size()};
-
        // Since this check is run distributed on every node, we avoid quadratic behavior by only checking for conflicts between all local chunks and the
        // region-union of remote chunks. This way, every conflict will be reported by at least one node.
-       const box<3> global_chunk(subrange(full_chunk.offset, full_chunk.range));
+       const box<3> global_chunk(subrange(tsk.get_global_offset(), tsk.get_global_size()));
        auto remote_chunks = region_difference(global_chunk, region(box_vector<3>(local_chunks))).into_boxes();

        // detect_overlapping_writes takes a single box_vector, so we concatenate local and global chunks (the order does not matter)
```

PR #324, opened 2024-12-20
Generate single await-push command per buffer for all local chunks

psalz:
This brings await-pushes in line with pushes, where we already compute the union of all regions required by remote chunks executed on the same node.


src/command_graph_generator.cc
```diff
@@ -388,33 +388,40 @@ void command_graph_generator::generate_pushes(batch& current_batch, const task&
...
 			const auto local_sources = buffer.local_last_writer.get_region_values(consumed);
-			region_builder<3> missing_part_boxes;
+			box_vector<3> missing_parts_boxes;
 			for(const auto& [box, wcs] : local_sources) {
 				// Note that we initialize all buffers as fresh, so this doesn't trigger for uninitialized reads
-				if(!box.empty() && !wcs.is_fresh()) { missing_part_boxes.add(box); }
+				if(!box.empty() && !wcs.is_fresh()) { missing_parts_boxes.push_back(box); }
 			}
 
 			// There is data we don't yet have locally. Generate an await push command for it.
```
GagaLP, line 406: For me this comment now seems a bit out of place since we don't generate the await push command in this if.

GagaLP (approved):
Other than the small comment i made this looks good.

include/grid.h
```diff
@@ -273,6 +273,13 @@ class region_builder {
 		m_boxes.push_back(box);
 	}
 
+	void add(const box_vector<Dims>& boxes) & {
+		m_boxes.reserve(m_boxes.size() + boxes.size());
+		for(const auto& b : boxes) {
```
fknorr, line 278: Maybe have a comment here that we use a loop rather than `insert(end)` to skip over empty boxes, something which `add(region)` does not need to do.

[only if my comment below does not apply]

src/command_graph_generator.cc
```diff
@@ -388,33 +388,40 @@ void command_graph_generator::generate_pushes(batch& current_batch, const task&
...
+			if(!missing_parts_boxes.empty()) {
 				assert(m_num_nodes > 1);
-				auto* const ap_cmd = create_command<await_push_command>(current_batch, transfer_id(tsk.get_id(), bid, no_reduction_id), missing_parts,
-				    [&](const auto& record_debug_info) { record_debug_info(buffer.debug_name); });
-				generate_anti_dependencies(tsk, bid, buffer.local_last_writer, missing_parts, ap_cmd);
-				generate_epoch_dependencies(ap_cmd);
-				// Remember that we have this data now
-				buffer.local_last_writer.update_region(missing_parts, {ap_cmd, true /* is_replicated */});
+				auto& required_boxes = per_buffer_required_boxes[bid]; // allow default-insert
+				required_boxes.add(missing_parts_boxes);
```
fknorr, line 410: Couldn't this just be moved into the loop in L401? Then we don't need `missing_parts_boxes` (or `region_builder::add(box_vector)` for that matter)

fknorr (commented):
Seems like the correct thing to do!

PR #325, opened 2024-12-23
Set up graph_query testing infrastructure for task graph

psalz:
This brings the TDAG (mostly) in line with CDAG and IDAG. What's missing is a hierarchical structure for `task_record`, but since unlike for commands and instructions, a call to `submit()` only creates a single task (except for horizons) it turns out we don't really need this functionality anyway.


GagaLP (approved):

fknorr (approved):
Very cool! This makes me want to start the TDAG refactor.

PR #327, opened 2025-01-14
Revisit DAG benchmarks

psalz:
This overhauls the DAG benchmarks to address some long standing issues, as well as to future-proof them for loop templates:

- Instead of measuring the combined time taken to generate a graph and all the graphs that come before it (i.e., TDAG -> CDAG -> IDAG), only measure the generation of the graph being benchmarked.
- Don't measure benchmark context creation / destruction.
- Fix IDAG benchmarks creating host tasks instead of device tasks (i.e., number of devices and p2p copy support had no effect until now).
- Fix scheduler benchmarks comparing against single-threaded TDAG + CDAG generation, instead of all three graphs.
- Remove outdated "throttled submission" scheduler benchmarks (these were designed to measure the impact of the scheduler's lock-step operation with the main thread, which has long been removed).

Due to these changes, we should expect CDAG benchmark times to go down, because they no longer include TDAG generation. IDAG benchmarks no longer include TDAG / CDAG, but due to the switch from `host_task` to `parallel_for`, this is a tossup (benchmarks with lots of communication, such as the (strangely named) `chain topology` now take a lot longer).

For the scheduler benchmarks I've decided to include the TDAG generation time in the main thread, which is overlapped with CDAG / IDAG generation in the scheduler thread, as this best reflects reality.

I've also made some improvements to our CI perf impact plotting script:
- Group benchmarks by category
- Left-align legend, one entry per line
- Don't create images for categories without changes ("No Data")
- Support newly added and removed benchmarks


PeterTh (approved):
LGTM

test/dag_benchmarks.cc
```diff
@@ -100,108 +100,159 @@ static constexpr instruction_graph_generator::policy_set benchmark_instruction_g
...
 	command_recorder crec;
 	command_graph_generator cggen{num_nodes, 0 /* local_nid */, cdag, test_utils::g_print_graphs ? &crec : nullptr, benchmark_command_graph_generator_policy};
 	test_utils::mock_buffer_factory mbf{tm, cggen};
 
-	explicit command_graph_generator_benchmark_context(const size_t num_nodes) : num_nodes(num_nodes) { tm.generate_epoch_task(epoch_action::init); }
+	explicit cdag_benchmark_context(const size_t num_nodes) : tdag_benchmark_context(num_nodes) {}
+
+	void initialize() {
+		tdag_benchmark_context::initialize();
+		create_all_commands();
```
fknorr, line 165: This doesn't do anything since `m_tasks` is empty, right?

Also, what is it supposed to do, pre-allocate `m_command_batches`? Maybe also add a comment.

test/dag_benchmarks.cc
```diff
@@ -100,108 +100,159 @@ static constexpr instruction_graph_generator::policy_set benchmark_instruction_g
...
+
+	void task_created(const task* tsk) override { m_tasks.push_back(tsk); }
+
+	void initialize() { tm.generate_epoch_task(celerity::detail::epoch_action::init); }
+
+	void prepare() { m_tasks.reserve(m_command_groups.size()); }
+
+	void execute() { create_all_tasks(); }
+
+	void finalize() {
```
fknorr, line 140: Maybe have a comment about what the purpose of these four functions is (I would naively expect all setup to happen in the ctor, and the benchmarked code to be in a single member function).

test/dag_benchmarks.cc
```diff
@@ -100,108 +100,159 @@ static constexpr instruction_graph_generator::policy_set benchmark_instruction_g
...
+	void execute() { create_all_commands(); }
 
-	command_graph_generator_benchmark_context(const command_graph_generator_benchmark_context&) = delete;
-	command_graph_generator_benchmark_context(command_graph_generator_benchmark_context&&) = delete;
-	command_graph_generator_benchmark_context& operator=(const command_graph_generator_benchmark_context&) = delete;
-	command_graph_generator_benchmark_context& operator=(command_graph_generator_benchmark_context&&) = delete;
+	void finalize() {
+		m_tasks.clear();
+		tdag_benchmark_context::finalize();
+		create_all_commands();
```
fknorr, line 179: Why do commands need to be created in what appears to be a teardown function?

fknorr (commented):
Like the new visualization!

PR #333, opened 2025-08-14
Remove flush_async workaround as it is no longer needed in newer ACPP versions

GagaLP:
Removes the `flush_async` workaround for newer ACPP versions, as the asynchronous submission worker thread was removed in [Acpp PR #1798](https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1798).
Maintains compatibility with older ACPP versions.


PeterTh (approved):
LGTM

PeterTh:
The coverage reporting run is currently failing not because of anything on our side, but because coveralls.io seems to have problems.

Since this PR doesn't substantially change anything about coverage we could still merge it, but that would mean that the main build would be red, so let's wait a bit to see if this clears up.

PR #335, opened 2026-05-28
Fix: undefined buffer_snapshot move assignment

GagaLP:
This fixes a problem I came across while porting Celerity to Windows. The move assignment operator never returned `*this`, which resulted in really weird bugs on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [WIP]
+
+This release is WIP
+
+### Fixed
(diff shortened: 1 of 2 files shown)


PeterTh:
Surprising that this went unnoticed so long. Can we get a regression test for it? (Ideally one that actually fails on Linux before the fix)

GagaLP:
> Surprising that this went unnoticed so long. Can we get a regression test for it? (Ideally one that actually fails on Linux before the fix)

This went unnoticed for so long because the move assignment operator was never actually used in our code. Interestingly, we do use the flag `-Werror=return-type`, which should normally catch this, but since `buffer_snapshot` is a class template, the compiler only checks a member function's body (including missing-return checks) when it's actually instantiated. Since nothing in our codebase move-assigns a `buffer_snapshot`, that function was never instantiated, so the compiler never analyzed it and the warning never had a chance to fire.

PeterTh (approved):

PR #336, opened 2026-07-14
Fix Windows compatibility issues

GagaLP:
This PR addresses a collection of smaller fixes to get Celerity working on Windows.

The most notable changes are:

* Adding and updating header files that were missing or incompatible with Windows.
* Adding support for demangling names from pointer types.
* Using MSVC's `[[msvc::no_unique_address]]` attribute.


include/workaround.h
```diff
@@ -6,6 +6,16 @@
...
+#ifdef _WIN32
+// Prevent Windows headers from polluting global namespace with min/max macros.
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#endif
+
```
PeterTh: This is generally a good idea, but usually done right before `windows.h` is included. Is there a reason why it is in this particular spot?

src/backend/sycl_backend.cc
```diff
@@ -116,7 +116,7 @@ struct sycl_backend::impl {
 		sycl::context sycl_context;
 		std::vector<sycl::queue> queues;
 		std::optional<detail::thread_queue> submission_thread;
-		std::atomic_flag active_async_error_check = false;
+		std::atomic_flag active_async_error_check; // since c++20 atomic_flag is set to clear by default (false)
```
PeterTh, line 119: Is this change necessary?

src/utils.cc
```diff
@@ -20,24 +20,76 @@
...
+	}
+
+	return s;
+}
+#endif
+
 std::string get_simplified_type_name_from_pointer(const std::type_info& pointer_type_info) {
-#if !defined(_MSC_VER)
+// This type of demangling can be done if the abi header is available. This also works for Clang on Windows, if
+// compiled with Itanium ABI on Windows (e.g. MinGW/MSYS2). MSVC
```
PeterTh: MSVC ?

src/utils.cc
```diff
@@ -20,24 +20,76 @@
...
 	const std::unique_ptr<char, void (*)(void*)> demangle_buffer(abi::__cxa_demangle(pointer_type_info.name(), nullptr, nullptr, nullptr), std::free);
 	std::string demangled_type_name = demangle_buffer.get();
-#else
-	std::string demangled_type_name = pointer_type_info.name();
-#endif
 
 	// get rid of the pointer "*"
 	if(!demangled_type_name.empty() && demangled_type_name.back() == '*') { demangled_type_name.pop_back(); }
+#else
+	std::string demangled_type_name = msvc_undecorate_type_name(pointer_type_info.name());
```
PeterTh, line 91: this kind of assumes that anything not fulfilling `__has_include(<cxxabi.h>)` is necessarily `MSVC`. That's probably not unlikely in practice, but I'd prefer an additional check (and maybe error out in the final else).

PeterTh (changes requested):

src/backend/sycl_backend.cc
GagaLP: It is interesting that we never had a problem with this until now, since `= false` was never correct in the first place. In C++20 we don't need initialization since the default constructor performs value initialization, and this deprecated the previously "correct" way of initializing it with `= ATOMIC_FLAG_INIT`, which would have been the correct way until now.

I would say since we are on C++20, let's leave it like this to avoid the deprecation warning of `= ATOMIC_FLAG_INIT`.

include/workaround.h
GagaLP: The reason was that I thought about where it makes sense to put it. Workarounds, even platform-specific ones, would thematically belong in this header, but I agree that doing it right before `windows.h` makes it less prone to errors.

Even more so, I went through all the files with `windows.h` and unified them to be structured like this.

src/utils.cc
GagaLP: I don't remember what I intended here. I'll remove it.

src/utils.cc
GagaLP: I completely agree. I'll fix it.

PeterTh:
LGTM now.

PeterTh (approved):

PR #337, opened 2026-07-14
Adapt tests for Windows compatibility

GagaLP:
This PR is a follow-up to PR #336 and fixes several tests to improve Windows support.

The changes affect task, command, and instruction graph printing tests as well as `live_executor` starvation/active-time tests:

* The task, command, and instruction graph printing tests compare exact output strings, which depend on the iteration order of `std::unordered_map`. Since MSVC's standard library uses a different iteration order, these tests fail on Windows even though graph construction is correct. The tests are disabled for now until a more robust graph comparison method is available.
* The `live_executor` starvation/active-time test tolerance is now platform-dependent, using 20ms on Windows (where the previous tolerance was too strict) and 10ms on other platforms, instead of increasing the tolerance globally.


test/print_graph_tests.cc
```diff
@@ -13,7 +13,19 @@ using namespace celerity::test_utils;
 
 namespace acc = celerity::access;
 
+namespace {
+#ifdef _WIN32 // Windows' std::unordered_map sorts differently than other standard library implementations, which causes the printed graph to differ. Since the
+              // tests at the moment rely on exact string matching, we disable them on Windows until we have a better solution.
+#define SKIP_UNSUPPORTED()                                                                                                                                     \
```
PeterTh: This should be a bit more specific, maybe something like `SKIP_WIN32_GRAPH_PRINT_UNSUPPORTED`.

PeterTh (changes requested):
One small change request, looks good otherwise.

test/print_graph_tests.cc
GagaLP: Changed the name

PR #338, opened 2026-07-14
Add Windows thread affinity support

GagaLP:
Thread pinning was previously a no-op on Windows: `thread_pinner` only emitted a warning (`"Thread pinning is currently not supported on Windows."`) and did not apply any affinity settings, regardless of configuration.

This PR adds a real implementation by introducing a small pthread/`cpu_set_t` compatibility layer (`platform_specific/affinity_win32.h`/`.cc`). The layer maps the POSIX affinity APIs already used by Celerity (`sched_getaffinity`/`sched_setaffinity`, `pthread_self`, `pthread_get/setaffinity_np`, `CPU_SET`/`CPU_ISSET`/`CPU_COUNT`, etc.) to their Windows equivalents (`GetActiveProcessorGroupCount`, `GROUP_AFFINITY`, `SetThreadGroupAffinity`).

With this compatibility layer in place, `affinity_win.cc` no longer requires separate pinning logic and now closely mirrors the Linux implementation. It initializes and tears down the pinning plan in the same way and pins threads to sequential cores using the same approach.

One limitation to note: Windows organizes logical processors into groups of up to 64 processors, and a single `GROUP_AFFINITY` mask cannot span multiple groups. If a requested core set crosses a group boundary, the layer currently emits a warning and skips pinning instead of attempting cross-group affinity. This should not affect the common case of pinning a small number of sequential cores, but remains a limitation on systems with more than 64 logical processors.

Added corresponding tests in `affinity_tests.cc`.


test/affinity_tests.cc
```diff
@@ -63,7 +63,7 @@ class raii_test_runtime {
 };
 
 #ifdef _WIN32
-#define SKIP_UNSUPPORTED() SKIP("Affinity is not supported on Windows");
+#define SKIP_UNSUPPORTED() // SKIP("Affinity is not supported on Windows");
 #else
 #define SKIP_UNSUPPORTED()
 #endif
```
PeterTh: This whole code block (and the invocation sites) can just be removed.

include/platform_specific/affinity_win32_adapter.h
```diff
@@ -0,0 +1,105 @@
+#pragma once
+
```
PeterTh, line 4: This file should have a comment at the start explaining its purpose (i.e. implementing unix-like affinity using windows APIs).

maybe the name should also be something like `affinity_win32_adapter`?

PeterTh (changes requested):
Instead of duplicating the win.cc / unix.cc, I think the win version could just be reduced to a include of the helper header plus a include of the unix.cc, with an explanation (of the windows helper introducing compat with the unix impl).

test/affinity_tests.cc
GagaLP: Done!

include/platform_specific/affinity_win32_adapter.h
GagaLP: Done!

PR #339, opened 2026-07-23
Make cgf_diagnostics::teardown() a no-op instead of resetting the instance

GagaLP:
After a while of debugging a strange `Access Violation` error on Windows during teardown of Celerity, I traced it back to `cgf_diagnostics::teardown()`.

`cgf_diagnostics` stores its singleton in a `thread_local std::unique_ptr`. `teardown()` is called from `runtime::impl`'s shutdown path, which runs when `runtime::s_instance` (a plain `static` member) is destroyed, and if the application never calls `runtime::shutdown()` explicitly, that happens as part of normal static-object teardown after `main()` returns.

The C++ standard leaves the relative order of static-object destruction and `thread_local` destruction for the main thread undefined. On Windows, these use separate mechanisms (the CRT's atexit/exit machinery vs. TLS callbacks), and in this case the `thread_local` storage for the main thread was already gone by the time `teardown()` called `m_instance.reset()` on it, resulting in access to unavailable memory.

Removing the `m_instance.reset()` call fixes this issue while keeping the logic the same: the whole point of using a smart pointer here is that its cleanup is already handled automatically, so an explicit reset in `teardown()` is redundant.


PeterTh (approved):
We discussed this a lot offline and I think this is the best solution.

